PR #597, opened 2013-12-12
Optionally allow gem dependencies to be installed

svanhess:
An option should be provided to include all gem dependencies when creating a package.  This is useful when creating a rubygem debian package that is "all inclusive".  You wouldn't have to create separate debian packages for each gem dependency.

### DIFF
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -50,7 +50,7 @@ Gem::Specification.new do |spec|
   # For creating FreeBSD package archives (xz-compressed tars)
   spec.add_dependency("ruby-xz") # license: MIT
 
-  # For sourcing from pleaserun 
+  # For sourcing from pleaserun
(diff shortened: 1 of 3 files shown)


jordansissel:
I'm open to merging this. I have one question though - any thoughts on renaming the flag? I think `--include-dependencies` isn't clear enough on the behavior. Maybe something like `--embed-dependencies` or something? I dunno.. IDeas?

hatt:
+1 for `--embed-dependencies`, it's a bit clearer and allows us the possibility of recursive dependencies packaging vs bundling.

josephholsten:
@svanhess this sounds really useful. Would you be able to help get the patch updated to be mergable?

jordansissel:
:heart: for taking the time for rebasing this after nearly 2 years ;)

Thank you for making FPM better! :)

svanhess:
@jordansissel no problem!  Glad to help a project that I've benefited from throughout the years.

PR #626, opened 2014-01-28
Add input and output rpm trigger support

makushimu:
I needed to be able to use rpm triggers and I noticed that fpm didn't support them... I just added the support with rpm-specific options.

It would be nice to be able to support "generic" for package systems that support them (I'm thinking at .deb specifically), but I didn't have time to play with this.


rbjorklin:
@makushimu could you provide an example on how to use this? I don't understand the documentation.

PR #671, opened 2014-04-06
Add --package-template-dir option to overwrite the spec by user's templates

repeatedly:
From this thread: "How to specify own spec?" https://groups.google.com/forum/#!topic/fpm-users/GeH7xHhBo80

I didn't get the answer so I implemented this feature.
To use --package-template-dir, we can specify full customized spec instead of bundled template.


josephholsten:
@repeatedly I'd be happy to rebase this, but do we really want it? do you think fpm still needs spec support?

PR #683, opened 2014-04-25
Added rpm-provides option

robkinyon:
This adds the ability to specify additional provides elements. For example, I might want to say "This RPM also provides 'mail_reader'" so that another RPM can require 'mail_reader' and be satisfied.

I didn't see if any other packaging system allows for additional provides elements like RPM does. If so, please modify as needed.


robkinyon:
Sure - probably could use additional documentation. Given that, I'm good with dropping this PR.

jordansissel:
+1 that we should document this better. <3 for the feedback.

PR #687, opened 2014-04-29
rpm: Add a option to insert raw Require statements

vStone:
This can be useful to have requirements for different stages.

Examples:

    --rpm-add-require 'Requires(post): /usr/sbin/alternatives'
    --rpm-add-require 'Requires(preun): /usr/sbin/alternatives'


jordansissel:
Thanks for your work revising this patch <3

Merged!

Issue #748, opened 2014-08-06
RedHat Perl cannot find dependencies


hatt:
Initially reported in #515 but unrelated to the actual bug in that issue, RedHat's packaging of Perl module directories is incorrect and currently breaks FPM and cpanm --self-contained installs that require 3rd party core modules in RHEL and CentOS 7.

The RedHat packaging for perl separates directories for core, 3rd party, and CPAN installed modules. Some parts of core are developed and maintained by 3rd party developers, and as such get installed into vendorlib. The directories used by cpanm for finding core modules are handled by only::core::lib, and evaluate to privlib and archlib specifically. As an example, ExtUtils::MakeMaker is installed to privlib/vendor_perl, which is not read as a core directory. A separate circular dependency between EU::MM and Data::Dumper prevents installing this into the sandbox, so we rely on core for the modules which it ends up not being able to find.

I believe @satoshi is filing a bug with RedHat concerning this. Essentially they need to update their list of core modules and install the 3rd party core modules as if they were internal core. Until then, cpanm is unable to find modules it needs to function normally and due to the circular dependency of EU::MM and DD, there is no way to run in a sandbox entirely. While inconvenient, I do not think this is a bug on EU::MM and DD's part as modules can be installed manually in this case.

satoshi:
I believe this issue is fixed after #752 has been merged and fpm-1.4.0 is released.  Thanks for this excellent patch and close out this issue whoever that can do so.

Issue #515, opened 2013-08-06
Can not package Data::Structure::Util


romabysen:
When I try to package Data::Structure::Util with "fpm -s cpan -t rpm Data::Structure::Util" I get the following error:

```
! Installing /tmp/package-cpan-build20130806-31164-i8r0u2/module failed. See /home/lars/.cpanm/work/1375775860.31190/build.log for details. Retry with --force to force install it. {:level=>:error, :file=>"cabin/mixins/pipe.rb", :line=>"46"}
Process failed: cpanm failed (exit code 1). Full command was:["cpanm", "-L", "/tmp/package-cpan-build20130806-31164-i8r0u2/cpan", "/tmp/package-cpan-build20130806-31164-i8r0u2/module"] {:level=>:error, :file=>"fpm/command.rb", :line=>"411"}
```

Interestingly enough it works fine with "cpanm Data::Structure::Util".
This is on CentOS6.

PR #752, opened 2014-08-07
CPAN test and sandboxing fixes

hatt:
As discussed in #515 and #748, make sandboxing optional for packaging with system dependencies. This doesn't fix the core issue with RHEL and derivatives, but it does allow users to not install every dependency if they want a build to use system installed modules.

Up til now, `--cpan-test` would skip tests when told to test and then run them when told not to. That was bad logic on my part I think, so this reverses the behavior and makes it do what a user would expect.

### DIFF
--- a/lib/fpm/package/cpan.rb
+++ b/lib/fpm/package/cpan.rb
@@ -109,6 +109,7 @@ class FPM::Package::CPAN < FPM::Package
       cpanm_flags = ["-l", build_path("cpan"), moduledir]
     end
 
+    cpanm_flags += ["-n"] if !attributes[:cpan_test?]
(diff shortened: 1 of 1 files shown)


satoshi:
@jordansissel is there a chance this pull request can be merged?  @hatt  has done some wonderful work on this and I'd like to see it incorporated into master.

jordansissel:
+1 on the idea, sorry for the delays. @hatt, can you rebase against master? I'll work on merging after that.

jordansissel:
Actually, I May be able to patch this in by hand now.

hatt:
@jordansissel sorry, I just saw comment now, thanks for doing that for me!

Issue #730, opened 2014-06-26
--config-files in conflict with --prefix option


vazic:
$mkdir test;cd test
$echo '1' > default
$fpm -s dir -t rpm  \
  --name "TEST" -v "1" \
  --iteration 1 --epoch 1 \
  --license "PRIVATE" --vendor "TEST" --architecture "noarch" \
  --config-files "default" \
  --prefix "/opt" \
  "./"

/home/vaz/.rvm/rubies/ruby-1.9.3-p545/lib/ruby/1.9.1/find.rb:38:in `block in find': No such file or directory (Errno::ENOENT)

CentOS 5

example works fine either without `--prefix` or `--config-files` options

zoresvit:
The same issue happens with `dir` source when specifying destination directories. So command 

```
fpm -s dir -t deb -n redis --config-files /etc/redis/redis.conf -v 2.6.10 --prefix /opt ./src/redis-server=/usr/bin redis.conf=/etc/redis
```
fails with error:

```
/usr/share/ruby/find.rb:40:in `block in find': No such file or directory (Errno::ENOENT)
        from /usr/share/ruby/find.rb:40:in `collect!'
        from /usr/share/ruby/find.rb:40:in `find'
        from /usr/local/share/gems/gems/fpm-1.4.0/lib/fpm/package/rpm.rb:447:in `block in output'
        from /usr/local/share/gems/gems/fpm-1.4.0/lib/fpm/package/rpm.rb:446:in `each'
        from /usr/local/share/gems/gems/fpm-1.4.0/lib/fpm/package/rpm.rb:446:in `output'
        from /usr/local/share/gems/gems/fpm-1.4.0/lib/fpm/command.rb:475:in `execute'
        from /usr/local/share/gems/gems/clamp-0.6.5/lib/clamp/command.rb:67:in `run'
        from /usr/local/share/gems/gems/fpm-1.4.0/lib/fpm/command.rb:535:in `run'
        from /usr/local/share/gems/gems/clamp-0.6.5/lib/clamp/command.rb:132:in `run'
        from /usr/local/share/gems/gems/fpm-1.4.0/bin/fpm:8:in `<top (required)>'
        from /usr/local/bin/fpm:23:in `load'
        from /usr/local/bin/fpm:23:in `<main>'
```

I understand that the arguments are incompatible, but still the program shouldn't crash and should provide descriptive error message instead.

Issue #584, opened 2013-11-14
Config files confused by prefix


hatt:
When a prefix is passed, RPM gets confused as to where config files are. With current HEAD, I can't use a relative config file path (see #568 for that), but I get an error as follows when the non-full path WITH prefix is used (ie opt/app/configs/whatever.ini). The full path cannot have a leading slash or it fails with the same result.

```
/Users/matt/.rbenv/versions/2.0.0-p247/lib/ruby/2.0.0/find.rb:40:in `block in find': No such file or directory (Errno::ENOENT)
    from /Users/matt/.rbenv/versions/2.0.0-p247/lib/ruby/2.0.0/find.rb:40:in `collect!'
    from /Users/matt/.rbenv/versions/2.0.0-p247/lib/ruby/2.0.0/find.rb:40:in `find'
    from /Users/matt/code/fpm/lib/fpm/package/rpm.rb:368:in `block in output'
    from /Users/matt/code/fpm/lib/fpm/package/rpm.rb:365:in `each'
    from /Users/matt/code/fpm/lib/fpm/package/rpm.rb:365:in `output'
    from /Users/matt/code/fpm/lib/fpm/command.rb:389:in `execute'
    from /Users/matt/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/clamp-0.6.1/lib/clamp/command.rb:67:in `run'
    from /Users/matt/code/fpm/lib/fpm/command.rb:436:in `run'
    from /Users/matt/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/clamp-0.6.1/lib/clamp/command.rb:125:in `run'
    from /Users/matt/code/fpm/bin/fpm:8:in `<main>'
```

When it is provided in a way FPM is happy with, rpmbuild fails to build as the %config block defines for example, %config(noreplace) /opt/app/opt/app/configs/whatever.ini

I'm not sure if this is a problem in OSX, rpmbuild, or FPM but it's definitely a strange one that I can't find the cause of.

PR #755, opened 2014-08-12
Handle clash between prefix and config-files

michaeltchapman:
Previously the path under staging was ignoring the
specified prefix when searching for config files,
leading to file not found errors.

Addresses #730 and #584


jordansissel:
Is this patch still needing to be merged? Sorry for the delays again.. Let me know - worst case, we can merge without tests and move forward. :)

hatt:
How does it compare to #931? That seems more generalised and can be easily extended to Pacman coverage on top of just Deb/RPM.

c-aamm:
Should this be closed?

PR #756, opened 2014-08-13
Add RakeTask

pstengel:
This allows you to easily use `fpm` in a Rakefile without a bunch of backticks or other non-Ruby ugliness. For example, given a Rakefile:

```ruby
require "fpm/rake_task"

FPM::RakeTask.new(:awesome_pkg, :source => :dir, :target => :tar) do |pkg|
  pkg.args = %w(bin/)
  pkg.verbose = true
  pkg.config_files = %w(foo bar)
end
```

Run `rake awesome_pkg`:

```
Created package {:path=>"awesome_pkg.tar"}
```

### DIFF
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -43,6 +43,7 @@ Gem::Specification.new do |spec|
   # For calling functions in dynamic libraries
   spec.add_dependency("ffi") # license: GPL3/LGPL3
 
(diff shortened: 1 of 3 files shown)


pstengel:
@jordansissel: Just pushed a commit that makes it 1.8.7 compatible and adds specs.

jordansissel:
Eyeballed the code - looks good! Thank you much :)

PR #762, opened 2014-09-01
Enable loading configuration from YAML document

benwebber:
This commit adds a `--settings` option that allows users to load
configuration options from a YAML document.

This is a much more flexible implementation than the current support for
`.fpm` rc files.

Example
-------

Here is an example settings file for FPM:

    ---
    source: gem
    target: deb
    maintainer: Jordan Sissel <jls@semicomplete.com>
    input:
      - fpm

Apply the settings by calling FPM like so:

    $ fpm --settings settings.yaml

Package options
---------------

Use nested block mappings to group package-specific options. This
settings file:

    ---
    python:
      install_bin: /usr/bin
      install_lib: /usr/local/lib/python2.7/site-packages

is equivalent to:

    ---
    python_install_bin: /usr/bin
    python_install_lib: /usr/local/lib/python2.7/site-packages

If there is a conflict between settings defined at the top level of the
document and in a nested block mapping, the latter definition wins.

That is, using this settings file:

    ---
    python:
      install_bin: /usr/bin
    python_install_bin: /usr/local/bin

the package will install Python executables to `/usr/local/bin`.

Command-line overrides
----------------------

Command-line options override all settings defined in the YAML document.

For example, if the settings file contained:

    ---
    source: gem
    target: deb
    gem:
      bin_path: /usr/local/bin

one could produce an RPM instead by running:

    $ fpm -t rpm --settings settings.yaml fpm

If the mandatory options (`source`, `target`, and `input`) are not
specified, FPM will prompt the user to supply `-s`, `-t`, and the input
arguments.

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -6,6 +6,7 @@ require "clamp"
 require "ostruct"
 require "fpm"
 require "tmpdir" # for Dir.tmpdir
+require "yaml"
 
 if $DEBUG
   Cabin::Channel.get(Kernel).subscribe($stdout)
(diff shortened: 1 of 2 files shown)


benwebber:
I'm closing this as I've found an `.fpm` file accomplishes the same goals for simple builds. Thanks for your patience, @jordansissel.

PR #815, opened 2014-11-20
Fixed RPM spec file to not include 2 %pre tags, but 1 %pre and 1 %post

bbruun:
A fix for enabling --before-upgrade and --after-upgrade in the template/rpm.erb file which included 2 %pre tags where the second should be %post.


hho:
I hit this as well. Duplicated by #830 (same fix, different bug description).

robertwatson3:
Why isn't this merged?

jordansissel:
I think https://github.com/jordansissel/fpm/commit/f6720c82d659bc44b52c833f5b447e375eadaeb2 fixed this, right?

josephholsten:
@bbruun I think @jordansissel is right, should be fixed in f6720c82d659bc44b52c833f5b447e375eadaeb2. I'm closing, but please reopen if you think I'm wrong. I'm often wrong.

Issue #803, opened 2014-11-07
--before-upgrade and --after-upgrade fails for RPM creation.


bbruun:
I'm testing FPM and love it's simplicity, but to be productive it needs to be able to upgrade RPM packages as well as just install them.

I've found that the build process fails if either or both of --before-upgrade and/or --after-upgrade parameters are present.

FPM version 1.3.0 and 1.3.2

Tested on ruby version 1.9.3 and 2.0.0

It would be nice that the documented feature also works so it is possible to maintain installations :-)

Example of error if either or both of --before-upgrade or --after-upgrade are present on the commandline:

Running the following command (with the before and after scripts in the scripts folder) produces an error and exits with exitcode 1

``` bash
$ fpm --verbose -s zip -t rpm -n akka-2.3.6 -v 0.1 --prefix /usr/local --template-scripts --before-install scripts/install/before.sh --after-install scripts/install/after.sh --before-upgrade scripts/upgrade/before.sh --after-upgrade scripts/upgrade/after.sh akka_2.11-2.3.6.zip
Setting workdir {:workdir=>"/tmp", :level=>:info}
Archive:  /var/tmp/akka-test-install/akka_2.11-2.3.6.zip {:level=>:info}
   creating: akka-2.3.6/ {:level=>:info}
   creating: akka-2.3.6/bin/ {:level=>:info}

<removed very verbose unzip output>

Converting zip to dir {:level=>:info}
Setting from flags: epoch= {:level=>:info}
Setting from flags: iteration= {:level=>:info}
Setting from flags: name=akka-2.3.6 {:level=>:info}
Setting from flags: url=http://example.com/no-uri-given {:level=>:info}
Setting from flags: version=0.1 {:level=>:info}
Converting zip to rpm {:level=>:info}
no value for epoch is set, defaulting to nil {:level=>:warn}
Reading template {:path=>"/usr/local/lib64/ruby/gems/1.9.1/gems/fpm-1.3.2/templates/rpm.erb", :level=>:info}
no value for epoch is set, defaulting to nil {:level=>:warn}
[...]

PR #822, opened 2014-12-09
Second %pre should be %post

dlanderson:
Fixes https://github.com/jordansissel/fpm/issues/803

### DIFF
--- a/templates/rpm.erb
+++ b/templates/rpm.erb
@@ -139,7 +139,7 @@ then
 fi
 <%   end -%>
 <%   if script?(:after_upgrade) or script?(:after_install) -%>
-%pre
+%post
 upgrade() {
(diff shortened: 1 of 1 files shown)


djhaskin987:
+1. Thanks for fixing this, my bad! :)

templates/rpm.erb
```diff
@@ -139,7 +139,7 @@ then
 fi
 <%   end -%>
 <%   if script?(:after_upgrade) or script?(:after_install) -%>
-%pre
+%post
```
jordansissel: doh! Good catch.

jordansissel:
Eyeballed; lgtm! I didn't test yet, but that's ok. :)

djhaskin987:
I tested this :+1:

jordansissel:
:heart:

PR #828, opened 2014-12-19
Add package metadata to slug file format

gerbercj:
We're starting to use the SH package format for a number of different project types. As part of this, we're adding some metadata to the packages so that we can easily determine how to deploy them. For example, a Rails project will want to roll unicorns, but a package of JavaScript assets would be installed differently. For other users, this is a no-op, with the exception of seeing a new `__METADATA__` line in the file, just prior to the `__ARCHIVE__` line.
@jordansissel

### DIFF
--- a/lib/fpm/package/sh.rb
+++ b/lib/fpm/package/sh.rb
@@ -11,7 +11,7 @@ require "digest"
 #
 # This class only supports output of packages.
 #
-# The sh package is a single sh file with a bzipped tar payload concatenated to the end.
+# The sh package is a single sh file with a tar payload concatenated to the end.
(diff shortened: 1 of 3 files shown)


gerbercj:
@jordansissel, any concerns with this enhancement to our package format?

jordansissel:
Nope! Patch reads OK to me. Any thoughts on tests?

gerbercj:
Totally reasonable request. Let me see what I can do for tests (in between shoveling breaks :snowflake:).

jordansissel:
Tests pass <3

jordansissel:
Sorry for the delay in attending to this. <3

PR #834, opened 2015-01-21
Ensure the post-pre install deb script functions aren't empty.

keithpitt:
I was getting errors locally because these functions were empty when a script hadn't been defined. Apparently there needs to be something inside them for these files to be valid bash?


jordansissel:
Sorry for the delays. Thank you for your efforts here! :)

I believe this was resolved by dfc3fac5 (#888)

PR #836, opened 2015-01-22
Fix up debian maintenance script templates

SpikesDivZero:
These templates are only used when --before-upgrade and --after-upgrade are used.


icarus1966:
i  need  all informations about  veeronika.michalekova@gmail.com slovakian girl.

mrijke:
Would be nice to see this merged, got bit by it today.

jordansissel:
@mrijke thanks for the poke, sorry about losing track of this. <3

SpikesDivZero:
Thanks @jordansissel and @mrijke! :)

PR #839, opened 2015-01-29
Adding guntar installation note for OS X

nfriedly:
I think this one's fairly self-explanatory, but it should save mac users a few minutes of googling around.


jordansissel:
<3

PR #842, opened 2015-01-30
Adding a deb-mode flag to allow control of the permissions

whpearson:
This simply passes on the flag to tar to control
how the files are stored in the data tar.

It will be useful in order to deploy things more securely.


lib/fpm/package/deb.rb
```diff
@@ -83,6 +83,9 @@ class FPM::Package::Deb < FPM::Package
 
   option "--group", "GROUP", "The group owner of files in this package", :default => 'root'
 
+  option "--mode", "MODE", "The mode modification to make to the files in this " \
+    "package. Use the same format as passing the mode tag into tar", :default => 'NOCHANGE'
```
jordansissel: Does this need a default value? If not specified in this definition (remove `:default => ...`), the default is `nil`. Unless ... "NOCHANGE" has some special meaning to tar?

whpearson:
Updated it to not use the NOCHANGE. I prefer not to have/use nils, but I understand this is not idiomatic ruby.

PR #844, opened 2015-02-03
Add --python-disable-dependency option to disable specific python dependencies

cure:
This is useful for example in the case of a dependency on the python
PyYAML package.

By default, this dependency gets translated to a debian package
dependency on python-pyyaml. That package does not exist on Debian
systems (it is called python-yaml), so a way to override the dependency is needed.

A similar option exists already for gems (--gem-disable-dependency).

### DIFF
--- a/lib/fpm/package/python.rb
+++ b/lib/fpm/package/python.rb
@@ -69,6 +69,10 @@ class FPM::Package::Python < FPM::Package
     "current python interpreter (sys.executable). This option is equivalent " \
     "to appending 'build_scripts --executable PYTHON_EXECUTABLE' arguments " \
     "to 'setup.py install' command."
+  option "--disable-dependency", "python_package_name",
+    "The python package name to remove from dependency list",
+    :multivalued => true, :attribute_name => :python_disable_dependencies,
+    :default => []
 
   private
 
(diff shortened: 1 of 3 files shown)


cure:
Ping? Any love for this pull request? I'd really like to upstream this functionality so I don't have to depend on my special fpm version.

Thanks,
Ward.

lib/fpm/package/python.rb
```diff
@@ -242,7 +249,7 @@ def load_package_info(setup_py)
         # convert dependencies from python-Foo to python-foo
         name = name.downcase if attributes[:python_downcase_dependencies?]
         "#{name} #{cmp} #{version}"
-      end
+      end.compact!
```
jordansissel: What does this fix?

lib/fpm/package/python.rb
```diff
@@ -228,6 +231,10 @@ def load_package_info(setup_py)
         end
         name, cmp, version = match.captures
 
+        if attributes[:python_disable_dependencies]
+          next if attributes[:python_disable_dependencies].include?(name)
```
jordansissel: the outer `if` here can be removed, I think, if you set the `:default => []` in the option

jordansissel:
I haven't tested this yet, and there are no tests for this. Can you write some tests for this?
Can someone test this before we merge it?

Otherwise, I left some comments inline.

cure:
Thanks for your comments. I've addressed them all, and added a number of tests. Does this look better?

cure:
ping? Sorry to keep bothering you - does the modified patch look better?

jordansissel:
I have not yet had time or energy to work on this.

jordansissel:
Alrighty! Code looks good. Tests pass. Thank you so much :)

I appreciate your patience and efforts! Sorry for the delays.

PR #849, opened 2015-02-10
Use MetaCPAN API query data as fallback for package metadata

BaxterStockman:
Some CPAN packages lack `META`/`MYMETA` files, which causes `fpm` to raise an error and exit.  However, the MetaCPAN query returns distribution metadata, so there's no need for the lack of a `META` file to cause a fatal error.

This pull request merges the MetaCPAN query data with whatever data can be read from a local `META` file, with the local data taking precedence.  If the `metadata` hash is empty after the merge, a fatal error is raised.


BaxterStockman:
On deeper exploration, it appears that some of the attributes that should appear in the `metadata` hash are not included in the results of the MetaCPAN query.  For instance, the 'license' field is associated with queries to `http://api.metacpan.org/v0/release`, but the code currently queries only `http://api.metacpan.org/v0/module`.

There are a couple of options for getting additional data:
- More `GET` requests to the MetaCPAN endpoints, using data gleaned from earlier requests to build up a fuller picture of a package's metadata; or
- Integrating the `elasticsearch` gem and using that to query MetaCPAN.

MetaCPAN's only rule is ['be polite'](https://github.com/CPAN-API/cpan-api/blob/master/docs/API-docs.md), which seems to be operationalized as 'If you're going to query more than 5000 modules, use Elasticsearch'.  I don't know what the likelihood is that `fpm` users will be making such large queries, so I'd tend to go with the first, simpler approach.  Thoughts?

jordansissel:
Reviewed the code - looks fine to me :)

As for your other concern for fetching additional data, I am happy to have that as well. If you wish this patch merged first and want to work on the extra data fetching later, I'm OK with that. Let me know :)

jordansissel:
Code looks ok; tests pass.

Merging!

Issue #843, opened 2015-02-03
INSTALL_BASE/--install-base not undefined for some CPAN modules


BaxterStockman:
`cpan.rb` deals with an issue in the `ExtUtils::MakeMaker` module by passing an empty `INSTALL_BASE` environment variable when running `Makefile.PL`/`Build.PL`.  *See*, e.g., lines 192-196 of `cpan.rb`:

```
safesystem(attributes[:cpan_perl_bin],
                     "-Mlocal::lib=#{build_path("cpan")}",
                     "Makefile.PL", "PREFIX=#{prefix}",
                     # Empty install_base to avoid local::lib being used.
                     "INSTALL_BASE=")
```

This seems to work nicely for most CPAN modules I've tried to build, but a few are naughty and misbehave.  One example is `Template::Toolkit`, which contains these lines in its `Makefile.PL`:

```
# read command line args putting TT_* into $ttconfig and
# everything else (regular Makefile.PL args, e.g. PREFIX)
# goes into $config

my (%config, %ttconfig);
while ($_ = shift) {
    my ($k, $v) = split(/=/, $_, 2);
    if ($k =~ /^TT/) {
        $ttconfig{ $k } = $v || 0;
    }
    else {
        $config{ $k } = $v || 0;
    }
};
```

The way this loop is written, the `%config` hash ends up looking like this:

```
%config = (
...
INSTALL_BASE => 0
...
)
```

because the right-hand side of the expression `$config{$k} = $v || 0` when `$v` is unset.  Later, the `%config` hash gets passed to the `ExtUtils::MakeMaker::WriteMakefile` subroutine, and from there gets propagated until eventually `ExtUtils::MM_Any::init_INSTALL` detects that both `INSTALL_BASE` and `PREFIX` are set, and dies with the message "Only one of PREFIX or INSTALL_BASE can be given.  Not both."  (Lines 1798-1800 in version 7.09 of `ExtUtils::MakeMaker`).

There are a number of ways to deal with this issue, some hackier than others:

- Do as the docs for `local::lib` suggest, and unset the `PERL_MM_OPT` environment variable when dealing with `Makefile.PL` or the `PERL_MB_OPT` environment variable when dealing with `Build.PL`.  This might be difficult to do while maintaining the same style of `safesystem` call shown above, since `local::lib` injects some environment variables directly into the Perl process rather than exporting them into the shell environment when `perl` is called with a filename.  For instance:

```
[...]

PR #850, opened 2015-02-10
cpan.rb - export the environment variables output by local::lib.

BaxterStockman:
See #843 -- `ExtUtils::MakeMaker` doesn't like it when both `PREFIX` and `INSTALL_BASE` are defined, and some CPAN packages don't play nice with the `INSTALL_BASE=` trick.  This pull request deals with this issue by exporting `local::lib`'s environment variables into the environment of the current process prior to calling `Build.PL`/`Makefile.PL`, skipping the `PERL_MM_OPT` and `PERL_MB_OPT` environment variables that were at the root of the problem.

### DIFF
--- a/lib/fpm/package/cpan.rb
+++ b/lib/fpm/package/cpan.rb
@@ -233,11 +233,10 @@ class FPM::Package::CPAN < FPM::Package
   end
 
   def export_local_lib_env(local_lib_path)
-    # Iterate over the output of local::lib
+    # Iterate over the output of local::lib line by line
     `#{attributes[:cpan_perl_bin]} -Mlocal::lib=#{local_lib_path}`.lines.each do |line|
       # local::lib's output is in the form:
-      # ENV_VAR="some value"; export $ENV_VAR;
-      #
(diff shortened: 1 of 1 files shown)


lib/fpm/package/cpan.rb
```diff
@@ -251,6 +240,27 @@ def input(package)
...
+      (k, v) = line.chomp.gsub(/;\s+export.*$/, "").split("=", 2)
+
+      # Skip these, since they wreak havoc when PREFIX # is set.
+      next if k == "PERL_MM_OPT" or k == "PERL_MB_OPT"
+
+      # Set the current environment in accordance with local::lib's output.
+      # The subshell call to echo is necessary because some environment
+      # variables are defined in terms of other such variables; this technique
+      # expands the variables-within-variables.
+      ENV[k] = `echo #{v}`.chomp
```
jordansissel: Thank you for the comment explaining why you do this subshell echo. I was super confused until I read the comment :heart:

BaxterStockman:
Fixed this up to be a bit saner -- no more subshell `echo`es, in other words :).

PR #851, opened 2015-02-10
cpan.rb -- avoid hangs by overriding interactive prompts with PERL_MM_USE_DEFAULT and AUTOMATED_TESTING

BaxterStockman:
Some CPAN modules default to requiring user input during the make/test/install process.  fpm does not hand control back to the user at such prompts, causing the build process to mysteriously hang.  Setting the `PERL_MM_USE_DEFAULT` and `AUTOMATED_TESTING` environment variables to `1` deals with the most common sources of such hangs.*

Although it might be preferable to allow interactive input under such circumstances, implementing that is likely to be much harder, and, in any case, if users wanted to do stuff interactively they probably wouldn't be using `fpm` ;).

* Confusingly, `Module::Build` also uses `PERL_MM_USE_DEFAULT`, rather than `PERL_MB_USE_DEFAULT`.

### DIFF
--- a/lib/fpm/package/cpan.rb
+++ b/lib/fpm/package/cpan.rb
@@ -60,7 +60,7 @@ class FPM::Package::CPAN < FPM::Package
       require "yaml"
       metadata = YAML.load_file(File.join(moduledir, ("MYMETA.yml")))
     else
-      raise FPM::InvalidPackageConfiguration, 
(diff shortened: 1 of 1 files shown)


jordansissel:
Hmm; this one now merge conflicts. Can you resolve that?

BaxterStockman:
Resolved, a mere 12 weeks later ;).

PR #853, opened 2015-02-11
--[no-]cpan-test now works as advertised

BaxterStockman:
There was a missing `!` in front of `attributes[:cpan_test]`, such that the `-n` (no test) flag was passed to `cpanm` if and only if `attributes[:cpan_test]` was `true`.  This inverts the proper logic: `cpanm` should skip tests if and only if `attributes[:cpan_test]` is `false`.


jordansissel:
This patch makes a bunch of whitespace changes :\

jordansissel:
I was able to merge this with some git trickery:

```
% git diff -w master..pull/853 | patch -p1
% git commit -a --author="Matt Schreiber <EMAIL>"
```
(The author info was taken from the git log from the PR)

jordansissel:
Merged! Thank you :)

Issue #857, opened 2015-02-18
config files external to gem


jakerobinson:
As a pattern, we keep config files in `/etc/my_app/`. When building an RPM from Gem, there does not seem to be a straightforward way to include the files in `/etc/my_app`.

I've included our entire workflow so you can see what we're up to. Am I missing an easy option somewhere?

```
 # Our app runs in jruby, so some dependencies won't download without it
rvm use jruby 1.7.18
gem install my_app --no-document --install-dir .

# fpm doesn't seem to work with jruby, and pleaserun requires >2.0 (due to mustache)
rvm use ruby 2.1.2 

# Generate init scripts
pleaserun -p sysv -v lsb-3.1 --install --install-prefix . my_app

# Copy our default configs from the within our gem
mkdir -p ./etc/my_app
cp gems/my_app-*/config/* etc/my_app/

# copy the gem a second time to build the RPM a second time including the init and config files. fpm does something to the gem that does not allow us to build it from the same gem file.
cp ./cache/my_app*.gem .

# build rpms for my_app and all dependencies
find ./cache -name '*.gem' | xargs -rn1 fpm --prefix /opt/jruby-1.7.18/lib/ruby/gems/shared --gem-bin-path /opt/jruby-1.7.18/bin -s gem -t rpm 

# remove the RPM that does not have init and conf files
rm my_app*.rpm

# build RPM with init and config files. config files do not currently work 
fpm -s gem \
        -t rpm \
       --prefix /opt/jruby-1.7.18/lib/ruby/gems/shared \
       --gem-bin-path /opt/jruby-1.7.18/bin \
       --config-files ./etc/my_app/service.yaml 
       --config-files ./etc/my_app/logging.yaml 
       --rpm-init ./etc/init.d/my_app my_app*.gem

#Copy RPMs to server and run createrepo
```

```
/home/jenkins/.rvm/gems/ruby-2.1.2/gems/fpm-1.3.3/lib/fpm/package/rpm.rb:456:in `block in output': Config file path /tmp/package-gem-staging20150218-38342-16szg8c/etc/my_app/service.yaml does not exist (RuntimeError)
    from /home/jenkins/.rvm/gems/ruby-2.1.2/gems/fpm-1.3.3/lib/fpm/package/rpm.rb:454:in `each'
    from /home/jenkins/.rvm/gems/ruby-2.1.2/gems/fpm-1.3.3/lib/fpm/package/rpm.rb:454:in `output'
[...]

PR #860, opened 2015-02-19
copy config files external to source

jakerobinson:
proposed fix to #857 

Allows the addition of configuration files external to source.

e.g. We want config files for a ruby gem to be located in `/etc/my_app`, which is outside of the gem package. This works very similar to the way init files are added.

@jordansissel Looking for guidance on writing tests and testing regression.


Suirtimed:
@jordansissel is there any reason this hasn't been merged in?  I'd like to get this in the "current" version and start working on the next issue I've run into related to getting a changelog file passed in.

ryantanner:
Would love to have this.

josephholsten:
fixed with the rebase in #1379. Thanks!

Issue #856, opened 2015-02-13
`fpm -s pleaserun ...` support


jordansissel:
Time to make fpm and pleaserun usable together. It would be cool to have this:

```
fpm -s pleaserun -t rpm --pleaserun-platform upstart /some/program
```

And have it generate an rpm with the correct post-install steps + files to install an upstart task that runs /some/program.

jakerobinson:
YES. :)

nicocesar:
+1

jordansissel:
`fpm -s pleaserun ...` was added in May 2016 -

Initial commit was here: 6881690bfc5e3105e2bfa5ab25c1bb287acf853e

PR #862, opened 2015-02-24
add basic systemd service option

rhelmer:
It'd be nice to have an option as simple as e.g. `--rpm-init` for installing systemd services, especially when using source types other than `dir`, like `python`.

The service unfortunately needs to be activated/deactivated separately (I am doing this in `after-install` and `after-remove`), but it's pretty simple at least and works with all source types.


phrawzty:
:+1:

jordansissel:
Any chance you're interested in waiting on https://github.com/jordansissel/fpm/issues/856 or similar implementation?

rhelmer:
@jordansissel pleaserun integration would be great!

Incidentally, it turns out I can't really use this for what I was planning, since I need to drop in config files to ```/etc``` and also set them as config files... I can't figure out a way to do this with ```python``` source type so I dropped back to using the ```dir``` source type.

jordansissel:
For clarity on #856, my intention is to have pleaserun fully supported in fpm.

At first implementation, this will require you to make two packages. One for your app, one using pleaserun support. Second phase is some way to chain package creation to say "Package X is python Django + my app directory + pleaserun systemd"

josephholsten:
@jordansissel @rhelmer I'm guessing it's not valuable for me to rebase this, would be better to implement #856, right? Should we close this?

jordansissel:
@josephholsten agreed. pleaserun support was added in May 2016; I forgot to close #856 (just did now).

jordansissel:
fpm supports `-s pleaserun` which lets you create a package with a service definition (systemd, upstart, etc) that is automatically detected at package-install time -- I think we can close this

PR #863, opened 2015-02-24
symlinks need full dest path for FileUtils.copy_entry

fetep:
From http://ruby-doc.org/stdlib-2.2.0/libdoc/fileutils/rdoc/FileUtils.html#method-c-copy_entry:

```
Both of src and dest must be a path name. src must exist, dest must not exist.
```

(tested & working)


jordansissel:
I'd like to see tests for this, but we can do that later. Code LGTM.

Issue #697, opened 2014-05-05
Support virtualenvs for python package building


josegonzalez:
FPM currently requires separate packages for every piece of a given python package. Using the [`beaver`](https://github.com/josegonzalez/beaver) repository, you would need to create the initial beaver package, as well as iterate over each line of the requirements file and create packages for those. 

The above works great if you do not also distribute other fpm-based python packages on the server with conflicting requirements. If you do, then tough shit.

A better way would be to build a [python virtualenv](https://pypi.python.org/pypi/virtualenv) for the package. Thus we'd only have to install `python-beaver` to get everything up and running. This is contrary to the debian way -lawl - but is more immediately useful to those attempting to run services in production.

[dh-virtualenv](http://dh-virtualenv.readthedocs.org/en/latest/) does something similar - though is python specific - and it doesn't create binaries for scripts like fpm does without need to learn more about debian packaging.

Issue #636 somewhat touches on this, though without any concrete outcome.

neoice:
+1

I tried to package some Python webapps with virtualenvs in early 2013 and was not successful due to symlinks/PATHs within the virtualenvs. I'm revisiting the problem now and I'm bummed that it hasn't been resolved yet.

there's a number of issues on the FPM tracker that this feature would solve:

https://github.com/jordansissel/fpm/issues/262
https://github.com/jordansissel/fpm/issues/780
https://github.com/jordansissel/fpm/issues/636 (maybe)

I could also add to @thedrow's bounty (charity or BTC)

thedrow:
ping @jordansissel

thedrow:
We can use https://github.com/PolicyStat/terrarium instead of virtualenvwrapper-tools.
It has a much better API and you can simply download the compressed virtualenv from somewhere and use the CLI to uncompress it.

evilsocket:
Pull request waiting :) https://github.com/jordansissel/fpm/pull/864

jravetch:
+1  Been using dh-virtualenv as a substitute until fpm gets this feature.

djhaskin987:
If no one objects, I'll start on this feature based on the script found here --> https://github.com/brutasse/graphite-api/blob/master/fpm/build-deb.sh . It seems pretty straight forward; expect a pull request soon.

thedrow:
It's exactly what @evilsocket did but for some reason he could not inherit from the FPM::Package::Dir and other problems as well.
Maybe you guys can join forces in order to finish this.

[...]

PR #864, opened 2015-02-25
New virtualenv source module.

evilsocket:
I've created a python virtualenv source module, fixing issue #697

### DIFF
--- a/lib/fpm/package/virtualenv.rb
+++ b/lib/fpm/package/virtualenv.rb
@@ -1,0 +1,217 @@
+require "fpm/package"
+require "fpm/util"
+require "backports"
+require "fileutils"
+require "find"
+require "socket"
+
+# A virtualenv package.
+class FPM::Package::VirtualEnv < FPM::Package
+    option "--installation-path", "DIRECTORY", "The directory to install the final environment."
+
+    def input(path)
+        raise ArgumentError, "Parameter --virtualenv-installation-path is mandatory." if attributes[:virtualenv_installation_path].nil?
+
+        dest_path = attributes[:virtualenv_installation_path]
+        path = File.expand_path path
+        venv = get_virtualenv path
+        build_folder = "build_#{Time.now.to_i}"
+
+        raise ArgumentError, "Path '#{path}' does not contain a valid virtualenv." if venv.nil?
+
+        root_folder = "#{build_folder}#{File.dirname dest_path}"
+        final_folder = "#{build_folder}#{dest_path}"
+
+        logger.info( "Creating temporary build folder #{build_folder} ..." )
+
+        # copy the python project into a temporary one
+        FileUtils.mkdir_p root_folder
+        FileUtils.cp_r path, final_folder
+
+        logger.info( "Updating virtualenv paths ..." )
+
+        # update the virtualenv to point to the installation folder
+        venv = get_virtualenv "#{final_folder}"
+        ::Dir.chdir(venv) do
+            safesystem( "virtualenv-tools --update-path #{dest_path}" )
+        end
+
+        # remove precompiled python files
+        remove_python_compiled_files final_folder
+
+        path = '.'
+        chdir = build_folder
+        source = path
+        destination = "/"
+
+        if attributes[:prefix]
+          destination = File.join(attributes[:prefix], destination)
+        end
+
+        destination = File.join(staging_path, destination)
+
+        logger.info( "Creating source '#{destination}' ..." )
+
+        logger["method"] = "input"
+        begin
+          ::Dir.chdir(chdir) do
+            begin
+              clone(source, destination)
+            rescue Errno::ENOENT => e
+              raise FPM::InvalidPackageConfiguration,
+                "Cannot package the path '#{File.join(chdir, source)}', does it exist?"
+            end
+          end
+        rescue Errno::ENOENT => e
+          raise FPM::InvalidPackageConfiguration,
+            "Cannot chdir to '#{chdir}'. Does it exist?"
+        end
+
+        # Set some defaults. This is useful because other package types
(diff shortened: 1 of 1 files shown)


thedrow:
Woot!

lib/fpm/package/virtualenv.rb
```diff
@@ -0,0 +1,217 @@
...
+    # This is special because it respects the full path of the source.
+    # Aditionally, hardlinks will be used instead of copies.
+    #
+    # Example:
+    #
+    #     clone("/tmp/hello/world", "/tmp/example")
+    #
+    # The above will copy, recursively, /tmp/hello/world into
+    # /tmp/example/hello/world
+    def clone(source, destination)
```
jordansissel: I've ... see this code before :P 

I think you can subclass Dir instead of copying the code:

```diff
-class FPM::Package::VirtualEnv < FPM::Package
+class FPM::Package::VirtualEnv < FPM::Package::Dir
```

jordansissel: Subclassing Dir will remove lots of copy/pasted code and should clean up this plugin a bit :)

lib/fpm/package/virtualenv.rb
```diff
@@ -0,0 +1,217 @@
+require "fpm/package"
+require "fpm/util"
+require "backports"
+require "fileutils"
+require "find"
+require "socket"
+
+# A virtualenv package.
+class FPM::Package::VirtualEnv < FPM::Package
+    option "--installation-path", "DIRECTORY", "The directory to install the final environment."
```
jordansissel: I think we can just reuse fpm's `--prefix` flag here, instead of adding a new flag. Thoughts?

lib/fpm/package/virtualenv.rb
evilsocket: I tried of course, but I got a few errors

evilsocket: When subclassing FPM::Package::Dir I get:

    ERROR: Unrecognised option '--virtualenv-installation-path'

The 

    option "--installation-path", "DIRECTORY", "The directory to install the final environment."

Somehow is not recognized ....

lib/fpm/package/virtualenv.rb
evilsocket: uhm yeah why not, sounds good :)

lib/fpm/package/virtualenv.rb
evilsocket: And:

    Invalid input package -s flag) type "virtualenv". Expected one of: cpan, deb, dir, empty, gem, npm, osxpkg, pear, pkgin, puppet, python, rpm, sh, solaris, tar, zip {:level=>:warn}

thedrow: If we use the ```--prefix``` option I don't think that the first bug matters much but we should open a new issue about it.
The second error is wierd and is definately a problem that we should solve before merging this PR.

thedrow:
@jordansissel Care to help us fix the errors?

neoice:
I've checked out @evilsocket's fork to give this a try... it didn't go well.

```
mkdir ~/virtualenv
cd ~/virtualenv
virtualenv .
source bin/activate
pip install some-python-package
```

`fpm --debug --virtualenv-installation-path /opt -s virtualenv -t rpm ~/virtualenv `

=>

`Path '/home/neoice/virtualenv' does not contain a valid virtualenv. (ArgumentError)`

additionally, based on [this line](https://github.com/evilsocket/fpm/blob/master/lib/fpm/package/virtualenv.rb#L13), I would expect to get an ArgumentError. instead I get:

```
in `path': no implicit conversion of nil into String (TypeError)
```

am I using this source incorrectly or does the software need some additional polish?

lib/fpm/package/virtualenv.rb
djhaskin987: I believe this happens because of @jordansissel 's metaclass code. From the docs for ruby's `Class`:
>  inherited(subclass)
>
> Callback invoked whenever a subclass of the current class is created.

But Jordan uses a bunch of metaclass `class << self` magic to get the options to work right with Clamp. The code he wrote is awesome, but I suspect it won't work across inheritance.

This means I think you can't just inherit from another package class and have everything work. @jordansissel please disabuse me if you can of this idea, I hope I'm wrong.

A solution I have come up with is to move the function `copy_metadata` from the `Dir` class and put it in the `util` FPM module, so that we can use both it and `copy_entry` to copy files properly.

However, I think for the purposes of the present problem of reusing `Dir` code, we should simply do what the `FPM::Package::Tar` class does:
```
    # use dir to set stuff up properly, mainly so I don't have to reimplement
    # the chdir/prefix stuff special for tar.
    dir = convert(FPM::Package::Dir)
    if attributes[:chdir]
      dir.attributes[:chdir] = File.join(build_path, attributes[:chdir])
    else
      dir.attributes[:chdir] = build_path
    end

    cleanup_staging
    # Tell 'dir' to input "." and chdir/prefix will help it figure out the
    # rest.
    dir.input(".")
    @staging_path = dir.staging_path
    dir.cleanup_build
```
It uses *composition* rather than *inheritance* here, arguably a better strategy in general when it can be used.

thedrow: Yeh that should work.
@djhaskin987 If you'd like you can create a PR against @evilsocket's branch and we'll proceed from there.

djhaskin987: I actually wrote my own code before I knew about @evilsocket's PR. I edited that code to add some code from @evilsocket's PR and made my own PR, #930.

thedrow:
@jordansissel @evilsocket You guys should probably close this one since #697 is already resolved by #930 already resolved this issue.

PR #865, opened 2015-02-25
Default to HTTPS for --python-pypi URL

timsutton:
Here's a PR for changing the default PyPi base URL for --python-pypi. I've tested this with both `pip` and `easy_install` on OS X.

Thanks for FPM!


jordansissel:
lgtm!

PR #869, opened 2015-03-04
Negates "implicit *" behavior when --inputs is given.

tai:
Hi. I was hit by following behavior of fpm, and think attached patch can improve it.

With following command, "--inputs file.list" part is essentially ignored as "." is implicitly added to args when -C dir is given.

    fpm -s dir -C workdir --inputs file.list -t deb -n mypkg

This is because fpm adds "."  implicitly when -C dir is given and command-line arg is empty. But this is kind of unexpected when --inputs is also used.

FYI, I am using fpm for automatic package generation from workdir, which is wrapped by a union filesystem (aufs). So  just need to do "make install" (or whatever package requires) and let unionfs pick out actually written files for packaging.

For many cases, I try to minimize files to package, so I do

    (cd unionfs-backend-dir; find -type f) > wanted.txt
    $EDITOR wanted.txt
    fpm -s dir -C unionfs-backend-dir --inputs wanted.txt ...

I can't just delete "unwanted" files and use ".", as it may turn out that I actually do need them after trial-and-error. Without this fix, I need to workaround by doing

    head -n -1 wanted.txt | fpm -C unionfs-managed-dir --inputs /dev/stdin ... $(tail -1 wanted.txt)

which is somewhat kludgy.

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -271,9 +271,11 @@ class FPM::Command < Clamp::Command
     #   'fpm -s dir -t ... -C somepath'
     # fpm would assume you meant to add '.' to the end of the commandline.
     # Let's hack that. https://github.com/jordansissel/fpm/issues/187
+    args_assumed = false
     if input_type == "dir" and args.empty? and !chdir.nil?
       logger.info("No args, but -s dir and -C are given, assuming '.' as input") 
       args << "."
+      args_assumed = true
     end
 
     logger.info("Setting workdir", :workdir => workdir)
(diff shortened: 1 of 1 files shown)


tai:
I wanted to avoid adding "args_assumed" flag, but couldn't as validation done by a plugin checks for empty args.

thatothermitch:
I'm struggling with this sort of problem too. This type of change would resolve it in my eyes.

thatothermitch:
As a workaround, I had some success abandoning "--inputs," and instead catting the file I was using for input as path parameters. Ex: `fpm ... --inputs /path/to/input-file` became `fpm ... $(cat /path/to/input-file)`

tai:
Well, I can also live with that for now. I guess ARG_MAX is big enough these days...

lib/fpm/command.rb
```diff
@@ -271,9 +271,11 @@ def execute
     #   'fpm -s dir -t ... -C somepath'
     # fpm would assume you meant to add '.' to the end of the commandline.
     # Let's hack that. https://github.com/jordansissel/fpm/issues/187
+    args_assumed = false
```
jordansissel: I think we can do this an easier way, I'll try something.

jordansissel:
@tai I do agree this is a bug. I'll see what I can do to write tests to reproduce it (and verify it stays fixed in the future).

jordansissel:

PR #871, opened 2015-03-06
Fix syntax error in Debian prerm template

mrijke:
Bash functions may not be empty (http://tldp.org/LDP/abs/html/functions.html) otherwise this error is thrown:

```
prerm: line 9: syntax error near unexpected token `}'
prerm: line 9: `}'
```

Instead, put a single colon (':') in the function body as a NULL operator.


mrijke:
Didn't see #836 at first, closing this.

PR #874, opened 2015-03-13
Add command line flag for --exclude-file.

jalawrence:
Works like a series of `--exclude`s; sort of a counterpart to --include.

The rationale is to allow a file under developer control to specify exclusion patterns separate from fpm execution, which is under CI control (in our world).

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -140,7 +140,6 @@ class FPM::Command < Clamp::Command
     next excludes
   end # -x / --exclude
 
-  # jal 2015.03.13
   option "--exclude-file", "EXLCUDE_PATH",
     "The path to a file containing a newline-sparated list of "\
     "patterns to exclude from input."
@@ -352,7 +351,6 @@ class FPM::Command < Clamp::Command
       end
     end
 
-    # jal
     # If --exclude-file was specified, read it as a file and append to
     # the exclude pattern list.
     if !exclude-file.nil?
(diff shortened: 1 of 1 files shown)


ahammond:
+1 this would be a nice feature

lib/fpm/command.rb
```diff
@@ -346,6 +351,22 @@ def execute
       end
     end
 
+    # If --exclude-file was specified, read it as a file and append to
+    # the exclude pattern list.
+    if !exclude-file.nil?
```
jordansissel: You probably meant `exclude_file` not `exclude-file` (probably confusing if you're looking at the flag name, but `exclude-file` in ruby means 'subtract `file` from `exclude` variable')

lib/fpm/command.rb
```diff
@@ -346,6 +351,22 @@ def execute
...
+      if !File.exists?(exclude-file)
```
jordansissel: same here, `exclude_file`

lib/fpm/command.rb
```diff
@@ -139,6 +139,11 @@ def help(*args)
     excludes << val
     next excludes
   end # -x / --exclude
+
+  option "--exclude-file", "EXLCUDE_PATH",
```
jordansissel: typo, fix: `EXCLUDE_PATH`

jordansissel:
Small typos in the code. Otherwise this looks OK to me. Let me know when you've fixed the typos and I'll take another look!

Thank you for helping make fpm more awesome :)

jordansissel:
Merged with #906

PR #876, opened 2015-03-17
Enable packaging Perl modules from Github

humppa:
While CPAN is used for publishing Perl modules, Github nowadays hosts the code of many modules. This patch allows *fpm* to create packages from test or development code not yet published in CPAN, but found from Github.

Three package identifier formats are supported:
+ Github:user/repository
+ git@github.com:user/repository.git
+ https://github.com/user/repository.git

Example usage:
    fpm -s cpan -t deb https://github.com/plack/Plack.git
 
Currently only master branch is supported.

### DIFF
--- a/lib/fpm/package/cpan.rb
+++ b/lib/fpm/package/cpan.rb
@@ -39,8 +39,13 @@ class FPM::Package::CPAN < FPM::Package
     require "net/http"
     require "json"
 
+    user, repo = fromGithub(package)
+
     if (attributes[:cpan_local_module?])
       moduledir = package
+    elsif user && repo
+      tarball = octopus(user, repo)
+      moduledir = unpack(tarball)
     else
       result = search(package)
       tarball = download(result, version)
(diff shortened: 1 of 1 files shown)


jordansissel:
Thank you for your contribution :)

I wonder - if we built in better git support as a kind of utility library within fpm, then we could allow almost any package source (gem, python, perl, etc) to be built from Github (or gitlab, etc).

Any thoughts on the idea? I'm not requiring any work, just brainstorming a way to have all (or most) package types benefit from a "build from github"-style of thing.

rjeschmi:
I imagine there are a lot of packages that aren't dist ready in github (I know mine just has a dist.ini, no makefile.pl)

It is a good idea though.

jordansissel:
@rjeschmi I think most ruby libraries are ready-from-source. Some of the ones that use the `hoe` library and I think `nokogiri` itself isn't, but the rest probably are.

To note, I won't block this PR while we wait on the "fetch from some source repository" idea we are discussion. Will review shortly.

humppa:
@rjeschmi Good point. I noticed that many packages don't include `Makefile.PL` or `META.*` files. Luckily there's no reason why *cpan.rb* could not support Dist::Zilla and I already have a todo item on that (i.e. I'll fix it as soon as I find time for it).

@jordansissel I like the idea of git utility library and I'd love to implement it too, but I haven't got spare time at least for a few days.

I recommend to leave this PR hanging for a while until I or someone else comes up with something better.

rjeschmi:
@humppa Seems like you are on top of it. I wasn't suggesting it is a blocking issue, but interesting none-the-less. I think it is probably a bug for the repos and might clean my own up so it builds.

Some of the repos I found have some sort of hack to post-build copy the make or build file back into the root.

As a reference: https://github.com/rjbs/Dist-Zilla does not.

humppa:
@rjeschmi I'm sorry, I'm not sure if I understood you correctly here.

Nevertheless, I believe it is customary and intended only to have `dist.ini` or some other minimal set of files committed in repository. It seems, that the tools support it, and it is so common, that *fpm* should support it too.

I'll get back when I have written something useful.

rjeschmi:
@humppa Yeah, it is customary, but I think it is a practice that frustrates testing, thus counter-productive. In any case it isn't an fpm bug, though supporting building using Dist::Zilla is obviously going to handle more cases.

jordansissel:
This PR fails for me, but I think it's local misconfiguration (I'm missing local/lib.pm?)

```
% fpm --debug -s cpan -t deb https://github.com/plack/Plack.git
Running command {:args=>["perl", "-Mlocal::lib=/var/folders/4r/4fpyt_2d1qvfrgb2fnngrqm40000gn/T/package-cpan-build20150507-22740-9wk4iq/cpan", "Makefile.PL", "PREFIX=/usr/local", "INSTALL_BASE="], :level=>:debug, :file=>"fpm/util.rb", :line=>"59", :method=>"safesystem"}
Process is running {:pid=>30823, :level=>:debug, :file=>"fpm/util.rb", :line=>"72", :method=>"safesystem"}
Can't locate local/lib.pm in @INC (you may need to install the local::lib module) (@INC contains: /Library/Perl/5.18/darwin-thread-multi-2level /Library/Perl/5.18 /Network/Library/Perl/5.18/darwin-thread-multi-2level /Network/Library/Perl/5.18 /Library/Perl/Updates/5.18.2 /System/Library/Perl/5.18/darwin-thread-multi-2level /System/Library/Perl/5.18 /System/Library/Perl/Extras/5.18/darwin-thread-multi-2level /System/Library/Perl/Extras/5.18 .). {:level=>:info, :file=>"cabin/mixins/pipe.rb", :line=>"47", :method=>"block in pipe"}
BEGIN failed--compilation aborted. {:level=>:info, :file=>"cabin/mixins/pipe.rb", :line=>"47", :method=>"block in pipe"}
Process failed: perl failed (exit code 2). Full command was:["perl", "-Mlocal::lib=/var/folders/4r/4fpyt_2d1qvfrgb2fnngrqm40000gn/T/package-cpan-build20150507-22740-9wk4iq/cpan", "Makefile.PL", "PREFIX=/usr/local", "INSTALL_BASE="] {:level=>:error, :file=>"fpm/command.rb", :line=>"468", :method=>"rescue in execute"}
```

jordansissel:

Issue #873, opened 2015-03-12
Files in `/etc` should be registered as config files by default


vincentbernat:
Hi!

When building Debian packages with debhelper, files in `/etc` are automatically marked as configuration files (this is done since compatibility level 3 which was introduced in 2000). Therefore, packages built with fpm may come as a surprise since they don't do that. A user can still add `--config-files /etc` but they have to know that (usually after a valuable modification has been erased), hence I think this should be better to have that by default.

If there is an agreement on that, I can do a PR to implement that.

jordansissel:
As long as there's a flag to disable this behavior, I'm OK with having a default `--config-files /etc` for deb platforms.

jordansissel:
As this will be a behavior-breaking change, part of the work to do here is to advertise this change loudly enough to existing users who will be upgrading.

dka23:
Seems to be a duplicate to #589

PR #877, opened 2015-03-18
For Debian packages, register files in `/etc` as conf files

vincentbernat:
When building Debian packages with debhelper, files in `/etc` are
automatically marked as configuration files (this is done since
compatibility level 3 which was introduced in 2000). Therefore, packages
built with fpm may come as a surprise since they don't do that. A user
can still add `--config-files /etc` but they have to know that (usually
after a valuable modification has been erased).

This change add `/etc` to the list of configuration files for Debian
unless fpm is invoked with `--no-config-files`.

Closes: #873 

Unlike what I was told in #873, I didn't add any warning for the behavior change. How should I add that?

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -124,6 +124,10 @@ class FPM::Package::Deb < FPM::Package
     next @custom_fields
   end
 
+  option "--no-default-config-files", :flag,
+    "Do not add all files in /etc as configuration files by default for Debian packages.",
+    :default => false
+
   option "--shlibs", "SHLIBS", "Include control/shlibs content. This flag " \
     "expects a string that is used as the contents of the shlibs file. " \
     "See the following url for a description of this file and its format: " \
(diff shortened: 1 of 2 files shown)


jordansissel:
@vincentbernat I think we can have a `logger.warn` when we are activating the default within fpm, basically telling the user something like "Debian packaging policy generally labels all files in /etc as config files, so fpm defaults to this behavior for deb packages. You can disable this default behavior with --no-config-files flag"

lib/fpm/command.rb
```diff
@@ -117,6 +117,9 @@ def help(*args)
     "configuration files, specify this flag multiple times.  If argument is " \
     "directory all files inside it will be recursively marked as config files.",
     :multivalued => true, :attribute_name => :config_files
+  option "--no-config-files", :flag,
```
jordansissel: Given this is a debian-specific patch, I'm have hesitations on having it listed as a default flag and not deb-specific. Thoughts?

lib/fpm/package/deb.rb
```diff
@@ -641,29 +641,39 @@ def write_scripts
...
+    end
+
+    # scan all conf file paths for files and add them
+    config_files.each do |path|
       begin
-        Find.find(cfg_path) do |p|
-          allconfigs << p.gsub("#{staging_path}/", '') if File.file? p
-        end
-      rescue Errno::ENOENT => e
+        add_path path, allconfigs
```
jordansissel: use parentheses - `add_path(path, allconfigs)`

lib/fpm/package/deb.rb
```diff
@@ -641,29 +641,39 @@ def write_scripts
...
     allconfigs = []
-    config_files.each do |path|
+
+    # expand recursively a given path to be put in allconfigs
+    def add_path(path, allconfigs)
       # Strip leading /
       path = path[1..-1] if path[0,1] == "/"
       cfg_path = File.expand_path(path, staging_path)
+      Find.find(cfg_path) do |p|
+        allconfigs << p.gsub("#{staging_path}/", '') if File.file? p
```
jordansissel: Since `Find.find` returns an `Enumerator`, I think moving the conditional `File.file?` to a `select` would make it easier to read:

```ruby
Find.find(cfg_path).select { |path| File.file?(path) }.each do |path|
  allconfigs << path.gsub("#{staging_path}/", "")
end
```

lib/fpm/package/deb.rb
```diff
@@ -641,29 +641,39 @@ def write_scripts
...
 
   def write_conffiles
-    return unless config_files.any?
-
-    # scan all conf file paths for files and add them
     allconfigs = []
-    config_files.each do |path|
+
+    # expand recursively a given path to be put in allconfigs
+    def add_path(path, allconfigs)
```
jordansissel: I'm not sure how I feel about passing the `allconfigs` in for the purposes of modification, hmm... Seems like this would be better moved into a separate class where you do configs.add_path(path) instead.

I'm OK if your code stays as-is, for now, while I ponder this :)

lib/fpm/package/deb.rb
```diff
@@ -641,29 +641,39 @@ def write_scripts
...
+        add_path path, allconfigs
+      rescue Errno::ENOENT
         raise FPM::InvalidPackageConfiguration,
           "Error trying to use '#{cfg_path}' as a config file in the package. Does it exist?"
       end
     end
+
+    # Also add everything in /etc
+    begin
+      add_path '/etc', allconfigs unless attributes[:no_config_files?]
```
jordansissel: We should log that we are doing this, as you asked about the warning about the behavior change, maybe somethign like:

```ruby
# Written assuming '--no-config-files' stays as the flag name
if !attributes[:no_config_files?]
  logger.warn("Something about what we are doing, and why....")
  add_path("/etc", allconfigs)
end
  ```

File: spec/fpm/package/deb_spec.rb [hunk 6]
```diff
@@ -366,7 +366,6 @@ def dpkg_field(field)
       @deb.attributes[:deb_user] = "root"
       @deb.attributes[:deb_group] = "root"
 
-      @deb.instance_variable_set(:@config_files, ["/etc/init.d/test"])
```
jordansissel: :heart:

jordansissel:
Thank you for helping fpm build better things by default :)

I did one pass at reviewing the code. Overall, it looks great! I made some comments on naming and style and some other things.

lib/fpm/command.rb
vincentbernat: At first, I thought that it would be the same for RPM, but all spec files seem to have to declare the files as a configuration file (with `%config`). So, I will change the option to be Debian specific.

vincentbernat:
I have updated the PR with the requested changes: use of `--deb-no-config-files`, various coding style. I left the `add_path` method as is. Using a separate class may be a bit overkill. There is a spurious commit that I have also pushed into a separate PR (#890) to make unittests work.

jordansissel:
@vincentbernat looks good!

One last question, I hope! I was thinking since this is specific to /etc and "automatic" or "default" behavior, maybe the flag should be --no-default-config-files ? `-no-config-files` seems too blunt of a phrasing when we mean something specific.

Otherwise, LGTM!

vincentbernat:
Just updated the PR with `--no-default-config-files`. This makes sense.

jordansissel:
LGTM

PR #878, opened 2015-03-25
Add --rpm-unit switch to add systemd unit files

delhage:
Added --rpm-unit which works similar to --rpm-init but for systemd unit files.


delhage:
I see now that this is basically the same change as #862. Closing

Issue #587, opened 2013-11-19
Support Wheel python binary format


thedrow:
See [here](https://pypi.python.org/pypi/wheel) for details about wheel.

Basically the feature should allow you to download a package from PyPi and turn it into a wheel.

mcallaway:
+1

itsgrimetime:
is there any plan for this to be supported? It is preventing me from packaging mxnet, mxnet-cu80, or tensorflow-gpu:

https://github.com/tensorflow/tensorflow/issues/6540

itsgrimetime:
I'm interested in implementing this, as I'm currently unable to repackage mxnet, mxnet-cu80, or tensorflow-gpu from their pip packages on pypi: 

https://github.com/tensorflow/tensorflow/issues/6540

Based on the description here: https://pypi.python.org/pypi/wheel it seems that the .whl format is (in most cases) simply a zip file. Would an alright approach be unzipping this wheel (or installing it via `pip install`) and building any other packaging type as if the source was a zip?

jordansissel:
Your proposal sounds like a good start!

iFreilicht:
I've looked into this a little bit as well, I also want to repackage mxnet-cu80, which doesn't have a source package on pypi.

> Would an alright approach be unzipping this wheel (or installing it via pip install) and building any other packaging type as if the source was a zip?

Almost. You also need to generate a new `setup.py` file, as that is not distributed with a wheel. To do this properly, you will need to parse the data contained in the `.dist-info` and `.data` directories inside the wheel. A source package had different files for that. Apart from this, there seems to be no difference between a source package and a wheel, as can be seen in [PEP427](https://www.python.org/dev/peps/pep-0427/#file-format).

To parse the metadata properly, you'll need to look into [PEP426-2.0](legacy.python.org/dev/peps/pep-0426) and [PEP426-3.0](https://www.python.org/dev/peps/pep-0426/)(The title is incorrect here!). The `*.dist-info/metadata.json` contains most information needed, but for some special functionality like installable scripts, you will need to look at the `*.data` directory as well, which is defined in PEP427, linked above.

I've written a python script that can generate a minimal `setup.py` file. That could be used as a starting point. If you want, I can discuss with my employer whether I can release it.

etano:
Hi @iFreilicht , I'd be interested in this script. Trying to distribute as rpm's and tensorflow (for example) is only distributed as a wheel.

neeasade:
[...]

PR #885, opened 2015-04-07
Until python wheel format is supported, tell pip to --no-use-wheel

mcallaway:
See https://github.com/jordansissel/fpm/issues/587.

When using fpm to package python modules using pip, we need to avoid "wheel" format until it is explicitly supported.


jordansissel:
Got confirmation from a user in IRC that this patch solves their problem. Good enough for me. Merging!

Issue #875, opened 2015-03-16
--after-upgrade fails


ssenaria:
When attempting to upgrade my RPM I encounter the following error. Using `--after-install` works just fine, `--after-upgrade` creates the RPM just fine but upgrading poses an issue.

    fpm -s dir -t rpm -n mypackage -a x86_64 -v 2 --prefix /nfs/local/web --after-upgrade post-install/ariaapi.txt -C TMP/packagename .

    /var/tmp/rpm-tmp.y139zs: line 23: syntax error near unexpected token `}'
    /var/tmp/rpm-tmp.y139zs: line 23: `}'

Here's line 23:

23: install() {
24: }

b3n3d1k7:
I encountered the same problem - It appears if you provide only one of --after-upgrade or --before-upgrade but not both.
In this case the template is used (if script?(:before_upgrade) or script?(:after_upgrade)) which does not check if the upgrade scripts are emtpy - like for the install scripts.

jordansissel:
I think #888 fixes this.

jordansissel:
Please reopen if you find otherwise.

dividedmind:
Could this please be released?

jordansissel:
I will try to do an fpm release this weekend.

PR #888, opened 2015-04-10
Feature/875

djhaskin987:
Fixes issue #875 by making sure the functions `install()` and `upgrade()` in the pre- and post-install scripts in RPM and similar functions in DEB packages don't explode.

### DIFF
--- a/templates/deb/postinst_upgrade.sh.erb
+++ b/templates/deb/postinst_upgrade.sh.erb
@@ -1,15 +1,21 @@
 #!/bin/sh
 after_upgrade() {
+<%# Making sure that at least one command is in the function -%>
+<%# avoids a lot of potential errors, including the case that -%>
+<%# the script is non-empty, but just whitespace and/or comments -%>
+    :
 <% if script?(:after_upgrade) -%>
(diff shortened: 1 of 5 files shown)


djhaskin987:
Did some cursory tests: Made the packages, ran the pre/post scripts involved to make sure they ran correctly :) :+1:

templates/deb/postinst_upgrade.sh.erb
```diff
@@ -1,15 +1,21 @@
 #!/bin/sh
 after_upgrade() {
+<%# Making sure that at least one command is in the function -%>
+<%# avoids a lot of potential errors, including the case that -%>
+<%# the script is non-empty, but just whitespace and/or comments -%>
```
jordansissel: +1 on your rationale

jordansissel: My agreement, for posterity: We could also just sanitize ensure at least one character is non whitespace, but I think it's just safer to prepend `:` as the first command.

jordansissel:
Eyeballed things; looks good to me. Didn't test, but it'll be allright :P

PR #890, opened 2015-04-10
rspec: fix lintian test

vincentbernat:
lintian expects the package name to ends with ".deb". Also, rspec
disallows the use of a let variable in an `after(:context)` hook. Also
fix the name of instance variable `@staging_path`.


jordansissel:
lgtm

PR #892, opened 2015-04-10
Add support for Solaris 11 style IPS packages using *.p5p format

jbcraig:
I've tested this on Solaris 11.2 VM and it should be good to go.  Seems fairly feature complete but I wasn't sure where I should document the functionality.  I must confess I'm not a ruby developer so some of my constructs could probably be tighter as I've had to borrow from other output types to accomplish this implementation.


jordansissel:
I don't have much solaris 11 experience (nor immediate access right now), so I can't test this immediately. However, I'll do some code review :)

Thank you very much for helping improve fpm!

jordansissel:
First question is a general one, is "p5p" a good name? By "good" I mean is that what most Solaris 11 admins would call the packages? Would "ips" be better? Or "solaris11" ? I have no strong opinions here, just want to make sure you like the name.

Any other solaris  11 users are welcome to comment also :)

lib/fpm/package/p5p.rb
```diff
@@ -0,0 +1,116 @@
...
+    when nil, "default"
+      @category = 'Applications/System Utilities'
+    end
+
+    # Generate a package manifest.
+    pkg_generate = safesystemout("pkgsend", "generate", "#{staging_path}")
+    File.open("#{build_path}/#{name}.p5m.1", "w") do |manifest|
+      manifest.puts pkg_generate
+    end
+    safesystem("cp", "#{build_path}/#{name}.p5m.1", "#{build_path}/#{name}.p5m")
```
jordansissel: Why create the file with '.1' on the end and then copy it to without the '.1' ?

Also, for file renaming, you could just use `File.rename("#{build_path}/#{name}.p5m.1", "#{build_path}/#{name}.p5m")`. This may be preferable to shelling out to `cp` if you just want a rename.

lib/fpm/package/p5p.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    # Add necessary metadata to the generated manifest.
+    metadata_template = template("p5p_metadata.erb").result(binding)
+    File.open("#{build_path}/#{name}.mog", "w") do |manifest|
```
jordansissel: This can be written shorter with the `File.write` shortcut:

```ruby
File.write("#{build_path}/#{name}.p5m.1", metadata_template)
```

lib/fpm/package/p5p.rb
```diff
@@ -0,0 +1,116 @@
...
+    
+    # Fixup the category to an acceptable solaris category
+    case @category
+    when nil, "default"
+      @category = 'Applications/System Utilities'
+    end
+
+    # Generate a package manifest.
+    pkg_generate = safesystemout("pkgsend", "generate", "#{staging_path}")
+    File.open("#{build_path}/#{name}.p5m.1", "w") do |manifest|
```
jordansissel: For all these `"#{build_path}/...` examples, the `build_path` method lets you pass an argument for what to append to it. The goal was to avoid having tons of string evaluations `"#{build_path}/whatever"` which you can do instead as `build_path(whatever)`.

In this example, rewriting:

```ruby
File.open("#{build_path}/#{name}.p5m.1", "w")
```

As this:

```ruby
File.open(build_path("{name}.p5m.1"), "w")
```

It's intended that the 2nd way is more readable, but readability is open to interpretation, so this is only a suggestion :)

lib/fpm/package/p5p.rb
```diff
@@ -0,0 +1,116 @@
...
+    # Add any facets or actuators that are needed.
+    # TODO(jcraig): add manpage actuator to enable install wo/ man pages
+
+    # Verify the package.
+    if attributes[:p5p_lint] then
+      safesystem("pkglint", "#{build_path}/#{name}.p5m")
+    end
+
+    # Publish the package.
+    safesystem("pkgrepo", "create", "#{build_path}/#{name}_repo")
```
jordansissel: `"#{build_path}/#{name}_repo"` is used repeatedly in this section. I recommend you save this in a variable and use that variable:

```ruby
repo_path = "#{build_path}/#{name}_repo"
safesystem("pkgrepo", "set", "-s", repo_path,
  "publisher/prefix=#{attributes[:p5p_publisher]}")
```

templates/p5p_metadata.erb
```diff
@@ -0,0 +1,12 @@
+set name=pkg.fmri value="<%=@name%>@<%=@version%>"
+set name=pkg.summary value="<%=@description%>"
+set name=pkg.human-version value="<%=@version%>"
+set name=pkg.description value="<%=@description%>"
+set name=info.classification value="org.opensolaris.category.2008:<%=@category%>"
+set name=variant.opensolaris.zone <%= attributes[:p5p_zonetype] %>
+set name=variant.arch value=<%=@architecture%>
+<transform file dir -> set owner <%= attributes[:p5p_user] %>>
+<transform file dir -> set group <%= attributes[:p5p_group] %>>
+<transform dir path=opt$->drop>
```
jordansissel: These "drop" actions are to tell pkgmogrify to ignore the `set owner` commands, right? (I'm trying to eyeball the behavior without really knowing much about the p5p process)

jordansissel:
I left a few comments inline in the code, but otherwise this looks fine to me. Given it's a new package support, I'm OK merging without much additional testing done by others.

You said you're not experienced with ruby, and that's totally fine! Your code is good :)
I can help steer things if you aren't sure what to do. Most of my inline comments are suggestions, not requirements.

Anyway; read over my comments and let me know what you think. Once you feel this is ready for merging, I'm happy to merge it!

lib/fpm/package/p5p.rb
jbcraig: The .p5m is the extension used to designate a package manifest (I assume thats akin to the specfile for rpm's).  The .1, .2, ... are progressions in building the final manifest and I left them behind as debug artifacts because it makes it easier to figure out where the process went wrong.

lib/fpm/package/p5p.rb
jbcraig: Excellent, I'll work that back into the pull request.

lib/fpm/package/p5p.rb
jbcraig: Works for me.

lib/fpm/package/p5p.rb
jbcraig: agreed

templates/p5p_metadata.erb
jbcraig: The drop command tells pkgmogrify to drop matching dir's by path from the manifest and prevents conflicts over directories that are delivered in other packages.

jbcraig:
I'll rework the pull request with your helpful suggestions.  

As to the naming, I started with "ips" as the type but the file format for the end artifact is a .p5p file.  P5P files can be installed directly like an rpm or loaded into a network repo for network installs.  Output format uses the fpm type as the extension, so this seemed to be a better fit than .ips.

jbcraig:
Sorry, first time using github as well; didn't mean to close it.

jordansissel:
Btw you don't need to close this, you can push new changes to your branch
and it will be updated here automatically :)

jordansissel:
@jbcraig thanks for clarifying the choice on the 'p5p' naming, that helps a lot!

Let me know when you want me to take a look at this. I anticipate merging after your next set of changes :)

jbcraig:
Added your suggested changes and revised the work-flow.  I've made some simple packages and successfully tested the installs.  Now it just needs to be used and improved through experience in the wild.  My intention is to use this in conjunction with puppet. So, I haven't put much time into thinking about configuration scripts and such as that will leverage puppet.  Thanks for creating a really useful tool and helping me along w/ ruby.

jordansissel:
Sweet! Code looks good to me. Merging :)

binarycrusader:
pkg(5) author here:  I would note there's no reason to provide both a pkg.summary and pkg.description if you're going to set them to the same value.  Just provide pkg.summary (which is hopefully a short piece of text).  pkg.description is for a "long sentence" or paragraph or more of text describing the package.  The system will automatically do the right thing if you omit either one of them.

Likewise, there's no need to set pkg.human-version if it's just going to be the same string provided in the FMRI itself.  The primary purpose of pkg.human-version is when you have letters or other text you want to use to describe the version such as "1.0 beta" or the like.

Note these are all "nits" and not a "bug" per se so I just thought I'd mention them.

PR #897, opened 2015-04-15
Feature/deb scripts via input

djhaskin987:
This pull request fixes fpm's `deb` package class so that it properly places maintainer script contents into the Package class's `scripts` dictionary so that maintainer scripts can be preserved when fpm translates the debian package into a different kind of package.

I ran manual tests on this PR, making sure that scripts were preserved when a debian package was translated into an rpm package. Looks good :+1: . I also checked to make sure that scripts still get overridden by command line options, which is correct.


lib/fpm/package/deb.rb
```diff
@@ -267,6 +267,22 @@ def extract_info(package)
...
+        self.scripts[:after_install] = File.read(File.join(path, "postinst"))
+      end
+      if File.file?(File.join(path, "prerm"))
+        self.scripts[:before_remove] = File.read(File.join(path, "prerm"))
+      end
+      if File.file?(File.join(path, "postrm"))
+        self.scripts[:after_remove] = File.read(File.join(path, "postrm"))
+      end
+      if File.file?(File.join(path, "conffiles"))
+        self.config_files = File.read(File.join(path, "conffiles")).split("\n")
```
djhaskin987: Forgot to mention this -- the pull request also preserves the list of config files when converting from debian packages. Just tested this (just now!) :+1: Works when translating to an RPM package, whether one config file or more than one is originally listed in the debian package.

lib/fpm/package/deb.rb
```diff
@@ -267,6 +267,22 @@ def extract_info(package)
       #self.config_files = config_files
 
       self.dependencies += parse_depends(parse.call("Depends")) if !attributes[:no_auto_depends?]
+
+      if File.file?(File.join(path, "preinst"))
```
jordansissel: You don't have to make this change, but I find lots of File.join's to be noisy, so I might consider writing this using Object#tap:

```ruby
File.join(path, "preinst").tap do |f|
  self.scripts[:before_install] = File.read(f) if File.file?(f)
end
```

We could go further and have a mapping of script symbol to file names:

{ :before_install => "preinst", ... }.each do |script, name|
  File.join(path, name).tap do |f|
      self.scripts[script] = File.read(f) if File.file?(f)
  end
end
```

jordansissel:
code lgtm

djhaskin987:
@jordansissel I'll look into doing that improvement. I also want to start writing unit tests for my work, I just discovered the spec/ directory :)

PR #911, opened 2015-05-07
Feature: empty output package target

djhaskin987:
I think it is useful to have `empty` as an output target. Sometimes, when converting from a certain package type to another, there is a problem with how the conversion takes place. I want to be able to run fpm with `--debug-workspace` and see what the result of converting from the source type, without regard to the target type until I've worked out what went wrong converting the source type. I don't care what type of package it outputs. In this case, having an `empty` output target is nice.

PR comes pre-tested, and includes a unit test for the `#to_s` function used to make `empty` an output target.

### DIFF
--- a/lib/fpm/package/empty.rb
+++ b/lib/fpm/package/empty.rb
@@ -4,4 +4,10 @@ require "backports"
 # Empty Package type. For strict/meta/virtual package creation
 
 class FPM::Package::Empty < FPM::Package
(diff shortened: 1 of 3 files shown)


jordansissel:
Hmm.. I suppose I see the use case here (for debugging an input). Until we have a better way to debug things via rspec/pry like this, this is ok :)

PR #912, opened 2015-05-07
Add `:` to pre- and post- removal scripts to make them safer

djhaskin987:
This goes back to #875 . It makes sure that no-op scripts put into the templated version of removal scripts in the case of a `--*-upgrade` option doesn't explode. Tested :+1:

### DIFF
--- a/templates/rpm.erb
+++ b/templates/rpm.erb
@@ -184,6 +184,10 @@ fi
 %preun
 if [ "${1}" -eq 0 ]
 then
+<%# Making sure that at least one command is in the function -%>
+<%# avoids a lot of potential errors, including the case that -%>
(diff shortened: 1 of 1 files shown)


templates/rpm.erb
```diff
@@ -184,6 +184,10 @@ fi
 %preun
 if [ "${1}" -eq 0 ]
 then
+<%# Making sure that at least one command is in the function -%>
+<%# avoids a lot of potential errors, including the case that -%>
```
jordansissel: I'm OK with this change, but I wonder if we should just move this logic into the script() method? Something to ponder.

jordansissel:
<3

PR #913, opened 2015-05-07
Feature: sh before-install script

djhaskin987:
This adds `--before-install` functionality to the `sh` output target. It runs the `--before-install` script as part of the BASH script that is run before the tarball is extracted. Tested :+1:


jordansissel:
LGTM, would love to see specs for this, but won't block the merge.

jordansissel:
Thank you, @djhaskin987!

djhaskin987:
You're welcome! I'll try to get those specs in :)

Issue #231, opened 2012-05-30
Add Arch Linux packaging support


gtaylor:
I'd settle for just the PKGBUILD file generation, but generation of the package archives would be trivial to add if we could generate PKGBUILD files.

djhaskin987:
Just a status update. all my work for this ticket is on the branch `feature/arch-support` on my forked version of this repository. The work is nearly complete; I just need to implement an arch specific option, add some more unit tests, fix the ones I've added, and test my work generally. Almost there!

djhaskin987:
@hatt @jordansissel and anyone else who wants to help, I could use an extra pair of eyes.
My feature branch is so *very* near complete. I just hit this one last snag:

I wrote a test to make sure that permissions are properly preserved from the staging path's files:

```
Failures:

  1) FPM::Package::Pacman#output file permissions should preserve file permissions for /usr
     Failure/Error: dir)).mode & 07777 } == perm
     Insist::Failure:
       Expected 493, but got 509
     # ./spec/fpm/package/pacman_spec.rb:164:in `block (5 levels) in <top (required)>'

  2) FPM::Package::Pacman#output file permissions should preserve file permissions for /usr/bin
     Failure/Error: dir)).mode & 07777 } == perm
     Insist::Failure:
       Expected 493, but got 509
     # ./spec/fpm/package/pacman_spec.rb:164:in `block (5 levels) in <top (required)>'

  3) FPM::Package::Pacman#output file permissions should preserve file permissions for /usr/lib
     Failure/Error: dir)).mode & 07777 } == perm
     Insist::Failure:
       Expected 493, but got 509
     # ./spec/fpm/package/pacman_spec.rb:164:in `block (5 levels) in <top (required)>'

  4) FPM::Package::Pacman#output file permissions should preserve file permissions for /usr/share
     Failure/Error: dir)).mode & 07777 } == perm
     Insist::Failure:
       Expected 493, but got 509
     # ./spec/fpm/package/pacman_spec.rb:164:in `block (5 levels) in <top (required)>'

  5) FPM::Package::Pacman#output file permissions should preserve file permissions for /usr/share/doc
     Failure/Error: dir)).mode & 07777 } == perm
     Insist::Failure:
       Expected 493, but got 509
     # ./spec/fpm/package/pacman_spec.rb:164:in `block (5 levels) in <top (required)>'

  6) FPM::Package::Pacman#output file permissions should preserve file permissions for /etc
[...]

PR #916, opened 2015-05-09
Feature: Pacman (ArchLinux package) support

djhaskin987:
I finished it! Addresses #231.

Features
=======

* ArchLinux packaging support (`pacman` packages)
* Lots of rspecs to make sure it works :+1: Everytime I found an error, I wrote an rspec and coded to it. Broke none, made more.
* Both `input()` (conversion from pacman) and `output()` (conversion to pacman) works.
* Implementation is completely independent of `makepkg`. It doesn't call it, it doesn't use it. 
    * That makes fpm the first pacman implementation (that I know of) that is MIT licensed
    * That also means you can create a pacman package without being on ArchLinux.
    * All you need to have on your machine is `tar`, `du`, and `bsdtar`

Known issue(s)
============

* The `input()` function uses a method that sort-of parses bash so that it can slurp the maintainer scripts into FPM format. It's not perfect -- If it sees an unbalanced curly brace in valid bash, like `echo "{"`, it will choke. However, this is a small corner case and not common, and unless I made the function nearly a full bash parser itself, I couldn't get around this. However, for 98% of the cases out there, it should work.
* With other package types, the `--after-install` and `--before-install` , as well as the `--after-remove` and `--before-remove` scripts run regardless of whether it's an upgrade case or not. With pacman, if the installation is an upgrade, only the `--after-upgrade` and `--before-upgrade` scripts are run. This is more of a pacman nuance, but it should be noted. If those scripts are not specified, pacman will die

Other notes
=========

I had to move the `copy_metadata()` method from the `FPM::Package::Dir` class to the `FPM::Util` module. It's included in all the classes, so the `Dir` class still works, but pacman is *really* picky with permissions. To ensure that the permissions of the files in the pacman package match those of the files the packager gave FPM, I needed to use `copy_metadata` in `FPM::Package::Pacman` as well as `FPM::Package::Dir`, so I made it available to everyone by putting it in the `FPM::Util` module. Hope that's alright :) I tried to make sure I made as few disruptive changes as possible, so I hope this change is not unwelcome :)


lib/fpm/package/dir.rb
```diff
@@ -203,27 +203,5 @@ def copy(source, destination)
     copy_metadata(source, destination)
   end # def copy
 
-  def copy_metadata(source, destination)
```
djhaskin987: I mentioned this in the main notes. I needed this function in the pacman class, so I put it in `FPM::Util`.

README.md
```diff
@@ -115,6 +115,7 @@ Sources:
 * rpm
 * deb
 * node packages (npm)
+* pacman (`pkg.tar.xz`, or ArchLinux) packages
```
hatt: Other distros use Pacman format packages (Frugalware for example) and there are also other compression systems allowed. I know it seems weird for Arch users to not have everything up to date but for various reasons relating to that, a flag to allow selecting of lzma/xz/gz/bz2 would be a nice addition.

templates/pacman.erb
```diff
@@ -0,0 +1,47 @@
...
+<% end -%>
+group = <%= category %>
+<% conflicts.each do |confl| -%>
+conflict = <%= confl %>
+<% end -%>
+<% provides.each do |prov| -%>
+provides = <%= prov %>
+<% end -%>
+<% config_files.each do |conf| -%>
+<%   conf.gsub!(/^\/+/, "") -%>
```
ServiusHack: I get an error here:
```
/usr/lib/ruby/gems/2.2.0/gems/fpm-1.3.3/templates/pacman.erb:39:in `gsub!': can't modify frozen String (RuntimeError)
```

Replacing `gsub!` with `gsub` solves it.

djhaskin987:
Thanks! @hatt, @ServiusHack I'll look into it.

flaccid:
I'm very excited. Nearly there?

djhaskin987:
The only suggestion I've had for this is that I support more compression types other than `.xz`. Once that is addressed, I'd say we're there.

hatt:
@djhaskin987 any progress on that? Seems pretty trivial to add, we already do it in RPM. Happy to send over a quick patch in my lunch break if you like.

djhaskin987:
Just added support for different compression types, including no compression at all. Tested :+1: 
As far as I am concerned, this feature is done, done, done. It has been tested, specs written for it. It's ready to go.

djhaskin987:
LOLZ, of course I forgot something.

I just perused through the history of the ticket to make sure all my ducks were in a row, glad I did. Just added the fix to the bug you found @ServiusHack ! Thanks for bringing it up.

mark-rushakoff:
@jordansissel anything blocking this from being merged? I'm already using FPM for .deb and .rpm, and would love to be able to build packages for Arch as well.

RomanSaveljev:
I need this merged badly! :+1:

Vagrantfile
```diff
@@ -1,11 +1,17 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+$arch_install_puppet_script = <<EOF
+yaourt --sucre
```
jordansissel: Unrelated, but I pronounce yaourt as "yogurt" because... why not! :P

lib/fpm/package/pacman.rb
```diff
@@ -0,0 +1,394 @@
+# -*- coding: utf-8 -*-
+require "fpm/package"
+require "backports"
+require "fileutils"
+require "find"
+
+class FPM::Package::Pacman < FPM::Package
+
+  opt_depends = []
+  option "--optdepends", "PACKAGE",
```
jordansissel: For multivalue options, this trickery isn't needed anymore. Clamp has an option `:multivalued => true` that does this for us which is kind of nice.

Example:

```ruby
option "--optdepends", "PACKAGE", "Add an optional dependency to the pacman package", :multivalued => true
```

You don't necessarily need to change this, but just a heads up that clamp supports this well now :)

jordansissel: subjective - Is `--pacman-optdepends` a good name for the flag? Thoughts on `--pacman-optional-depends`  or something not abbreviated? It's your call :)

lib/fpm/package/pacman.rb
```diff
@@ -0,0 +1,394 @@
...
+    if !COMPRESSION_TYPES.include?(value)
+      raise ArgumentError, "Pacman compression value of '#{value}' is invalid. " \
+        "Must be one of #{COMPRESSION_TYPES.join(", ")}"
+    end
+    value
+  end
+
+  def initialize(*args)
+    super(*args)
+    attributes[:pacman_opt_depends] = []
```
jordansissel: Is this attribute for the `--optdepends` flag? If so, it should automatically show up as `:pacman_optdepends` as you have it defined. I don't think you'll need to do this, especially if you change the option to `:multivalued => true`

And also, since I think this is a typo, does this work? :P

lib/fpm/package/pacman.rb
```diff
@@ -0,0 +1,394 @@
...
+        return %x{uname -m}.chomp   # 'native' is current arch
+      when "all", "any", "noarch"
+        return "any"
+      else
+        return @architecture
+    end
+  end # def architecture
+
+  def iteration
+    return @iteration ? @iteration : 1
```
jordansissel: Style suggestion, this could be written as `return @iteration || 1` instead

lib/fpm/package/pacman.rb
```diff
@@ -0,0 +1,394 @@
...
+
+  def config_files
+    return @config_files || []
+  end # def config_files
+
+  def dependencies
+    bogus_regex = /[^\sA-Za-z0-9><=-]/
+    # Actually modifies depencies if they are not right
+    bogus_dependencies = @dependencies.grep bogus_regex
+    if !bogus_dependencies.empty?
```
jordansissel: Style recommendation: `if bogus_dependencies.any?`

lib/fpm/package/pacman.rb
```diff
@@ -0,0 +1,394 @@
...
+  # The idea is that you can keep pumping in new things to a package
+  # for later conversion or output.
+  #
+  # Implementations are expected to put files relevant to the 'input' in the
+  # staging_path
+  def input(pacman_pkg_path)
+    control = {}
+    # Unpack the control tarball
+    safesystem("tar", "-C", staging_path, "-xf", pacman_pkg_path)
+    pkginfo = File.join(staging_path, ".PKGINFO")
```
jordansissel: Style recommendation: staging_path is a method that can take an argument. Mainly to avoid having to explicitly call File.join all over the place, you can do this:

```ruby
# Equivalent to File.join(staging_path, ".PKGINFO")
staging_path(".PKGINFO")
```

lib/fpm/package/pacman.rb
```diff
@@ -0,0 +1,394 @@
...
+  end # def generate_mtree
+
+  # KNOWN ISSUE:
+  # If an un-matched bracket is used in valid bash, as in
+  # `echo "{"`, this function will choke.
+  # However, to cover this case basically
+  # requires writing almost half a bash parser,
+  # and it is a very small corner case.
+  # Otherwise, this approach is very robust.
+  def gobble_function(cons,prod)
```
jordansissel: Gobble gobble! (no complaints, just being silly)

turkey2

lib/fpm/package/pacman.rb
```diff
@@ -0,0 +1,394 @@
...
+    look_for.each do |fname|
+      functions[fname] = []
+    end
+
+    open(path, "r") do |iscript|
+      lines = iscript.each
+      begin
+        while true
+          line = lines.next
+          m = /^\s*(\w+)\s*\(\s*\)\s*\{\s*([^}]+?)?\s*(\})?\s*$/.match(
```
jordansissel: maybe put a comment above showing an example line entry for easier eyeballing of this to see if it's right

lib/fpm/package/pacman.rb
```diff
@@ -0,0 +1,394 @@
...
+            line)
```
jordansissel: put this on the line above?

Maybe put the regexp into a constant and then have `m = INSTALL_SCRIPT_LINE_REGEX.match(line)` or something?

jordansissel:
Lots of specs failing when I run them on OSX:

Example - all are related to `tar` invocations, it seems:

```
  1) FPM::Package::Pacman#output Test that empty epoch is tested properly should have the correct name
     Failure/Error: original.output(target)
     FPM::Util::ProcessFailed:
       tar failed (exit code 1). Full command was:["tar", "cJf", "/var/folders/4r/4fpyt_2d1qvfrgb2fnngrqm40000gn/T/studtmp-19648f37-87eb-4ec8-84fe-888b422caad2.pkg.tar.xz", "--numeric-owner", "--owner", "0", "--numeric-owner", "--group", "0", ".MTREE", ".PKGINFO"]
     # ./lib/fpm/util.rb:83:in `safesystem'
     # ./lib/fpm/package/pacman.rb:273:in `block (2 levels) in output'
     # ./lib/fpm/package/pacman.rb:272:in `chdir'
     # ./lib/fpm/package/pacman.rb:272:in `block in output'
     # ./lib/fpm/util.rb:146:in `call'
     # ./lib/fpm/util.rb:146:in `with'
     # ./lib/fpm/package/pacman.rb:271:in `output'
     # ./spec/fpm/package/pacman_spec.rb:89:in `block (4 levels) in <top (required)>'
```

lib/fpm/package/pacman.rb
djhaskin987: Yes, it is for the optdepends flag. I'm fixing it. It is not a typo; it works :dancer:

djhaskin987:
Yes, the tar invocations --

The only way to make all the options to work is to use `bsdtar` . Normally on systems, when you call `bsdtar`, it forwards the call to GNU tar, not the *actual* `bsdtar` . But they're actually different. You'll find that the tar spec problems disappear when you install `bsdtar` instead of just the usual tar.

`bsdtar` is a prerequisite for packaging arch linux, the tar format is unique to `bsdtar`. I ran rspec, it works (except for an unrelated python rspec :+1: )

djhaskin987:
I have looked at all of your suggestions, thanks for them! Let me know if you would like to see anything else.

djhaskin987:
Actually, looking at the error, I don't know if this is a `bsdtar` thing. But my specs run fine? I'm confused.

djhaskin987:
FYI
```
[ djhaskin987@djhaskin987-S301LA:~ ]$ tar --version
tar (GNU tar) 1.27.1
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by John Gilmore and Jay Fenlason.
[ djhaskin987@djhaskin987-S301LA:~ ]$ bsdtar --version
bsdtar 3.1.2 - libarchive 3.1.2
[ djhaskin987@djhaskin987-S301LA:~ ]$
```

djhaskin987:
Tested on CentOS 6 and Debian 6. As long as bsdtar was installed, I got no rspec pacman errors.

djhaskin987:
It looks like XZ compression [may not be entirely supported on OS X](https://www.aeyoun.com/technote/archive-utility-xz.html). I'll try using the `--xz` option and see what happens. The document in that link says it already works on OS X :)

djhaskin987:
I changed the code to use an option that is supposed to work better is OS X, But I don't have access to OS X. The change works in Ubuntu, Debian and CentOS. Would you mind testing it again, @jordansissel?

flaccid:
For OS X El Capitan users they will probably run into https://github.com/ffi/ffi/issues/461 which is what I'm running into.

djhaskin987:
In addition, OS X users *may* need to install [xz](http://tukaani.org/xz/), although supposedly with the `--xz` option OS X can create (if not decompress) tar.xz archives. I appreciate your help tracking OS X bugs down, @flaccid , and anyone else who would care to look at my stuff for that OS. I just don't have a mac available to me, I'm on Ubuntu 15.04 .

I do have access via vagrant to a FreeBSD 10.2 machine I can test it on, I'll look into making sure it runs on that. Would that help?

hatt:
The OS X bundled version of tar supports `--xz` by default anyway, assuming the user is on 10.7 or later. The -J flag seems to be arbitrary between `gnutar` and `bsdtar`

flaccid:
Testing with `fpm --debug -s dir -t pacman -n "slashbin" -v 1.0 /bin /sbin`.
...
```
Process is running {:pid=>15023, :level=>:debug, :file=>"fpm/util.rb", :line=>"74", :method=>"safesystem"}
tar: Option --owner is not supported {:level=>:info, :file=>"cabin/mixins/pipe.rb", :line=>"47", :method=>"block in pipe"}
Usage: {:level=>:info, :file=>"cabin/mixins/pipe.rb", :line=>"47", :method=>"block in pipe"}
  List:    tar -tf <archive-filename> {:level=>:info, :file=>"cabin/mixins/pipe.rb", :line=>"47", :method=>"block in pipe"}
  Extract: tar -xf <archive-filename> {:level=>:info, :file=>"cabin/mixins/pipe.rb", :line=>"47", :method=>"block in pipe"}
  Create:  tar -cf <archive-filename> [filenames...] {:level=>:info, :file=>"cabin/mixins/pipe.rb", :line=>"47", :method=>"block in pipe"}
  Help:    tar --help {:level=>:info, :file=>"cabin/mixins/pipe.rb", :line=>"47", :method=>"block in pipe"}
Process failed: tar failed (exit code 1). Full command was:["tar", "--xz", "-cf", "/private/tmp/slashbin-1.0-1-x86_64.pkg.tar.xz", "--numeric-owner", "--owner", "0", "--numeric-owner", "--group", "0", ".MTREE", ".PKGINFO", "bin", "sbin"] {:level=>:error, :file=>"fpm/command.rb", :line=>"492", :method=>"rescue in execute"}
Cleaning up staging path {:path=>"/var/folders/sx/ht6gznq979v5r725sqt7_p617p38sz/T/package-dir-staging20151020-15018-clex4o", :level=>:debug, :file=>"fpm/package.rb", :line=>"279", :method=>"cleanup_staging"}
Cleaning up build path {:path=>"/var/folders/sx/ht6gznq979v5r725sqt7_p617p38sz/T/package-dir-build20151020-15018-yq29wo", :level=>:debug, :file=>"fpm/package.rb", :line=>"286", :method=>"cleanup_build"}
Cleaning up build path {:path=>"/var/folders/sx/ht6gznq979v5r725sqt7_p617p38sz/T/package-pacman-build20151020-15018-1hud8fy", :level=>:debug, :file=>"fpm/package.rb", :line=>"286", :method=>"cleanup_build"}
```

jordansissel:
Just FYI, I'm OK if this doesn't work on OSX for the first version of it. (We can merge if folks agree it's ready)

djhaskin987:
You will hear no dissent from me :clap:

hatt:
:+1:

jordansissel:
damn you merge conflicts! I will try to fix and merge.

jordansissel:
Merged <3

djhaskin987:
YAY!!!!! Thanks!

coxley:
:+1

flaccid:
Thanks!

PR #917, opened 2015-05-09
Add specs for PR #913

djhaskin987:
@jordansissel You said in #913 that you would like to see some specs for that PR, so I have gone back and made them. They pass and the thing works, hope this helps.


jordansissel:
Wooo!

djhaskin987:
:palm_tree: :palm_tree: :palm_tree: :palm_tree:

PR #922, opened 2015-05-15
Add --[no-]cpan-cpanm-force flag to allow passing --force to cpanm

elisiano:
This is useful in case the user doesn't care wether or not some test fails


hatt:
Is there a reason the user wouldn't care if a cpan test fails? They're kind of important to know, and you can disable testing outright for cpanm already. The only thing I can think of is that RHEL has some weird directory structures that break things like MakeMaker, in which case not sandboxing or testing is fine.

elisiano:
In my case some SSL tests were failing because of wrong certificates. Not happy, but I had to get things done.

jordansissel:
Code and idea seem ok to me. Merging!

PR #927, opened 2015-05-27
Add more clear description of --directories option

pdaukintis:
Add (hopefully) more clear description of _'--directories'_ option, when it should be used multiple times if directories are not under same parent directory. Current explanation is not very clear about this, and it seems there are no other usage examples with _'--directories'_ option in fpm project' wiki, I think it is a good idea to explain it in this place.


jordansissel:
LGTM, Thanks for improving the docs!

PR #930, opened 2015-05-29
Feature/697

djhaskin987:
This PR is for #697 .

My shot at a virtualenv input plugin, after taking a look at @evilsocket 's code. Thanks for letting me stand on your shoulders, @evilsocket ! 

# Features
1. Virtualenv named based on the pip package it's built around
2. You can specify a particular version of a pip package to build a virtualenv around e.g. `psutil==2.2.1`
2. No surprises. I took the script from here --> https://github.com/brutasse/graphite-api/blob/master/fpm/build-deb.sh and put it straight into FPM as much as possible
3. Version resolution: The version of the "chief" package is used if no other version is specified
4. Optionally add extra files, like [the script above does](https://github.com/brutasse/graphite-api/blob/master/fpm/build-deb.sh) when it adds files that go under `/etc`
5. By default, prefixes virtualenv package names with `virtualenv`, but this can be overridden
6. Option exists to change the index url which pip uses to build the virtualenv
5. Tried as much as possible to conform to how other FPM plugins are coded

# Caveats
1. You cannot currently use '>=' or '<=' in the package specification, only '==' currently exists as an option.
5. Places virtualenvs in an install location specified by the `--virtualenv-install-location`. Couldn't use prefix, as this changes both where the virtualenv *and* the extra files go. If I specified `--prefix foo` and `--virtualenv-install-location /bar`, the virtual env for `psutil` would be installed under `/foo/bar/psutil`
2. You get bad results if you use a relative path for the value of `--virtualenv-install-location`.
3. No specs yet (they will be forthcoming) but I have manually tested everything here and it works :+1:

# Example

A default case:
```

~/Workspace/compute/src/github.com/djhaskin987/fpm $ bin/fpm -s virtualenv -t deb --virtualenv-other-files-dir pkg/ example
Created package {:path=>"virtualenv-example_1.0_amd64.deb"}
```

A rather involved example:
```
~/Workspace/compute/src/github.com/djhaskin987/fpm $ bin/fpm -f -n barg -v 3.0.0 --prefix /boogar --virtualenv-install-location /junie --no-virtualenv-fix-name -s virtualenv -t rpm --virtualenv-other-files-dir pkg/ psutil==2.2.1
virtualenv_folder: /junie/psutil
no value for epoch is set, defaulting to nil {:level=>:warn}
Force flag given. Overwriting package at barg-3.0.0-1.x86_64.rpm {:level=>:warn}
no value for epoch is set, defaulting to nil {:level=>:warn}
Created package {:path=>"barg-3.0.0-1.x86_64.rpm"}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,12 @@
 # vim
 .*.sw[a-z]
 
+# emacs
+\#*#
+
+# other editors
+*~
+
 # build byproducts
 build-*/*
 fpm.wiki
--- a/lib/fpm/package/dir.rb
+++ b/lib/fpm/package/dir.rb
@@ -203,27 +203,5 @@ class FPM::Package::Dir < FPM::Package
     copy_metadata(source, destination)
   end # def copy
 
-  def copy_metadata(source, destination)
-    source_stat = File::lstat(source)
-    dest_stat = File::lstat(destination)
-
-    # If this is a hard-link, there's no metadata to copy.
-    # If this is a symlink, what it points to hasn't been copied yet.
-    return if source_stat.ino == dest_stat.ino || dest_stat.symlink?
-
-    File.utime(source_stat.atime, source_stat.mtime, destination)
-    mode = source_stat.mode
-    begin
-      File.lchown(source_stat.uid, source_stat.gid, destination)
-    rescue Errno::EPERM
-      # clear setuid/setgid
-      mode &= 01777
-    end
-
-    unless source_stat.symlink?
-      File.chmod(mode, destination)
-    end
-  end # def copy_metadata
-
   public(:input, :output)
 end # class FPM::Package::Dir
--- a/lib/fpm/package/virtualenv.rb
+++ b/lib/fpm/package/virtualenv.rb
@@ -1,0 +1,127 @@
+require "fpm/namespace"
+require "fpm/package"
+require "fpm/util"
+require "rubygems/package"
+require "rubygems"
+require "fileutils"
+require "tmpdir"
+require "json"
+
+# Support for python virtualenv packages.
+#
+# This supports input, but not output.
+#
+# Example:
+#
+#     # Download the django python package:
+#     pkg = FPM::Package::Python.new
+#     pkg.input("Django")
+#
+class FPM::Package::Virtualenv < FPM::Package
+  # Flags '--foo' will be accessable  as attributes[:python_foo]
+  option "--pypi", "PYPI_URL",
+  "PyPi Server uri for retrieving packages.",
+  :default => "https://pypi.python.org/simple"
+  option "--package-name-prefix", "PREFIX", "Name to prefix the package " \
+  "name with.", :default => "virtualenv"
+  option "--fix-name", :flag, "Should the target package name be prefixed?",
+  :default => true
+  option "--install-folder", "BIN_PATH", "The path to where python scripts " \
+  "should be installed to.", :default => "/usr/share/python"
+  option "--other-files-directory", "DIRECTORY", "Optionally, the contents of the " \
+  "specified directory may be added to the package.", :default => nil
+
+  private
+
+  # Input a package.
+  #
+  def input(package)
+    m = /^([^=]+)==([^=]+)$/.match(package)
+    package_version = nil
+    if m
+      self.name ||= m[1]
+      self.version ||= m[2]
+      package = m[1]
+      package_version = m[2]
+    else
+      self.name ||= package
+    end
+
+    if self.attributes[:virtualenv_fix_name?]
+      self.name = [self.attributes[:virtualenv_package_name_prefix],
+                   self.name].join("-")
+    end
+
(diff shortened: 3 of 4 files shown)


jordansissel:
I'll see about testing this soon. Thank you and @evilsocket for working on this! :)

untergeek:
I volunteer to use https://github.com/elastic/curator as a test env.

evilsocket:
@jordansissel @djhaskin987 it's been my pleasure :)

jordansissel:
In testing this, I found a few things:

* utime: Operation not permitted @ utime_internal

This may only affect OSX, not sure yet. Here's the stack trace (trimmed for brevity)
```
/Users/jls/projects/fpm/lib/fpm/util.rb:170:in `utime': Operation not permitted @ utime_internal - /var/folders/4r/4fpyt_2d1qvfrgb2fnngrqm40000gn/T/package-virtualenv-staging20150615-48864-vpnc11/./usr/share/python/curator/.Python (Errno::EPERM)
        from /Users/jls/projects/fpm/lib/fpm/util.rb:170:in `copy_metadata'
        from /Users/jls/projects/fpm/lib/fpm/package/dir.rb:203:in `copy'
        from /Users/jls/projects/fpm/lib/fpm/package/dir.rb:148:in `block in clone'
...
        from /Users/jls/.rvm/rubies/ruby-2.1.6/lib/ruby/2.1.0/find.rb:42:in `find'
        from /Users/jls/projects/fpm/lib/fpm/package/dir.rb:146:in `clone'
        from /Users/jls/projects/fpm/lib/fpm/package/dir.rb:71:in `block in input'
        from /Users/jls/projects/fpm/lib/fpm/package/dir.rb:69:in `chdir'
        from /Users/jls/projects/fpm/lib/fpm/package/dir.rb:69:in `input'
        from /Users/jls/projects/fpm/lib/fpm/package/virtualenv.rb:120:in `input'
...
```

Works on Ubuntu 12.04, though. :)
This seems unrelated to this patch, probably, but still a bug to fix.

----

Testing the resulting package:

```
% head /usr/share/python/elasticsearch-curator/bin/curator
#!/usr/share/python/elasticsearch-curator/bin/python

% /usr/share/python/elasticsearch-curator/bin/curator --version
curator, version 3.1.0
```

So far so good!

jordansissel:
Installing the .deb on a system without any python installation:

```
root@3984584eefbf:/work# dpkg -i virtualenv-elasticsearch-curator_1.0_amd64.deb
Selecting previously unselected package virtualenv-elasticsearch-curator.
(Reading database ... 11528 files and directories currently installed.)
Preparing to unpack virtualenv-elasticsearch-curator_1.0_amd64.deb ...
Unpacking virtualenv-elasticsearch-curator (1.0) ...
Setting up virtualenv-elasticsearch-curator (1.0) ...
root@3984584eefbf:/work# /usr/share/python/elasticsearch-curator/bin/curator
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
ImportError: No module named site
```

It can't find the `site` module, but it's in the .deb: `./usr/share/python/elasticsearch-curator/lib/python2.7/site.py`

On systems _with_ python available without virtualenv, the system's site.py will be used when the virtualenv one cannot be found. I think this is a bug. @djhaskin987 - thoughts? Is this expected? (I don't know much about virtualenv)

djhaskin987:
@jordansissel I believe virtualenv works either by creating a hard link between or copying python outright from the system. It therefore assumes that python is on the system. The ideal thing would be guaranteeing that python is on the parent system by adding a dependency, but it would be a nightmare deciding what python is called by the target system's package manager.  I think the only thing feasible is to print a warning about this. Do you think that would suit?

djhaskin987:
@jordansissel as to the bug, it appears this line is the "culprit":

```
    120     dir.input(".")
```
Which means that this happens during cloning the directory.

jordansissel:
> ﻿I believe virtualenv works either by creating a hard link between or copying python outright from the system. It therefore assumes that python is on the system.

Yeah I noticed that it copies (or hardlinks, I didnt' check) the python binary but creates symlinks for things like site.py and other similar things.

Going forward, it's probably OK to have this as a known limitation. Unfortunately, it doesn't solve @untergeek's needs (self-contained package), yet, but I think this works well enough to move forward with merging.

jordansissel:
@djhaskin987 agreed that the bug isn't in your code (dir copying something funky happens on OSX)

jordansissel:
Huge thanks to @evilsocket and @djhaskin987 for making virtualenv support happen :)

djhaskin987:
Yay! thx :) :+1:

nffdiogosilva:
Hello there,

I'm having problems understanding correctly how the command works to generate a package with a specific virtualenv and its requirements.

How can I create a package for a virtualenv with **multiple** requirements?

Thanks in advance,
Nuno

PR #931, opened 2015-06-05
Feature/755 Take prefix path into account when searching for config files

djhaskin987:
I am not done with this pull request; I still need to test it and add specs. However, I wanted to start the discussion first to make sure I am going down the right path.

The PR #755 by @michaeltchapman is a good piece of code.

After reading it, I noticed two things:
1. The `staging_path` function might be used to greater effect in the code added by the PR
2. Way more importantly, @michaeltchapman is solving a much more generic problem here than just RPM packaging.

In this PR, I move his logic into the `convert()` function, which is called right before `output()`. `output()` is where the piece of code in PR #755 worked. I think this is better done in convert, where common variables are copied over and where, therefore, the configuration files could be straightened out in the generic way that Michael used.

I merge that logic for finding the right config file and taking into account the prefix path with another snippet of code in the `deb` output class that seems to do the exact same thing, except that the `deb` code has not historically taken the prefix path into account (hence this pull request).

I then removed the generic 'find the config files' code from the `output` functions in both the `deb` and `rpm` classes.

This is great, because this logic makes sense in any output context, not just RPM or DEB.

@jordansissel, if you think this looks good, let me know. Then, I'll test it and add specs. I'll post here when I do, and then you can merge it in if you like :) I'm just too tired to do that tonight, it's after midnight here. Cheers from Utah!


djhaskin987:
Ping @ronkorving @adrianlzt @hatt @dustinbrown @jeffgeiger #584 #730

.gitignore
```diff
@@ -1,6 +1,10 @@
 # vim
```
djhaskin987: gitignore = pork barrel :P If you don't like these entries, no worries :) I put the gitconfig changes in it's own commit, so I can revert it if need be pretty easy.

hatt:
Hi @djhaskin987, this looks pretty good as this has been a huge problem for quite a while for anyone in enterprise environments. I'm happy with it but I think @jordansissel may want some spec tests to prove it works, especially when we have some known fail conditions from previous issues. I know there is a PR for Arch support that'll definitely make use of this. Not sure about the gitignore stuff for now though. Thanks for your work!

jordansissel:
+1 on your proposed change. As @hatt predicted, I would love some rspec coverage of this new change ;)

(FYI, this PR now has merge conflicts)

jordansissel:

josephholsten:
@djhaskin987 are you able to help me get this rebased onto the current master? seems like this was once a very desired patch

djhaskin987:
I got it, thanks @josephholsten . I should be able to fix this PR soon, I just need to add some specs after I rebase. I'm also working on another PR for virtualenv stuffs, which is nearly complete.

Aidenir:
Am I right to assume this feature PR is dead? Can't really understand why, without it FPM is basically useless to anyone in an enterprise environment...

PR #936, opened 2015-06-10
Removed the numeric-owner tar flag when deb_user=root

jhooyberghs:
I removed a couple of lines of code that made it impossible to specify only a deb-user or deb-group without the other. I don't know why this was the way it was, so feel free to refuse this if this should be the way it is now.

Current situation
fpm --verbose -s dir -t deb -n testpackage --architecture all -C test --prefix /opt  --deb-user testuser --iteration 1 .
dpkg-deb -c testpackage_1.0-1_all.deb 
drwx------ 1000/0            0 2015-06-10 22:53 ./
drwxrwxr-x 1000/0            0 2015-06-10 22:53 ./usr/
drwxrwxr-x 1000/0            0 2015-06-10 22:53 ./usr/share/
drwxrwxr-x 1000/0            0 2015-06-10 22:53 ./usr/share/doc/
......
You can see the deb-user being set to 1000, where I would expect to see 'testuser'.

With the change, there's no more numerical tar tag and therefor (the for me expected) folllowing result:
/fpm --verbose -s dir -t deb -n testpackage --architecture all -C test --prefix /opt  --deb-user testuser --iteration 2 .
dpkg -c testpackage_1.0-2_all.deb 
drwx------ testuser/root     0 2015-06-10 22:55 ./
drwxrwxr-x testuser/root     0 2015-06-10 22:55 ./usr/
drwxrwxr-x testuser/root     0 2015-06-10 22:55 ./usr/share/
......


jordansissel:
hmm.. I agree with your problem statement and the solution. I wonder how lintian or other deb sanity checking tools will react. I"ll check that soon :)

jordansissel:
The patch is weird. The way I read it, non-root deb users should get named, not numbered. Hmm, I'll try some tests.

jordansissel:
```
% ./bin/fpm -f -s dir -t deb --deb-user jls -n fizzle ~/.zshrc; dpkg -c fizzle_1.0_amd64.deb| grep zshrc     
...
-rw-r--r-- 1000/0        12153 2015-07-17 00:26 ./home/jls/.zshrc
```

Interesting. It's actually like it ignores the --deb-user flag somehow.

Running fpm with --debug, I see this:

```
Running command {:args=>["tar", "-C", "/tmp/package-dir-staging20150806-5704-h68f7h", "-z", "--owner", "jls", "--numeric-owner", "--group", "0", "-cf", "/tmp/package-deb-build20150806-5704-7hmczn/data.tar.gz", "."], :level=>:debug, :file=>"/home/jls/projects/fpm/lib/fpm/util.rb", :line=>"61", :method=>"safesystem"}
```

Interesting that --numeric-owner is still present.

jordansissel:
This seems like a bug in gnu tar. The `--numeric-owner` flag affects _both_ user and group labeling.

jordansissel:
Here's my proposed patch that I _think_ may work towards solving this:

```
diff --git a/lib/fpm/package/deb.rb b/lib/fpm/package/deb.rb
index 3592389..cfd2d9e 100644
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -796,7 +796,7 @@ class FPM::Package::Deb < FPM::Package
 
       if !attributes[:deb_group].nil?
         if attributes[:deb_group] == 'root'
-          data_tar_flags += [ "--numeric-owner", "--group", "0" ]
+          data_tar_flags += [ "--group", "0" ]
         else
           data_tar_flags += [ "--group", attributes[:deb_group] ]
         end

```

jordansissel:
The problem I think is that the default deb_group is `root` and therefore (in master) adds the `--numeric-owner` flag when considering group ownership. My patch removes the `--numeric-owner` flag from the group setting, though I'm not sure if that's the best fit.

Results after my patch:

```
# Give --deb-user jls
% bin/fpm -s dir -t deb -n fizzle -f --deb-user jls ~/.zshrc ; dpkg -c fizzle_1.0_amd64.deb | grep .zshrc
...
-rw-r--r-- jls/root      12153 2015-07-17 00:26 ./home/jls/.zshrc

# defaults to use root, but numeric uid/gids:
% bin/fpm -s dir -t deb -n fizzle -f ~/.zshrc ; dpkg -c fizzle_1.0_amd64.deb | grep .zshrc           
...
-rw-r--r-- 0/0           12153 2015-07-17 00:26 ./home/jls/.zshrc
```

jordansissel:

PR #942, opened 2015-06-17
Ensure --deb-shlibs file permission to keep lintian happy

lovell:
Hello, this change prevents `lintian` (and therefore Ubuntu Package Manager) generating the
```
control-file-has-bad-permissions shlibs 0664 != 0644
````
error and the subsequent "violates the quality standards" alert when the `--deb-shlibs` option is used.

Originally discovered via https://github.com/dimensio/libpng16-deb/issues/1

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -714,6 +714,7 @@ class FPM::Package::Deb < FPM::Package
     File.open(control_path("shlibs"), "w") do |out|
       out.write(attributes[:deb_shlibs])
     end
+    File.chmod(0644, control_path("shlibs"))
   end # def write_shlibs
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by jordansissel** [comment]
Code lgtm. <3

**Comment by jordansissel** [comment]
I'd love to see tests for this (there's already lintian tests, but they don't cover deb shlibs), but wont' block the merge.

**Comment by lovell** [comment]
Great, thanks for merging this Jordan.

Would adding something like
```ruby
subject.attributes[:deb_shlibs] = "test 1 test (= 1.0.0)"
```
approximately [here](https://github.com/jordansissel/fpm/blob/master/spec/fpm/package/deb_spec.rb#L352) provide at least some test coverage?

PR #946, opened 2015-06-23
Fix file names with closing braces in their name.

### LINKED ISSUE #945: Glob expressions fail during RPM build for files that do actually exist
I am trying to package files that include braces and spaces in their name, and FPM is producing escaped glob expressions that are causing rpmbuild to fail to match file names.

```
{:timestamp=>"2015-06-23T05:44:22.127963+0100", :message=>"    File not found by glob: /tmp/package-rpm-build20150623-4696-1pedwym/BUILD/user1/live/sw/StdData/GB_wpphead/GB_wpphead_[\\[]][\\[]Supp Del Qty Anal].xdat", :level=>:info, :file=>"cabin/mixins/pipe.rb", :line=>"47", :method=>"block in pipe"}
```

iwonbigbro:
Fixes #945

### DIFF
--- a/lib/fpm/package/rpm.rb
+++ b/lib/fpm/package/rpm.rb
@@ -164,7 +164,7 @@ class FPM::Package::RPM < FPM::Package
   # Replace % with [%] to make rpm not expand macros
   def rpm_fix_name(name)
     name = "\"#{name}\"" if name[/\s/]
-    name = name.gsub(/([\[\]])/, '[\\\1]')
+    name = name.gsub(/(\[|\])/, { '[' => '[\\[]', ']' => '[\\]]' })
     name = name.gsub("*", "[*]")
     name = name.gsub("?", "[?]")
     name = name.gsub("%", "[%]")
--- a/lib/fpm/package/rpm.rb
+++ b/lib/fpm/package/rpm.rb
@@ -163,11 +163,15 @@ class FPM::Package::RPM < FPM::Package
   # Replace ? with [?] to make rpm not use globs
   # Replace % with [%] to make rpm not expand macros
   def rpm_fix_name(name)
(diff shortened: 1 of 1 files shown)


lib/fpm/package/rpm.rb
```diff
@@ -163,11 +163,15 @@ class FPM::Package::RPM < FPM::Package
   # Replace ? with [?] to make rpm not use globs
   # Replace % with [%] to make rpm not expand macros
   def rpm_fix_name(name)
-    name = "\"#{name}\"" if name[/\s/]
-    name = name.gsub("[", "[\\[]")
-    name = name.gsub("*", "[*]")
-    name = name.gsub("?", "[?]")
-    name = name.gsub("%", "[%]")
+    name = name.gsub(/(\ |\[|\]|\*|\?|\%|\$)/, {
+      ' ' => '?',
```
jordansissel: This isn't correct, I don't think. Two files: 'foo bar' and 'foo.bar' will be matched by the result of this.

I'm not really sure what the right solution here is given rpmbuild's bugs with quoted strings and wildcards.

iwonbigbro: Perhaps we need an option to explicitly define the behavior. But for me, in
mode -t dir, I would expect it to be harmless if it did match both "foo
bar" and "foo.bar". I've made the changes locally, so I'm happy with the
behavior. But it is really broken without the merge request if you have a
file named "foo [*baz%] bar". The quoting of the name stops the glob
expressions from working.

jordansissel:
Hmm, I suppose we could try this merge and see how folks are affected. I don't know if anyone would necessarily be broken by this change. We can revert or revisit it if it causes problems.

Thank you for this patch! LGTM

PR #947, opened 2015-06-25
Local project path support for virtualenv target.

vi4m:
Sometimes you don't want to upload to pypi server just to make distro package. 
Special flag --local-project-path just does this - like  `pip install -e.`


vi4m:
@djhaskin987 could you take a look at this?

djhaskin987:
It looks good to me. I have only a few questions. If these questions are answered, I would feel comfortable with this patch, but of course the final say goes to our chief @jordansissel .

### Question 1

[Pip's documentation](https://pip.pypa.io/en/latest/reference/pip_install.html#usage) says that `-e` means installing something in editable mode:

> -e, --editable <path/url>
> Install a project in editable mode (i.e. setuptools "develop mode") from a local project path or a VCS url.)

Does that mean that pip assumes interaction with this command? Unless `-e` is supplied to fpm itself, I don't think the command should be interactive, but this point is moot if you know `pip -e` is not.

### Question 2

Will `pip -e` search pypi for dependencies of that local project and install them too? If so, it should supply the `-i` option with `attributes[:virtualenv_pypi]` as an argument so the packages are pulled from the right place. If not, we have a problem, since the whole point of a virtualenv is that it is self-contained. You'll be able to tell me if I'm just worrying over nothing, though :)

### Question 3

Does this mean that you still want the supplied `package` name (with optional supplied `package_version`) to be the name of the virtualenv package? That is what will happen, as coded: It will grab the python code from the local project path, but if `foo==3.0` is supplied, the virtualenv package will be called `virtualenv-foo` version `3.0`.

jordansissel:
@djhaskin987 thanks for reviewing!

I have an additional question about the user interface. Can we achieve this without adding any flags? I'm not familiar with this use case, but with the 'python' support in fpm, you can give the path to a directory instead of a name of a module to fetch from pypy. For example, if you do `fpm -s python -t deb something` and `something/setup.py` exists, fpm will try to use it. Can we achieve something similar for virtualenv (which won't add flags?)?

Happy to help brainstorm here. Further, I am in favor of this improvement and also will accept additional flags if we cannot achieve this without flag changes. Let me know how I can help :)

zoresvit:
Huh, `virtualenv` source in FPM appears to be a little bit confusing. Intuitively I assumed that #930 by default actually does just this — you provide it with local `virtualenv` and it creates a package out of it. Logical, isn't it? Instead it does exactly the same thing as `-s python`, but also bakes in the dependencies using virtualenv.

However the usecase when one wants to get package from Pypi with all its dependencies baked in is also very valid (very useful for old-school systems like CentOS to bring in some newer packages without dependency hassle).

But I think current change (#947) should actually relate to #930 and essentially be the default behavior. Because what  #930 currently does looks much more similar to `-s python`, just with dependencies backed in right away.

So, @jordansissel, @djhaskin987, do you think it would make sense to tweak these a bit, precisely:
1. Move the ability to create target from Python package with dependencies baked in to `-s python` and add flag like `--with-dependencies`.
2. Make the default behavior of `-s virtualenv` to create package from locally created virtualenv?

I'm not much into Ruby, but if you think it would make more sense (it seems to me), I could look into it.

Thanks!

kozmaz87:
I am working on this and I was wondering if it was ok to reuse some of the existing functionality in the python.rb module or should I just copypaste stuff over to virtualenv.rb? Thing is there is quite a bit of heuristics that the python.rb seems to be doing in order to get package meta information.

To the questions above pip install -e is a bad approach. The pull request above does not provide the solution you want. pip install -e is equivalent to python setup.py develop. You do not want to do that. That would link in your package to the virtualenv but the virtualenv won't contain it. pip install accepts stuff like: file:// and also accepts the folder name where the setup.py is located The only problem is that then the code around line 81 complains that it cannot regex the package version out of pip freeze. I was thinking of lifting that information over from python.rb that already has that kind of information...

vi4m:
I tried approach without "-e" wich uses local file:/// protocol and it worked as expected. I couldn't reproduce the 81 line error, as you wrote @kozmaz87.

zmingee:
+1

All of the develop that we do at my company is internal-only. I'd rather not use a private pypi server.. That's why I'm using fpm in the first place. Getting this working/merged would be great for me.

vi4m:
Another idea.

Well, the more "flexible" solution would be to just use `pip install -r requirementes.txt`  instead, you can then decide if use local projects paths or pypi server ones or mix of both. 

Example:
> cat requirements-local.txt

    file:///home/user/directory/python-markdown-graphviz@#egg=python-markdown-graphviz
    git+https://github.com/vi4m/python-markdown-graphviz@#egg=python-markdown-graphviz

What do you think @djhaskin987 ?

beddari:
The workflow @vi4m describes is the one I would prefer.

alexprengere:
+1, what @zmingee said ;).

The development we do at my company is internal-only. We do however use private pypi servers, but still there is a need to build a package from a local directory for development and testing purposes.

djhaskin987:
Sorry this has taken so long.

@zoresvit , it is possible to do what you are asking (`-s python --with-dependencies`), but not as coded. The functionality found in `lib/fpm/package/virtualenv.rb` as what is presently in the master branch would have to be merged into `lib/fpm/package/python.rb`. I know the two files are similar, so this might be possible. It deserves some research.

@vi4m , I think your suggestion is a good one. In fact, I think I see a resolution here. Consider the following idea (your idea, stolen and put into another form):

if the user provides `fpm -s virtualenv -t deb ./requirements.txt`; in short, if the provided argument is a text file which exists on the system, then virtualenv uses the requirements.txt to figure out what to install. Else, it assumes it is the name of a package and attempts to install it from pypi. This can be accomplished without any additional arguments, does not violate least-surprise, and as you say, it provides more flexibility since you can mix and match what you want installed. In general, I am in favor of this idea @jordansissel , for whatever that is worth :P

alexprengere:
I hope I did not miss anything here, but I think using *only* the ``requirements.txt`` is a bit restrictive, because:

* some packages list their dependencies in the ``setup.py`` using ``install_requires``
* some packages have a ``setup.py`` file that reads the ``requirements.txt`` to do the same
* some packages simply have a ``requirements.txt``

Regarding this, I believe the more general approach is to let ``pip`` decide what to do, e.g. perform a ``pip install .`` after activating the virtualenv.

On a side note, I believe the default behavior for local paths handling should be (for ``-t virtualenv`` and ``-t python``):
* check first if the argument exists locally as a directory, if so, use it
* if not, search on PyPI (or else if specified)

djhaskin987:
@alexprengere It need not be just requirements.txt . It would be easy to determine if the given argument were a directory instead of a text file for example, and then treat that directory as if it were a local python project and use it instead of the requirements.txt file when calling pip, no?

jordansissel:
I trust y'all to figure out when this is ready for merging; just pign me when you want it merged! :)

djhaskin987:
I have thought long and hard about a solution.

@vi4m , what do you think about this suggestion?

The `input` function takes a string parameter. Right now, that variable is called `package`.

We can test to see if the string `package` describes a directory. If it does, we can run a series of checks to see what to do.

If the file `$package/bin/pip` exists and is executable, we may assume we were given the path to a local virtualenv, and if so, the command `$package/pip/freeze` can be used to determine what to install to the staging directory in preparation for packaging.

If the file `$package/requirements.txt` exists, we may assume the directory is actually a python project, and we may use its contents to determine what to install into the staging directory.

Finally, if the file `$package/requirementes.txt` does not exist, but the file `$package/setup.py` does, we may use the output of the command `$package/setup.py --install-requires` to determine what pip packages to install into the staging directory. We should also in that case use the python interpreter installed  to the staging directory as part of the installed virtualenv and use it to run `$package/setup.py install` to install the project itself into the staging directory virtualenv.

Additionally, if `$package/setup.py` exists (whether or not `$package/requirements.txt` exists), we can use `$package/setup.py --name` and `$package/setup.py --version` to set the name and version of the virtualenv package if they have not already been set.

If none of the above files are found, but `$package` is a directory, an error occurrs.

If `$package` does not describe a directory (no such directory exists), but describes a text file, we can assume it points to a `requirements.txt`-like file and use its contents to populate the virtualenv in the staging directory.

Finally, FINALLY, if `$package` is not a file at all, we may assume it describes a pip package (or list of pip packages) and use it to create a virtualenv (as already coded).

Does that make sense? I know I wrote a book, but I feel my suggestion warrants discussion.

alexprengere:
Wow! It looks really good to me. I wonder if we could improve the virtual detection a little bit. Looking for a `$package/bin/pip` in the local directory might miss some common patterns I came across:
* some projects put the virtual environment outside of the local directory, like [virtualenvwrapper](https://virtualenvwrapper.readthedocs.org/en/latest/)
* others expect the virtualenv to be stored locally in a hidden folder, like [autoenv](https://github.com/kennethreitz/autoenv) (`.env`)

Unfortunately I do not see a simple and global detection method. Some ideas:
* relying on `$VIRTUAL_ENV`. This would require sourcing the `activate` script before calling `fpm`
* looking for the `bin/pip` with a `find` in the local directory
* using an extra flag or parameter to specify the path to the `bin/pip`
* using a convention regarding where `fpm` looks for the `bin/pip` (like in the `$package/bin/pip` :smile:)

Any ideas regarding this?

djhaskin987:
@alexprengere I think getting too magic about how we find the virtual environments is a mistake, because if may confuse users. In particular:
  * Relying on environment variables quicly leads to confusing behavior. The fpm command should "just work" regardless of when `activate` was called. If the user did call `activate`, they can just feed the location of the virtualenv to fpm using `fpm -s virtualenv -v rpm $VIRTUAL_ENV`. This way, they know exactly what they are packaging -- the virtualenv at that location.
  * Using `find` can, I forsee, quickly frustrate users, and lead to nondeterministic behavior. Suppose the user wants to package the local python project found at `a/`. Does fpm use its virtualenv, found at `a/ve/bin/pip`, or does vpm use its requirements.txt, found at `a/requirements.txt`? If FPM doesn't use `find` it lets the user decide this question by letting the user pick what input is given.

As you said, other than doing basic file checks, there is no simple global detection method. I feel it is best simply to let the user explicitly tell fpm what to package up. fpm can figure out what KIND of thing the user wants packaged (a local project or a virtualenv folder), but should not try to guess at WHAT the user wants packaged up. Also, because there is such a thing as `$VIRTUAL_ENV`, the user can use that in the commandline to tell fpm what to package, even if they're using something like `virtualenvwrapper`.

jordansissel:
This PR fell a bit out of sync with master and will not merge due to conflicts, just FYI.

jordansissel:

nicobrevin:
#1131 adds support for building a virtualenv from requirements.txt, instead of a package from pypi.  I kind of agree that the virtualenv support is weird, in that you can't just bundle your requirements *with* your application, but that's not too hard to do using a Makefile.  It also just seems to be one of the less-nice things about virtualenvs compared to, say, using `bundler --standalone`

Kyrluckechuck:
Any chance this is going to get picked up again?

jordansissel:
If someone reviews and tests this to their satisfaction, we could merge it
-- no rush -- just want to open invitation to review or test or fork the pr

djhaskin987:
I'll have some time this weekend, I'll see what I can do @Kyrluckechuck  :)

Kyrluckechuck:
I tried taking a brief look at using it as it is now, but there is too much missing and it's in need of a git rebase (such as being able to install to systemd, using extra-pypi urls for the other packages, etc.). 

I attempted it but frankly do not have enough Ruby know-how or experience for this type of project.

If anyone is able to rebase it to be more up-to-date with master, I can give it a whirl/test!

vi4m:
Thanks everyone for discussion about this issue. I'm closing this PR, since there is another PR, rebased.

PR #950, opened 2015-07-20
can't create packages with OUTPUT_TYPE 'virtualenv' from non-public python repos

skoenig:
I became aware that it is impossible to create a package from a python module which is just stored in a private repository because when you pass the option '--pypi' to fpm with a custom repo it tries to install 'pip' and 'distribute' from the same repo which always fails if you don't include this modules into your custom repo.
It seems like an uglish hack which is also stated at the comment https://github.com/jordansissel/fpm/blob/master/lib/fpm/package/virtualenv.rb#L69.
So either the `pip` and `distribute` modules are not found in your custom repos (why should they be installed at all, `pip` is already installed in the virtualenv and `distibute` is uninstalled a line later), or the custom module which isn't available on https://pypi.python.org/pypi will not be found and the package build process will abort.
I'd suggest to add the option '--pypi-extra-index-url' so that the custom / private repo could be really only additional to the standard pypi repos.
I'd also commit a pull request for this feature, because it prevents me to build custom packages from our own pypi repositories.

### DIFF
--- a/lib/fpm/package/virtualenv.rb
+++ b/lib/fpm/package/virtualenv.rb
@@ -12,6 +12,9 @@ class FPM::Package::Virtualenv < FPM::Package
   option "--pypi", "PYPI_URL",
   "PyPi Server uri for retrieving packages.",
   :default => "https://pypi.python.org/simple"
+  option "--pypi-extra-index-url", "PYPI_EXTRA_INDEX_URL",
+  "Extra URLs of package indexes to use.",
+  :default => nil
   option "--package-name-prefix", "PREFIX", "Name to prefix the package " \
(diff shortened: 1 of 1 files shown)


skoenig:
This is actually not needed, instead of passing another mirror via command line option `--extra-index-url` it can be set in `/etc/pip.conf`:

    [install]                                              
    index-url = https://pypi.python.org/simple/            
    extra-index-url =  https://my.custom.repo/pypi/

This works perfectly and is sufficient for our needs - without adding another option to the already crowded `fpm` CLI.

jordansissel:
@skoenig oh cool, I'm glad you found a workaround! Thank you for sharing it so others can use it too! <3

PR #962, opened 2015-07-27
RPM summary

### LINKED ISSUE #903: description and summary information for RPMs
Hi,

While creating my first RPM, I noticed that the *first line* of the multi-line description is automatically
used for the Summary.

Would it be possible to add a separate "--summary" option?
*** OR, is this already possible and I am missing it ?!? :)

Regards,
Shantanu

hatt:
This turned out a bit more complex than expected since defaulting to another attribute turns out to not actually work the way I thought it did. Feedback from maintainers would be good since I'm sure there's a better way to make this work in less code. Resolves #903.

### DIFF
--- a/lib/fpm/package/rpm.rb
+++ b/lib/fpm/package/rpm.rb
@@ -86,6 +86,9 @@ class FPM::Package::RPM < FPM::Package
     File.read(File.expand_path(file))
   end
 
+  option "--summary", "SUMMARY",
+    "Set the RPM summary. Overrides the first line on the description if set"
+
   option "--sign", :flag, "Pass --sign to rpmbuild"
 
   option "--auto-add-directories", :flag, "Auto add directories not part of filesystem"
@@ -512,6 +515,14 @@ class FPM::Package::RPM < FPM::Package
     #return File.join("BUILD", prefix)
   end # def build_sub_dir
 
+  def summary
+    if !attributes[:rpm_summary]
+      return @description.split("\n").first || "_"
+    end
+
+    return attributes[:rpm_summary]
+  end # def summary
+
   def version
     if @version.kind_of?(String) and @version.include?("-")
       logger.warn("Package version '#{@version}' includes dashes, converting" \
@@ -552,5 +563,5 @@ class FPM::Package::RPM < FPM::Package
 
   public(:input, :output, :converted_from, :architecture, :to_s, :iteration,
          :payload_compression, :digest_algorithm, :prefix, :build_sub_dir,
-         :epoch, :version, :prefixed_path)
+         :summary, :epoch, :version, :prefixed_path)
 end # class FPM::Package::RPM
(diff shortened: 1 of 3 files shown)


lib/fpm/package/rpm.rb
```diff
@@ -512,6 +515,14 @@ def build_sub_dir
     #return File.join("BUILD", prefix)
   end # def build_sub_dir
 
+  def summary
+    if !attributes[:rpm_summary]
+      return @description.split("\n").first.empty? ? "_" : @description.split("\n").first
```
jordansissel: See #973. @description could be an empty string and this will crash:

```
>> "".split("\n").first
=> nil
```

The crash will be `NoMethodError: undefined method `empty?' for nil:NilClass`

spec/fpm/package/rpm_spec.rb
```diff
@@ -39,6 +39,19 @@
...
+  describe "#summary" do
+    it "should default to description" do
+      expected = subject.description
+      insist { subject.summary } == expected
+    end
+
+    it "should return description override" do
+      subject.attributes[:rpm_summary] = "a summary"
+      expected = subject.description
+      insist { subject.summary } != expected
```
jordansissel: Don't you want to assert that it equals subject.attributes[:rpm_summary] ?

jordansissel:
Overall the code looks good! Some minor comments. Love the tests. I'm done with this round of review. @hatt - back to you! :)

jordansissel:
(You'll probably have to rebase because #973 changed how summary is computed)

hatt:
Thanks @jordansissel, this definitely needed review when I submitted it since it seemed more complex than required, should have it cleaned up in the next few days. Cheers!

spec/fpm/package/rpm_spec.rb
hatt: I've changed it to subject.attributes[:rpm_summary] but setting the attribute alters the summary in the subject anyway since subject.summary is a method that returns the summary and not the @summary class attribute I think? If you could confirm that'd be good.

hatt:
Changes made to bring inline with #973, I don't think there is any issue with the tests as in my comment above.

templates/rpm.erb
```diff
@@ -43,7 +43,7 @@ Epoch: <%= epoch %>
 <% end -%>
 Release: <%= iteration or 1 %><%= "%{?dist}" if attributes[:rpm_dist] %>
 <%# use the first line of the description as the summary -%>
-Summary: <%= description.split("\n").first || "_" %>
```
v6: // , In older versions, this works as a way to add a summary by using the part in the --description argument before the newline, right?

hatt: Yes. The behavior is still the same, just the location has changed. A summary method was added which defaults to the first line of the description, or otherwise the summary argument. If no summary or description are passed, then the underscore is used. See `lib/fpm/package/rpm.rb` in this PR for it.

v6:
// , Who's the integrator for this? What would it take for it to get merged?

jordansissel:
Code LGTM; haven't run the tests. Will merge :)

Issue #976, opened 2015-08-06
The flags in the .fpm file cannot contain source files


mildred:
Follow up from issue #944

It is not possible to have a package with a .fpm file containing:

    -s dir -C root-dir .

(assuming root-dir contains the root hierarchy of the package to create)

Running `fpm -t deb` in that folder results in:

    Loading flags from rc file .fpm {:level=>:warn}
    All flags should be before the first argument (stray flags found: ["-t"] {:level=>:warn}
    Missing required -t flag. What package output did you want? {:level=>:warn}
    Fix the above problems, and you'll be rolling packages in no time! {:level=>:fatal}

This is because the command line is transformed into:

    fpm -s dir -C root-dir . -t deb

(the `-t deb` part is out of order and should appear before `.`)

Perhaps argument merging should be a little more intelligent. Or we can define that the second line on the .fpm file should be added to the end of the parameter list (while the first line is added to the beginning just as now).

jordansissel:
Agreed this looks like a big we can fix.

PR #977, opened 2015-08-11
.fpm parsing: separate options and arguments and add FPMOPTS handling

mildred:
Allow arguments in .fpm files, old behaviour put those arguments in the middle of the command line, and further options were not recognized. Fix issue #976

Also, add optional default arguments handling from the FPMOPTS environment variable. This allows default arguments to be specified in regard to the .fpm file. This could be useful for a build bot that can specify a default version number (git revision) in case the .fpm file doesn't specify it.

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -509,7 +509,7 @@ class FPM::Command < Clamp::Command
     end
   end # def execute
 
-  def run(*args)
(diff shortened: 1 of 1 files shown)


jordansissel:
I like both ideas. I haven't tested this, but the code looks ok to me :)

Thank you :)

PR #978, opened 2015-08-12
Clarified tempdir error message

lepht:
Per the [source for ::Dir.tmpdir](http://ruby-doc.org/stdlib-1.9.3/libdoc/tmpdir/rdoc/Dir.html#method-c-tmpdir), these are the actual environment variables that should be checked:

```
    for dir in [ENV['TMPDIR'], ENV['TMP'], ENV['TEMP'], @@systmpdir, '/tmp']
```


jordansissel:
LGTM

PR #982, opened 2015-08-18
Fix broken behaviour in --excludes-file option

wyaeld:
`--exclude-file` doesn't work due to a variable error on L367

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -364,7 +364,7 @@ class FPM::Command < Clamp::Command
       end
 
       # Read each line as a path
-      File.new(exclude-file, "r").each_line do |line| 
+      File.new(exclude_file, "r").each_line do |line|
         # Handle each line as if it were an argument
         # 'excludes' is defined above near the -x option.
         excludes << line.strip
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -363,11 +363,13 @@ class FPM::Command < Clamp::Command
         return 1
       end
 
+      # Ensure hash is initialized
+      input.attributes[:excludes] ||= []
+
       # Read each line as a path
       File.new(exclude_file, "r").each_line do |line|
         # Handle each line as if it were an argument
-        # 'excludes' is defined above near the -x option.
-        excludes << line.strip
+        input.attributes[:excludes] << line.strip
       end
     end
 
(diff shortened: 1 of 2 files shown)


wyaeld:
hmm, fixing the typo is just resulting in further problems, so this fix isn't complete yes

wyaeld:
Now its working. 

exclude-file implementation was always broken, @ahammond missed one of the 3 variables in https://github.com/jordansissel/fpm/commit/a88027e831036230f166ab94fc8827131dd37f39

exclude-file was trying to set the attributes on the exclude array from https://github.com/jordansissel/fpm/blob/master/lib/fpm/command.rb#L138 which @jordansissel introduced in https://github.com/jordansissel/fpm/commit/cfee4c7fad315e56364c8650032c35dbec6ed01f but it doesn't seem like later methods in the class can reference that, because its nil when exclude-file tries.  Maybe that only works if both exclude and exlude-file are used together?  I could use some advice

I changed path removal from remove_entry_secure to rm_r
Have been running into what looks like the same problem the nanoc guys faced here, https://github.com/nanoc/nanoc/commit/aeb71156d6a97851f9f262ae68eb09074432b314 so I borrowed their solution.

In my case /tmp was definitely modded to 1777 but was still hitting the problem.

wyaeld:
A few other notes.

Path globbing with `File.fnmatch` isn't very intuitive, and is quite different from tar/git/docker

Here is a working excludes file with a few combinations

```
**.git
**.vagrant
**dist
**docs
**archive/**
**input/**
**output/**
**run/**
**etc/results.conf
```

You can ignore entire dirs, contents of dirs, or individual files still, but particularly if you've used the `--directories` option, then its going to be path patching based on the OUTPUT file structure, not the source structure, so that gets confusing, and probably why someone trying to exclude `.git` gets nowhere

wyaeld:
oh yeah, and if you try to exclude `tmp`, everything dies

wyaeld:
@jordansissel is this just dead now?  Its a pain having to run all our packaging off a fork if no-one is going to look at the fix

jordansissel:
>  Its a pain having to run all our packaging off a fork 

Sorry for your frustration.

> if no-one is going to look at the fix

I haven't worked on fpm lately because of a lack of energy and lack of time to do so. In the meantime, I invite anyone else to review + test this fix for style, backwards-compatibility, added tests, etc. If someone else reviews this and says it works well and doesn't break anything else, I'll be happy to click the merge button.

imoore76:
I can vouch for it. This works well and the merge would be very much appreciated! We're using fabric to launch fpm, which won't pass multiple flags with duplicate names. E g multiple --exclude flags. A single --exclude-file does the trick though

jordansissel:
@imoore76 thaks for the confirmation!

@wyaeld thank you for your efforts here. LGTM! :)

PR #986, opened 2015-08-20
Install local module if package is file

mdom:
I just removed the unused switch --cpan-local-module and install package if it's file or directory.


jordansissel:
I like this change. Thanks! <3

PR #1001, opened 2015-09-07
add rpm internal names as comments

glensc:
adds comment in code how the fpm triggers are mapped to rpm names

as it's not document anywhere of the mapping, and coming from rpm world it's hard to understand what fpm names mean.


hatt:
+1, I never got what this code did and TIL

jordansissel:
++++ thank you

In truth, I think this code really belongs in the `arr-pm` library, but for whatever reason this coded landed in fpm instead, hehe. 

<3

PR #1002, opened 2015-09-07
Add Architecture to rpm output

### LINKED ISSUE #999: Cannot specify architecture 'atom' when making RPMs
Result is RPMs won't install--they end up with an odd string instead of 'atom' as the architecture.

epylar:
Resolve #999


epylar:
Ah, looks like it's more complicated than this.  Closing until I get it working properly, or someone else does...

PR #1006, opened 2015-09-14
Feature/shell template improvements

ehealy:
A couple of small improvements and some new functionality that makes the package metadata useful in the post install actions.


ehealy:
@jordansissel

jordansissel:
Haven't tested; but change looks ok :)

Thanks! <3

ehealy:
@glensc apologies for the delayed response. I'll see what I can do about fortifying this and increasing test coverage. The respond_to? logic was added in order to allow fpm to be backwards compatible with some older packages we had created with it prior to the addition of metadata to packages.

glensc:
just remember, empty functions are not allowed in shell syntax. this results syntax error:

```sh
foo() {
}
```

PR #1008, opened 2015-09-16
Change FreeBSD to use gtar

liv3d:
The FreeBSD version of tar doesn't include the options --owner and
--group among others. Instead change to use gnutar as the default.
This adds a dependency of installing
https://freshports.org/archivers/gtar/ for anyone creating packages
under FreeBSD (similar to doing brew install gnu-tar under OSX)


jordansissel:
LGTM!

PR #1012, opened 2015-09-17
add support for pypi --extra-index-url

paul-krohn:
Add support for URLs passed to pip with pypi --extra-index-url, so you can include code from both public & private PyPi repos.

### DIFF
--- a/lib/fpm/package/virtualenv.rb
+++ b/lib/fpm/package/virtualenv.rb
@@ -25,6 +25,10 @@ class FPM::Package::Virtualenv < FPM::Package
   option "--other-files-dir", "DIRECTORY", "Optionally, the contents of the " \
   "specified directory may be added to the package. This is useful if the " \
   "virtualenv needs configuration files, etc.", :default => nil
+  option "--pypi-extra", "PYPI_EXTRA_URL",
+  "PyPi extra-index-url for pointing to your priviate PyPi",
+  :multivalued => true, :attribute_name => :virtualenv_pypi_extra_index_urls,
+  :default => nil
 
   private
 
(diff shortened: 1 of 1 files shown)


skoenig:
Perhaps not necessary, I just closed my PR https://github.com/jordansissel/fpm/pull/950 because I found another solution (see comment).

jordansissel:
I'm open to merging this. Code looks OK. I haven't tested.

lib/fpm/package/virtualenv.rb
```diff
@@ -25,6 +25,10 @@ class FPM::Package::Virtualenv < FPM::Package
   option "--other-files-dir", "DIRECTORY", "Optionally, the contents of the " \
   "specified directory may be added to the package. This is useful if the " \
   "virtualenv needs configuration files, etc.", :default => nil
+  option "--pypi-extra", "PYPI_EXTRA_URL",
```
jordansissel: Curious, Is this a good name for the flag? Would it be more descriptive as --pypi-extra-url ? Thoughts? :)

lib/fpm/package/virtualenv.rb
```diff
@@ -25,6 +25,10 @@ class FPM::Package::Virtualenv < FPM::Package
...
+  "PyPi extra-index-url for pointing to your priviate PyPi",
```
jordansissel: Indent these 3 lines (the help text, multivalued, and default) by 2 more spaces

lib/fpm/package/virtualenv.rb
```diff
@@ -72,9 +76,14 @@ def input(package)
                "pip", "distribute")
     safesystem(pip_exe, "uninstall", "-y", "distribute")
 
-    safesystem(pip_exe, "install", "-i",
-               attributes[:virtualenv_pypi],
-               package)
+    extra_index_url_args = []
+    if !attributes[:virtualenv_pypi_extra_index_urls].nil?
```
jordansissel: Suggestion - instead of checking !foo.nil?, you can just check foo:

```
# if !attributes[:virtualenv_pypi_extra_index_urls].nil?
if attributes[:virtualenv_pypi_extra_index_urls]
```

Non-nil values are "truthy" in ruby. Just an FYI, no code change required.

paul-krohn:
Thanks for the feedback; the solution/workaround of creating a config file doesn't work on hosted CI engines (Travis, Circle, etc).  In the mean time I needed a [solution](https://github.com/paul-krohn/python-virtualenv-packager); it does a couple more things like guessing the package name and version; any interest in porting any of that to get_metadata.py?

jordansissel:
LGTM! Thanks much :)

PR #1014, opened 2015-09-23
several improvement regarding the handling of Debian Conf file declar…

elvido:
Improved the handling of Debian package config files and fixed some minor issues:

  - init-scripts, default and upstart specified as command line option are automatically added as declared debian conf file
  - this behavior can be disabled using "--deb-no-auto-config-file"
  - reordered the process steps during package generation, thus init-script and config-files options can be used in parallel.
    without that fix, you will get an error that the script file can not be found during the processing of the config-files options

All changes are related processing Debian packages and should be backward compatible.

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -128,6 +128,11 @@ class FPM::Package::Deb < FPM::Package
     "Do not add all files in /etc as configuration files by default for Debian packages.",
     :default => false
 
+  option "--no-auto-config-files", :flag,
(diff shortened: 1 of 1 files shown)


jordansissel:
LGTM, but I"ll make some minor style changes after merging (fixing spacing, changing 'no_auto...' to be 'auto_...')

Thank yoU!

elvido:
BTW - FPM is a really cool thing!

Cheers, Ralf

PR #1027, opened 2015-11-03
Update --description in help to display a literal \n instead of a newline

marcteale:
The help currently displays a line break instead of a literal \n, this replaces \n with \\n.


jordansissel:
Thanks! <3

PR #1029, opened 2015-11-12
Add support for using python-package-name-prefix for python dependencies

tberg:
if --rpm-prefix-python-dependencies is specified, then the python-package-name-prefix value is prepended to auto-calculated dependencies.

This simplifies the packaging process for CentOS SCL python packages.

I am making possibly large assumptions about the format of the dependency strings, unsure how safe that is.


tberg:
This feature is already implemented with --python-fix-dependencies.

PR #1030, opened 2015-11-13
gem: make it posssible to set the shebang

ZeroPointEnergy:
This makes it possible to create packages from gems for ruby versions
which are not currently installed on the machine you build the package on.

Tests are included.

### DIFF
--- a/lib/fpm/package/gem.rb
+++ b/lib/fpm/package/gem.rb
@@ -30,6 +30,9 @@ class FPM::Package::Gem < FPM::Package
   option "--gem", "PATH_TO_GEM",
           "The path to the 'gem' tool (defaults to 'gem' and searches " \
           "your $PATH)", :default => "gem"
+  option "--shebang", "SHEBANG",
+          "Replace the shebang in the executables in the bin path with a " \
+          "custom string", :default => nil
   option "--fix-name", :flag, "Should the target package name be prefixed?",
     :default => true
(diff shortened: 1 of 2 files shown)


lib/fpm/package/gem.rb
```diff
@@ -196,6 +199,22 @@ def install_to_staging(gem_path)
...
+    if attributes[:gem_shebang]
+      ::Dir.entries(bin_path).each do |file_name|
+        # exclude . and ..
+        next if ['.', '..'].include?(file_name)
+        # exclude everything which is not a file
+        file_path = File.join(bin_path, file_name)
+        next unless File.ftype(file_path) == 'file'
+        # replace shebang in files if there is one
+        file = File.read(file_path)
+        if file.gsub!(/^#!.*$/, "#!#{attributes[:gem_shebang]}")
```
jordansissel: this will replace all lines starting with `#!` in the file and probably isn't what you want. I think this should only attempt to change the first line of the file.

jordansissel:

ZeroPointEnergy:
Just saw this. currently away from work, will rebase once I am back in the office and also fix that it only changes the first line of a file. Sorry for the delay

ZeroPointEnergy:
Branch is now rebased. Also fixed the code so that only the shebang on the first line is replaced

jordansissel:
Thanks! :)

PR #1032, opened 2015-11-18
Get package name from setup.py if its a dir

### LINKED ISSUE #949: virtualenv fails if version is not specified
When trying to install a package from a checked out repository (in this case https://github.com/diraimondo/gmusicproxy), the build fails with:

    /home/vagrant/fpm/lib/fpm/package/virtualenv.rb:81:in `input': undefined method `split' for nil:NilClass (NoMethodError)

Around this line we see:

    if package_version.nil?
      frozen = safesystemout(pip_exe, "freeze")
      package_version = frozen[/#{package}==[^=]+$/].split("==")[1].chomp!
      self.version ||= package_version
    end

The command line was

    bundle exec fpm --debug -s virtualenv -t deb gmusicproxy/

Version of fpm:

    vagrant@vagrant-ubuntu-trusty-64:~/fpm$ git show-ref HEAD
    24859a16463459fcbe045e610bccdf6d04dbc5d3 refs/remotes/origin/HEAD

russell:
Fixes #949

If the package name contains slashes and is an existing directory, then
pip will install from that, so we should get the package name from there
too.

Also please pay close attention to the use of the chomp! on line 94 since it never seemed to return a value for me, but I'm not totally sure what the intent was using chomp! so i may have broken something.


lib/fpm/package/virtualenv.rb
```diff
@@ -85,9 +85,13 @@ def input(package)
...
     safesystem(*pip_args.flatten)
 
+    if package =~ /\// and File.directory?(package)
+      package_name = safesystemout("cd #{package} && #{python_exe} setup.py --name | tail -1").strip
+    end
+
     if package_version.nil?
       frozen = safesystemout(pip_exe, "freeze")
-      package_version = frozen[/#{package}==[^=]+$/].split("==")[1].chomp!
+      package_version = frozen[/#{package_name}==[^=]+$/].split("==")[1].strip
```
jordansissel: I think package_name won't be set here if the `package` is not a directory, right? Won't `package_name` be unset otherwise?

jordansissel: Good catch on the '.chomp!' returning nil if no chomping actually occurs. That was a bug, oops!

jordansissel:
Testing this fails for me, maybe I'm doing something wrong:

```
% (cd /tmp/python-rightscale; python setup.py --name)
rightscale

% bin/fpm -s virtualenv -t deb /tmp/python-rightscale/
Parent directory does not exist: virtualenv-/tmp/python-rightscale - cannot write to virtualenv-/tmp/python-rightscale/_0.309_amd64.deb {:level=>:fatal}
```

Thoughts?

jordansissel:

josephholsten:
@russell I rebased this into #1377, I'd love your :eyes: on it.

russell:
Fpm is a great tool.  But I don't use Fpm any more, I can rebase this and get it working, but it's not trivial and I don't want to invest time if it's not going to be merged

PR #1039, opened 2015-12-01
'group' field may be absent in a pacman package, which prevented a valid build

slact:


lib/fpm/package/tar.rb
```diff
@@ -49,7 +49,7 @@ def input(input_path)
   def output(output_path)
     output_check(output_path)
     # Unpack the tarball to the staging path
-    args = ["-cf", output_path, "-C", staging_path]
+    args = ["--owner=root", "--group=root", "-cf", output_path, "-C", staging_path]
```
jordansissel: What does this fix?

slact: I'm sorry, this line does not belong in this pull request. It has nothing to do with the 'group' field in pacman packages, and was in fact a hack i used to automatically change the owner of files for a .tar.gz "non-package"  in fpm.

hatt: These flags also depend very explicitly on `gnutar` and it is preferable to use `uid`/`gid` with a numeric ID.

slact: Yep. Like I said, I included this line in this pull request in error, It doesn't belong anywhere other than my fork.

jordansissel:
@slact if you tell me what commits you want included in this PR, I'm happy to merge those specifically. Alternately, you can do whatever youw ish to this PR (close and open a new one, rebase this one, whatever) to make this one have that 5eed079 commit

slact:
moved to https://github.com/jordansissel/fpm/pull/1113

PR #1040, opened 2015-12-08
Follow debian systemd packaging guide

nervo:
According to debian systemd packaging guide (https://wiki.debian.org/Teams/pkg-systemd/Packaging#systemd_unit_files_naming_and_installation), distributed systemd unit files must be located in "/lib/systemd/system", and not "/etc/systemd/system"

PR #1041, opened 2015-12-08
Follow debian systemd packaging guide #2

nervo:
Oh god, sorry, i did'nt see that systemd unit files are handled two times !

(See: https://github.com/jordansissel/fpm/pull/1040)

If i don't replace this path too, unit files are presents 2 times in the debian package, first in /lib/systemd, second in /etc/systemd

Btw, i suspect another bug, that's a strange behaviour, don't you think ?


jordansissel:
That is weird behavior... I agree ;P

jordansissel:
It'd be sweet if we had more tests to verify expected behavior on some of these newer features - no pressure though :)

PR #1044, opened 2015-12-11
Fix parsing of --description

sengel:
Current handling of option --description does not accept a description 'value' anymore as former commit 52ac00b5b3f7bc56c13ea0a57d8af609f1075b18 seems to be incomplete.


jordansissel:
Hah, oops. Good catch! Code LGTM

PR #1047, opened 2015-12-13
Typo

nervo:
"...will be labeled asconfiguration files..." -> "...will be labeled as configuration files..."


jordansissel:
LGTM! Thank you for making fpm better :)

PR #1049, opened 2015-12-15
deb: explicitly chmod triggers to 0644 for lintian

fasterthanlime:
:star2: Should fix https://github.com/itchio/itch/issues/93 :bug:


fasterthanlime:
Can confirm this did fix https://github.com/itchio/itch/issues/93 for us!

fasterthanlime:
Still a good fix :+1:

fasterthanlime:
Would still love this merged as I need to maintain our fork as long as it's not :(

jordansissel:
LGTM

PR #1050, opened 2015-12-15
Added bsdtar installation instructions in README

sgnn7:
This is needed or many tests fail on Ubuntu since bsdtar isn't there by default. 9 tests fail on Ubuntu 15.10 even after that but much less with bsdtar installed.

PS: Cleaned up some whitespace too

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,10 +183,17 @@ Most importantly, you should see the following output from the bundle command
 when it lists the fpm gem:
 
     ...
-    Using json (1.8.1) 
+    Using json (1.8.1)
     Using fpm (0.4.42) from source at .
-    Using hitimes (1.2.1) 
+    Using hitimes (1.2.1)
     ...
+
+If your system doesn't have `bsdtar` by default, make sure to install it or some
+tests will fail:
+
+    apt-get install bsdtar
+    
+    yum install bsdtar
 
 Next, run make in root of the fpm repo.  If there are any problems (such as
 missing dependencies) you should receive an error


hatt:
This is required for Arch Linux packaging, might be worth noting *why* some of our dependencies exist so people can opt in/out as required.

README.md
```diff
@@ -183,11 +183,18 @@ Most importantly, you should see the following output from the bundle command
 when it lists the fpm gem:
 
     ...
-    Using json (1.8.1) 
+    Using json (1.8.1)
```
jordansissel: trailing whitespace was stripped here, probably by some editor misconfiguration? I'll ignore it for this PR since it does'nt break anything. <3

jordansissel:
LGTM

jordansissel:
@hatt +1 your proposal. For tar, I'd love to stop invoking `tar` as a subprocess and just write the tar files directly with something like the `archive-tar-minitar` rubygem -- but that'll take a bit of work to get there...

hatt:
I notice we need the xz gem for the FreeBSD stuff, the bsdtar binary for MTREE Arch packages, and gnutar binary for most other Arch packages (since bsdtar breaks on setting uid/gid properly so we can't just use that.) An actual all-in-one compression gem would be great. I looked at libarchive but the rubygem for it seems out of date, and I'm not sure it handles the ownership flags properly anyway, let alone compression. See a bit more info [here](https://github.com/jordansissel/fpm/issues/231#issuecomment-199603480).

Issue #381, opened 2013-03-13
avoid defaulting epoch=1 for rpms


lexinator:
[sent this to fpm-users, but putting it here as well]

fpm is defaulting to epoch=1 for RPMs.  I found this when upgrading from 0.4.6 to 0.4.29(with some hair pulling to debug).  I am working around by using --epoch ''.

I'd like to point out this seems like the wrong default if you read about epoch use, http://www.rpm.org/max-rpm-snapshot/s1-rpm-depend-manual-dependencies.html.  If I understand correctly they recommend to use epoch only if the package version isn't sort order friendly.  (and tell the package provider to read http://semver.org =)

In the next major revision of fpm would it be possible to revert epoch to empty?

thanks
lex

brendanlong:
Is there a way to make the warning go away?

jordansissel:
@brendanlong we could add a flag for it, maybe?

brendanlong:
That would work. It's just annoying to be "warned" that nothing is wrong every time, especially if you integrate fpm into scripts.

jordansissel:
Agreed. Maybe since the warning was added 2+ years ago as part of a way to help notify users of the behavior change --- maybe it'd be fine to remove it?

I'm somewhat in favor now of removing the warning.

lexinator:
+1 on removing the warning.

josephfrazier:
+1 on removing the warning, or providing a flag to silence it

PR #1053, opened 2015-12-24
Don't warn when RPM Epoch isn't set

josephfrazier:
See https://github.com/jordansissel/fpm/issues/381#issuecomment-162644413


jordansissel:
LGTM

Issue #1055, opened 2015-12-29
virtualenv error


hejeke:
When package a virtualenv, it will cause a error of File does not have the world_readable function

jordansissel:
@hejeke what version of ruby are you using? I see it in Ruby 2.2 - http://ruby-doc.org/core-2.2.2/File.html#method-c-world_readable-3F

PR #1056, opened 2015-12-29
File class do not have a world_readable function

hejeke:
Fix [issue#1055](https://github.com/jordansissel/fpm/issues/1055)


jordansissel:
It does, but I think it was added in Ruby 2.0 or beyond.

I'm ok merging your proposed fix.

jordansissel:
Thank you! :)

PR #1061, opened 2016-01-26
deb: Create directories as 0755 as per Debian policy

rimmington:
This removes some lintian non-standard-dir-perm warnings and moves output
deb packages slightly closer to Debian policy compliance.


jordansissel:
specs pass. code LGTM.

PR #1072, opened 2016-02-29
minor README fix...

dcaba:
yum/rpm based distros also require "make" package


jordansissel:
Hmm. I haven't hit this problem, but I trust you. <3

PR #1084, opened 2016-04-01
Fix for --help output

svmaris:
The help output for --prefix states:

> A path to prefix files with when building the target package. This may be necessary for all input packages. For example, the 'gem' type will prefix with your gem directory automatically.

This should probably be:

> A path to prefix files with when building the target package. This may **not** be necessary for all input packages. For example, the 'gem' type will prefix with your gem directory automatically.


jordansissel:
LGTM

PR #1086, opened 2016-04-04
Remove artifact of old code in README

meshy:
This removes a slightly confusing bit of information from the readme that no longer looks relevant.

I think this is a result of the work done in 7ced849. The package name "awesome" was removed from the example `fpm` command in the readme, but the leading paragraph was not updated.


meshy:
Thank you for this exciting program, by the way `:)`

jordansissel:
Thanks! <3

PR #1087, opened 2016-04-07
OS X support (travis) + container-based infrastructure (travis) + fixes to pass all travis tests

ptomulik:
What changed:

- added new option ``--deb-upstream-changelog`` to allow upstream changelogs (*changelog.gz*) in addition to debian-specific changelog (*changelog.Debian.gz*),
- if only one changelog is provided, then it is packaged as *changelog.gz*, not *changelog.Debian.gz*, see https://www.debian.org/doc/debian-policy/ch-docs.html#s-changelogs,
- fixed code that should include */etc/* into ``allconfigs``, even if none of ``deb-init-list``, ``deb-default-list``, ``deb-upstart-list`` is defined,
- modified ``spec/fixtures/deb/stagging/etc/init.d/test`` such that it contains LSB headers, sheabang, is free of syntax errors and has executable attribute set.
- adjusted ``spec/fpm/package/deb_spec.rb`` to the above changes.


ptomulik:
All lights green on travis-ci. Finally managed to fix some tests. This PR also includes changes proposed in #1088 (switching to container based infrastructure on travis). I leave it for review, IMHO it's ready.

ptomulik:
In the meantime - enabled OS X in travis-ci. Tests are performed on Linux and OS X. Stats:
- Linux - 255 tests are active(same number as in master)
- OS X - 256 tests are active

All the active tests pass.

.travis.yml
```diff
@@ -1,15 +1,32 @@
...
 before_install:
-  - sudo apt-get update -qq
-  - sudo apt-get install -qq rpm
-  - sudo apt-get install -qq lintian
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update           ; fi
+    # dpkg has also gnu-tar and xz as dependencies, these packages are used
+    # by pacman.rb, for example.
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install dpkg ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then which dpkg-deb || 'dpkg-deb not found'; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then which lintian || echo 'lintian not found'; fi
```
hatt: This will never pass, Lintian can't be installed on OSX 😞

lib/fpm/package/pacman.rb
```diff
@@ -96,7 +97,7 @@ def dependencies
   def input(pacman_pkg_path)
     control = {}
     # Unpack the control tarball
-    safesystem("tar", "-C", staging_path, "-xf", pacman_pkg_path)
+    safesystem(FPM::Util.tar_cmd, "-C", staging_path, "-xf", pacman_pkg_path)
```
hatt: No need to namespace here, just `tar_cmd` is fine although that's more of a style thing, same at line 268.

.travis.yml
ptomulik: I used to never say "never", I believe there will be a day when we'll even see lintian under OS X. So, got prepared for that day :)

lib/fpm/package/pacman.rb
ptomulik: Ok, thanks. I'll correct it.

spec/fpm/package/deb_spec.rb
```diff
@@ -362,13 +366,13 @@ def dpkg_field(field)
...
-        "init.d-script-missing-lsb-section",
-        "non-standard-file-permissions-for-etc-init.d-script"
+        "script-in-etc-init.d-not-registered-via-update-rc.d"
       ]
 
       it "should return no errors" do
-        lintian_output = `lintian #{target} --suppress-tags #{lintian_errors_to_ignore.join(",")}`
+        lintian_output = `lintian #{target} --suppress-tags '#{lintian_errors_to_ignore.join(",")}'`
         expect($CHILD_STATUS).to eq(0), lintian_output
+        expect($CHILD_STDERR).to be nil
```
jordansissel: what is $CHILD_STDERR?

ptomulik: It's stderr (standard error) output from lintian. This ensures, that lintian runs cleanly without warnings. Previously, only errors were discovered by the test.

ptomulik: Ah, maybe too restrictive, I didn't noticed the test title "should return no error". Should I revert it or it's acceptable to leave the test in this stricter version?

jordansissel: I didn't articulate my question very well, sorry, ;)

$CHILD_STDERR is not a global variable I am aware of. The way lintian is invoked above doesn't capture stderr. Global variables are nil by default, so I think this new expect only passes by accident.

```
# from irb, showing an unknown global $WHOA being nil by default
>> $WHOA
=> nil
```

jordansissel:
:+1:  for having tests passing on travis <3

My only concern is the $CHILD_STDERR test which I don't think actually tests anything (Since $CHILD_STDERR is probablynot a global that is set, it defaults to nil, so the test happens to pass)

spec/fpm/package/deb_spec.rb
ptomulik: Ah, see! Removed. Thanks.

ptomulik:
Yes, you're right. I've removed that line. PR ready for further processing.

ptomulik:
Thanks, can we expect new gem release?

jordansissel:
Yep! Will polish that up hopefully tonight. Thank you for helping so much :)

Issue #1075, opened 2016-03-08
Ancient rubies no longer supported, maybe...


jordansissel:
In master, there are several pieces of code that prevent master fpm from running under Ruby 1.8 and 1.9.

This ticket is to track those issues as well as to collect ideas and energy from users to continue supporting Ruby 1.8 and 1.9.

----

Known issues

* FreeBSD package support uses refinements (Requires minimum Ruby 2.0.0)

PR #1088, opened 2016-04-07
Container based travis

ptomulik:
1. Use [corefines](https://github.com/jirutka/corefines) to provide backward compatiblity with ruby 1.9.3 (broken due to refinements used in package/freebsd.rb).
2. Switch to [container-based infrastructure](https://docs.travis-ci.com/user/workers/container-based-infrastructure/) on travis-ci, because it's so cool :)


hatt:
How is this to work with OSX testing, or are you proposing Linux only? There is also #1075 to maybe not care about old rubies anymore. Allowing failure on <2.0 but still testing could be useful though I think.

ptomulik:
Can't tell how it works with OSX testing. I'd propose to run tests for #1087 on OSX, if there is one who has access to OSX. #1087 enhances #1088 in the sense it adds some fixes to test suites and core code that make tests to pass (at least on Linux).

I believe, supporting 1.9.3 ATM doesn't cost much (it's just two lines of code), except there are other issues than #1075, that break compatibility with 1.9 and older rubies on other platforms.

hatt:
It looks like there is no way to test this on OSX, given Lintian has a few Debian-only dependencies (some cray cray Perl stuff going on) so installing it is pretty much impossible to do simply. I don't like the idea of limiting testing to a given platform considering some of our packaging is already very platform specific (osxpkg only works on OSX, rpm requires rpmbuild...). Using containers for testing means we can't auto-test all features, no matter how cool it is, and using a mix of containers and not just seems like a headache. Just my $0.02 though.

I'm down with pseudo-supporting 1.9.3 but no real opinion on corefines, and I don't see what the value is given we can just fix issues as they come up for 1.9.x. 1.8.7 support is a bit different since the issues are more indepth and frequent and I'm not sure how corefines would help with this.

ptomulik:
I can't see how container-based infrastructure limits testing? Could you be more specific? An example? 

I've take a look at: https://docs.travis-ci.com/user/multi-os#Example-Multi-OS-Build-Matrix, it looks like it's possible to run multi-os tests on travis-ci with Linux and OS X and we can use container based infrastructure on Linux.

I've also put simple [.travis.yml](https://github.com/ptomulik/testmd/blob/master/.travis.yml) to [one of my projects](https://github.com/ptomulik/testmd/) that verifies the concept. Looks like container based infra should not interfere with OS X builds on travis. [Here are travis runs](https://travis-ci.org/ptomulik/testmd/builds/121625992) for this project.

ptomulik:
Finally, decided to enable OS X testing on travis-ci in 17ffa31dcb2381717a7730a723f8d549fb47e472. See if you like it. #1087 contains same + fixes such that all the tests are green.

hatt:
Ah ok,  i didn't realise they had OSX so separately handled. That sounds good to me as long as all the dependencies are there and no tests get skipped then. TBH the FreeBSD stuff is doing quite a few weird things that don't follow the way other stuff does it in FPM, but I'm not sure if we should update other stuff to use more recent Ruby versions or if we should get the FreeBSD packager functioning the same as everything else. Mainly around archive usage (`xz` gem vs gnutar/bsdtar for example).

I might get around to updating the FPM Omnibus HEAD to use 2.latest if there is a place to host the packages maybe. Realistically, Ubuntu's next LTS is out in a few days and 14.04 is one of the last distributions I can think of still using 1.9.3 that is a viable desktop distro. I'm not sure how relevant supporting anything under 2.0 is anymore.

Thanks for all the testing and verification you've done on this, looks good to me!

jordansissel:
I am thinking we can just make FreeBSD support only loaded under ruby 2.x

ptomulik:
Closing this PR, as all these changes were already provided by #1087.

PR #1091, opened 2016-04-13
add FPM::Util.execmd and fix python_spec.rb

ptomulik:
Created mainly to fix the issue spotted [here](https://github.com/jordansissel/fpm/commit/42df3cccc0a6cb7fc92ab026d57db938e731fbe4#commitcomment-17079001).

### DIFF
--- a/lib/fpm/util.rb
+++ b/lib/fpm/util.rb
@@ -45,6 +45,144 @@ module FPM::Util
     return shell
   end
 
+  ############################################################################
+  # execmd([env,] cmd [,opts])
+  #
+  # Execute a command as a child process. The function allows to:
+  #
+  # - pass environment variables to child process,
+  # - communicate with stdin, stdout and stderr of the child process via pipes,
+  # - retrieve execution's status code.
+  #
+  # ---- EXAMPLE 1 (simple execution)
+  #
+  # if execmd(['which', 'python']) == 0
+  #   p "Python is installed"
+  # end
+  #
+  # ---- EXAMPLE 2 (custom environment variables)
+  #
+  # execmd({:PYTHONPATH=>'/home/me/foo'}, [ 'python', '-m', 'bar'])
+  #
+  # ---- EXAMPLE 3 (communicating via stdin, stdout and stderr)
+  #
+  # script = <<PYTHON
+  # import sys
+  # sys.stdout.write("normal output\n")
+  # sys.stdout.write("narning or error\n")
+  # PYTHON
+  # status = execmd('python') do |stdin,stdout,stderr|
+  #   stdin.write(script)
+  #   stdin.close
+  #   p "STDOUT: #{stdout.read}"
+  #   p "STDERR: #{stderr.read}"
+  # end
+  # p "STATUS: #{status}"
+  #
+  # ---- EXAMPLE 4 (additional options)
+  #
+  # execmd(['which', 'python'], :process=>true, :stdin=>false, :stderr=>false) do |process,stdout|
+  #  p = stdout.read.chomp
+  #  process.wait
+  #  if (x = process.exit_code) == 0
+  #    p "PYTHON: #{p}"
+  #  else
+  #    p "ERROR:  #{x}"
+  #  end
+  # end
+  #
+  #
+  # OPTIONS:
+  #
+  #   :process (default: false) -- pass process object as the first argument the to block,
+  #   :stdin   (default: true)  -- pass stdin object of the child process to the block for writting,
+  #   :stdout  (default: true)  -- pass stdout object of the child process to the block for reading,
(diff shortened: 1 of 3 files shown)


hatt:
I might be missing something but this looks a bit more like it's replacing what the safesystem wrapper is doing, maybe better to update/patch it for the extra functionality?

ptomulik:
I can't imagine how to provide special argument with the input string (for stdin) to safesystem, as its interface is
```
def safesystem(*args)
```
and all args are treated as command-line arguments (except the first, which is a command name). I think, ``execmd`` is quite more generic, so ``safesystem`` could internally call ``execmd`` and preserve its interface.

ptomulik:
See 7b2d139

spec/fpm/package/python_spec.rb
```diff
@@ -164,24 +199,17 @@ def python_usable?
...
-        # Windows
-        # (ptomulik) I'm not sure about this
-        django_bindir = %x{python -c 'from distutils.sysconfig import get_config_var; print(get_config_var("BINDIR"))'}.chomp
-      else
-        # Unix
-        #django_bindir = '/usr/local/bin'
-        django_bindir = %x{python -c 'from distutils.sysconfig import get_config_var; print(get_config_var("BINDIR"))'}.chomp
+      django_bindir = nil
+      # Determine, where 'easy_install' is going to install scripts
+      execmd('python', code_easy_install_default_script_dir) do |code,out,err|
```
jordansissel: As an alternative, you could include this python script in fpm similar to the way the python package support already does this --

[python code](https://github.com/jordansissel/fpm/tree/master/lib/fpm/package/pyfpm) and [how it's invoked in fpm](https://github.com/jordansissel/fpm/blob/master/lib/fpm/package/python.rb#L181-L182).

Since this is just for a test, the python code could live in a file in spec/fixtures/ somewhere (common in other similar ruby projects) and you can invoke `safesystem("python spec/fixtures/default_script_dir.py")` possibly capturing the output to a file like done [here](https://github.com/jordansissel/fpm/blob/master/lib/fpm/package/python.rb#L182)

Thoughts?

ptomulik: I think, it may be good to leave ``execmd`` as, in opposite to ``safesystem*`` ones, it provides all standard communication channels to the executed command (``stdin``, ``stdout``, ``stderr``, ``exit_status``). As you may se, I managed to implement both versions of ``safesystem*`` on top of ``execmd``. I'm ok with moving the python code to fixture file (this could push me to organize it little bit better), but the actuall script consists of two pieces -- [definition of a class](https://github.com/ptomulik/fpm/blob/fix/python_spec/spec/fpm/package/python_spec.rb#L15) and [a code that uses this class](https://github.com/ptomulik/fpm/blob/fix/python_spec/spec/fpm/package/python_spec.rb#L45). I think, the usage code (second part) is not worth giving it a fixture (it's a very small piece of code, and I don't think it's reusable), but hopefully it can be just passed to python interpreter via CLI.

ptomulik:
So, I made changes, according to [previous proposition](https://github.com/jordansissel/fpm/pull/1091#discussion_r59784408). The new `execmd` provides convenient and still powerful interface to ``ChildProcess``. It allows passing environment variables to child process and communicating it via pipes (``stdin``, ``stdout`` and ``stderr``) .

hatt:
Looks good to me!

spec/fpm/package/python_spec.rb
```diff
@@ -12,6 +12,18 @@ def python_usable?
     "'python' and/or 'easy_install' isn't in your PATH")
 end
 
+# Determine default value of a given easy_install's option
+def easy_install_default(option)
+  result = nil
+  execmd({:PYTHONPATH=>"#{example_dir}"}, 'python', :stderr=>false) do |stdin,stdout|
```
ptomulik: Using "example_dir" here may look little-bit out of context, but it should be fine as long as the function is used within the ``describe FPM:Package::Python...`` block.

lib/fpm/util.rb
```diff
@@ -45,42 +45,158 @@ def default_shell
...
+  #
+  #
+  # OPTIONS:
+  #
+  #   :process (default: false) -- pass process object as the first argument the to block,
+  #   :stdin   (default: true)  -- pass stdin object of the child process to the block for writting,
+  #   :stdout  (default: true)  -- pass stdout object of the child process to the block for reading,
+  #   :stderr  (default: true)  -- pass stderr object of the child process to the block for reading,
+  #
+  def execmd(*args)
```
ptomulik: I don't like this ``execmd`` name, but didn't invented anything better. Any suggestions?

jordansissel:
Specs are passing with this PR, so I will merge.

PR #1093, opened 2016-04-15
fixes for FPM::Package#to_s and relatives

ptomulik:
The main issue that this patch tries to address it that ``freebsd`` packager ignored ``--package`` CLI option. Working on it I also decided to add some customization points to ``FPM::Package#to_s`` such that subclasses can easily overload certain bits of its behavior.

### DIFF
--- a/lib/fpm/package.rb
+++ b/lib/fpm/package.rb
@@ -339,15 +339,14 @@ class FPM::Package
   def to_s_arch; architecture.to_s; end
   def to_s_name; name.to_s;         end
   def to_s_fullversion
-    ver = to_s_version
-    ver += "-#{iteration}" if iteration
-    return ver
+    return "#{version}-#{iteration}" if iteration
+    return "#{version}"
   end
   def to_s_version;   version.to_s;   end
   def to_s_iteration; iteration.to_s; end
   def to_s_epoch;     epoch.to_s;     end
   def to_s_type;      type.to_s;      end
-  def to_s_extension; to_s_type;      end
+  def to_s_extension; type.to_s;      end
   #######################################
 
   def to_s(fmt=nil)
--- a/lib/fpm/package.rb
+++ b/lib/fpm/package.rb
@@ -336,17 +336,14 @@ class FPM::Package
   # easily override particular substitut-
   # ions performed by to_s for subclasses
   #######################################
-  def to_s_arch; architecture.to_s; end
-  def to_s_name; name.to_s;         end
-  def to_s_fullversion
-    return "#{version}-#{iteration}" if iteration
-    return "#{version}"
-  end
-  def to_s_version;   version.to_s;   end
-  def to_s_iteration; iteration.to_s; end
-  def to_s_epoch;     epoch.to_s;     end
-  def to_s_type;      type.to_s;      end
-  def to_s_extension; type.to_s;      end
+  def to_s_arch;        architecture.to_s;  end
+  def to_s_name;        name.to_s;          end
+  def to_s_fullversion; iteration ? "#{version}-#{iteration}" : "#{version}"; end
+  def to_s_version;     version.to_s;       end
+  def to_s_iteration;   iteration.to_s;     end
+  def to_s_epoch;       epoch.to_s;         end
+  def to_s_type;        type.to_s;          end
+  def to_s_extension;   type.to_s;          end
   #######################################
 
   def to_s(fmt=nil)
--- a/lib/fpm/package.rb
+++ b/lib/fpm/package.rb
@@ -331,17 +331,35 @@ class FPM::Package
     return erb
   end # def template
 
-  def to_s(fmt="NAME.TYPE")
-    fmt = "NAME.TYPE" if fmt.nil?
-    fullversion = version.to_s
-    fullversion += "-#{iteration}" if iteration
-    return fmt.gsub("ARCH", architecture.to_s) \
-      .gsub("NAME", name.to_s) \
-      .gsub("FULLVERSION", fullversion) \
-      .gsub("VERSION", version.to_s) \
-      .gsub("ITERATION", iteration.to_s) \
-      .gsub("EPOCH", epoch.to_s) \
-      .gsub("TYPE", type.to_s)
+  #######################################
+  # The following methods are provided to
+  # easily override particular substitut-
+  # ions performed by to_s for subclasses
+  #######################################
+  def to_s_arch; architecture.to_s; end
+  def to_s_name; name.to_s;         end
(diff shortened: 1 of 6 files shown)


lib/fpm/package.rb
```diff
@@ -331,17 +331,35 @@ def template(path)
...
+  #######################################
+  # The following methods are provided to
+  # easily override particular substitut-
+  # ions performed by to_s for subclasses
+  #######################################
+  def to_s_arch; architecture.to_s; end
+  def to_s_name; name.to_s;         end
+  def to_s_fullversion
+    ver = to_s_version
+    ver += "-#{iteration}" if iteration
```
jordansissel: If I'm understanding your proposal, shouldn't this use `"-#{to_s_iteration}"` ?

lib/fpm/package/freebsd.rb
```diff
@@ -128,8 +128,17 @@ def add_path(tar, tar_path, path)
     end
   end # def add_path
 
+  def to_s_extension; "txz"; end
+
+  def to_s_fullversion()
+    # iteration (PORTREVISION on FreeBSD) shall be appended only(?) if non-zero.
+    # https://www.freebsd.org/doc/en/books/porters-handbook/makefile-naming.html
+    return "#{to_s_version}_#{iteration}" if iteration and (iteration.to_i > 0)
```
jordansissel: should this also use `#{to_s_iteration}` ?

lib/fpm/package.rb
ptomulik: Maybe. Or, should we leave it as is and use ``type.to_s`` in the ``to_s_extension`` below? Maybe this second approach would be better, as it does not introduce "unexpected" interdependences between ``to_s_*`` functions?

lib/fpm/package/freebsd.rb
ptomulik: Or replace ``to_s_version`` with just ``version`` (to avoid interdependencies between ``to_s_*`` methods).

ptomulik:
Corrected ``to_s_*`` stuff.

hatt:
Looks good!

lib/fpm/package/deb.rb
```diff
@@ -905,7 +905,7 @@ def mkdir_p(dir)
   def to_s(format=nil)
     # Default format if nil
     # git_1.7.9.3-1_amd64.deb
-    return super("NAME_FULLVERSION_ARCH.TYPE") if format.nil?
+    return super("NAME_FULLVERSION_ARCH.EXTENSION") if format.nil?
```
ptomulik: I wonder why ruby people don't use the [ternary if](http://docs.ruby-lang.org/en/2.1.0/syntax/control_expressions_rdoc.html#label-Ternary+if) control expression? This line and line below could be simply reduced to:
```
super(format.nil? "NAME_FULLVERSION_ARCH.EXTENSION" : format)
```
Similarly in other modules.

ptomulik:
So, merge, maybe? :)

lib/fpm/package/deb.rb
hatt: I think for the most part it's good but some lines have long gsub conditions which could make it hard to read. Re merging, only Jordan can do that.

lib/fpm/package/rpm.rb
```diff
@@ -562,8 +562,8 @@ def epoch
 
   def to_s(format=nil)
     if format.nil?
-      return super("NAME-VERSION-ITERATION.DIST.ARCH.TYPE").gsub('DIST', attributes[:rpm_dist]) if attributes[:rpm_dist]
-      return super("NAME-VERSION-ITERATION.ARCH.TYPE")
+      return super("NAME-VERSION-ITERATION.DIST.ARCH.EXTENSION").gsub('DIST', attributes[:rpm_dist]) if attributes[:rpm_dist]
```
ptomulik: This is something that looks strange to me. Why we need to perform substitutions on a fixed (ad-hoc) string? Merely, it's same as
```
super("NAME-VERSION-ITERATION.#{attributes[:rpm_dist]}.ARCH.EXTENSION") if attributes[:rpm_dist]
```
except the current substitution is made on a string already processed by ``super`` (which doesn't really matter ATM, as ``super`` does not introduce any ``"DIST"`` substrings). I'd suggest to delegate this gsub to ``FPM::Package#to_s`` for readability.

ptomulik: Oh, no ``rpm_dist`` is RPM specific. Can't be delegated to ``FPM::Package``.

ptomulik: Anyway, if ``format`` argument is provided and is not ``nil``  and f already contains ``"DIST"``, the ``"DIST"`` will not be substituted.

lib/fpm/package/deb.rb
jordansissel: > I wonder why ruby people don't use the ternary if control expression?

I'm not what I would consider a "ruby" person (truthfully, I'm not quite sure what that means, so maybe I am! hehe). I have a habit of avoiding ternary operator, but I'm not sure why I do. ;P

jordansissel:
LGTM

lib/fpm/package/deb.rb
ptomulik: Don't mistake "ruby" with "rude".  :) I needed two iterations to discover why there are two ``return`` instructions one after the another. That was crazy, as I'm not used to read whole lines of code, only first one or two words :)

ptomulik:
Thanks.

lib/fpm/package/deb.rb
jordansissel: haha, I agree that this is more than a bit awkward. It certainly looks like one of those "code smell" type of problems where it just feels weird. We could rewrite it like this, using default argument:

```
def to_s(format="NAME_FULLVERSION_ARCH.EXTENSION")
  super(format)
end
```

And then I wonder why this method exists just to have a default value in a weird way that only really calls super(...). We could probably refactor this to have each package type be able to describe it's own default format, then not need to define `to_s` for each package type.

A refactor for another day, perhaps ;P

(If you dive into doing this, I support it, happy to review whatever)

ptomulik: Another thought -- ``to_s`` is usually meant to represent object in a string form for displaying purposes. Here it looks like it has a purpose of generating output file name. Is it intentional? Shouldn't these two things be separated?

jordansissel: Indeed! Good observation. In hindsight, maybe .to_s primarily for
formatting file names wasn't a great move.

PR #1094, opened 2016-04-22
Fix gem error

mildred:
Without `--no-user-install`, the gem command fails with:

    ERROR:  User --install-dir or --user-install but not both

Apparently, on some circumstances, gem can default to a user install and prevent it from using `--install-dir`.

I am using gem 2.5.1 on ruby 2.3.0p0 (2015-12-25 revision 53290)


jordansissel:
LGTM. Failing test is unrelated to this PR.

PR #1099, opened 2016-04-27
add support for scoped npm packages

danmartinez101:
Currently building an rpm from a scoped npm package explodes because of the `/` between the scope and name as well as the `@` in the scope.

If this is a PR you are interested in accepting, I can put some more time and effort into cleaning this up.

### DIFF
--- a/lib/fpm/package/npm.rb
+++ b/lib/fpm/package/npm.rb
@@ -62,7 +62,7 @@ class FPM::Package::NPM < FPM::Package
     npm_ls = JSON.parse(npm_ls_out)
     name, info = npm_ls["dependencies"].first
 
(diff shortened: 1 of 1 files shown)


hatt:
Some tests would probably be nice but looks good to me. Probably doesn't really need any cleaning up unless we want to strictly order the name components.

jordansissel:
I support your proposal, but I don't know anything about scoped npm packages at this time so I'm not sure how to test or provide feedback. Can you advise?

danmartinez101:
we ended up just using the dir source for this instead

cwegener:
I tried this patch on Ruby 3.0 and FPM 1.15.1 .. but it now no longer seems to do the trick. Even with this patch, trying to install scoped npm packages now makes fpm think that it's a local directory.

PR #1101, opened 2016-05-02
Add fpm namespace require for rake task

evan2645:
The rake task class needs to require the fpm namespace in order
to be included elsewhere.


jordansissel:
Thank you :)

PR #1102, opened 2016-05-03
Rake task should not exit if successful

evan2645:
Exiting after fpm success means that further rake tasks fail to execute,
and Rake simply exits. Instead, only abort if the fpm exits non-zero


jordansissel:
Thank you :)

PR #1105, opened 2016-05-06
Erase release candidate spec from dependencies when converting from Gem to RPM

jhull123:
This week I encountered an issue when building RPMs for Nokogiri and its dependencies. The issue is that Nokogiri lists a release candidate version of mini_portile2 as a dependency.

When I build this RC version of mini_portile2 the resulting RPM's version will not include the RC suffix (2.0.0 instead of 2.0.0.rc2). When I build Nokogirl, the RPM dependencies will include mini_portile2 v2.0.0.rc2. Thus, Nokogiri will fail to install (via yum) since the mini_portile2 dependency cannot be satisfied.

I made small tweak to FPM to erase release candidate suffixes from RPM dependencies in order to prevent this situation.


lib/fpm/package/rpm.rb
```diff
@@ -301,6 +301,11 @@ def converted_from(origin)
         end
       end.flatten
     end
+   
+    # Erase troublesome release candidate declarations.
+    self.dependencies = self.dependencies.collect do |dep|
+      dep.gsub(/.rc\d*/, '')
```
jordansissel: prerelease gems can have any suffix, not just `rcN`... not sure how to solve this.

jhull123: Yeah. I figured release candidate was the worst offender and was the one the tripped me up in particular.

jordansissel:

jhull123:
I did a bad job of rebasing. 😢 

I am going to start over and submit a new PR.

Sorry for the confusion.

jordansissel:
@jhull123 oops. Let me know if you need any help with this. Git is a strange beast, and we can often only survive with teamwork ;P

jordansissel:
I think you should be able to automate this by comparing your changes against the `master-pre-git-maintenance-june-2016` branch. Lemme see what I can come up with...

jordansissel:
Ahh, I think you deleted your remote branch. Sorry for the problems here. Let me know if I can help at all. If you need to compare against the old "master" branch (before I did some git history maintenance), compare it with branch `master-pre-git-maintenance-june-2016`

PR #1108, opened 2016-05-10
workaround rubygems issue #1608 and remove dependency of corefines

### LINKED ISSUE #1107: freebsd packager broken due to bug in rubygems
For reference see: https://github.com/rubygems/rubygems/issues/1608

the freebds.rb uses ``TarWriter`` which fails for absolute file names of exact length 101 characters. I'm working on freebsd support for [vagrant-installers](https://github.com/mitchellh/vagrant-installers) and this blocks my progress. There is already a patch for [rubygems/#1608](https://github.com/rubygems/rubygems/issues/1608) in https://github.com/rubygems/rubygems/pull/1612 but I don't know how long it takes to merge it and whether it will be back-ported to older versions of rubygems.

My question is - is it possible to workaround it, for example by not using the ``Gem::Package::TarWritter`` or by not using absolute names in ``freebsd.rb``? The relevant function is [this](https://github.com/jordansissel/fpm/blob/master/lib/fpm/package/freebsd.rb#L112)

ptomulik:
Fixes #1107. I had to provide new class ``FPM::Util::TarWriter`` to patch problems with ``Gem::Package::TarWriter``. Refinemends didn't seem to be useful in this case so I decided to get rid of the only refinement used in ``freebsd.rb``, as well as to remove the dependency of ``corefines``.

Tested on FreeBSD 10.3 (used to succesfully build [vagrant-installers](https://github.com/mitchellh/vagrant-installers)).

BTW. Is it possible to have fpm-1.5.1 with this PR included soon?

### DIFF
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -21,10 +21,10 @@ Gem::Specification.new do |spec|
   # For parsing JSON (required for some Python support, etc)
   # http://flori.github.com/json/doc/index.html
   spec.add_dependency("json", ">= 1.7.7") # license: Ruby License
-  
+
   # For logging
   # https://github.com/jordansissel/ruby-cabin
-  spec.add_dependency("cabin", ">= 0.6.0") # license: Apache 2 
+  spec.add_dependency("cabin", ">= 0.6.0") # license: Apache 2
 
   # For backports to older rubies
   # https://github.com/marcandre/backports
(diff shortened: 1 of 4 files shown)


jordansissel:
Thank you! I will review this tonight and see about getting a release out.

jordansissel:
Code looks OK to me. Merging.

jordansissel:
For future todo items, it'd be nice if there were a test verifying that tar files with paths longer than 101 chars was supported.

ptomulik:
It looks like Gem::Rubygems::TarWritter uses tar format known as "ustar". It's quite cumbersome -- max path length is 256 characters, but it's not always achievable. Names of length <=100 are handled without problems. Names of length > 100 are split at a separator "/" into prefix (must be <= 155) and suffix (must be <= 100). Some paths can't be split so simply, so the other limiting factor is what the path looks like. For example a path "/aaa...a" of length 101 chars (including the leading "/") can't be handled by ustar (verified with GNU tar 1.28).

So in this case, the max length is implied by the ustar specification. It's not a matter of fpm, but rather the tar backend. The one used by freebsd.rb is kinda limited, I wonder what was reason for choosing the ``TarWritter``.

Concluding, instead of providing a test, maybe it could be better to choose appropriate tar backend knowingly.

PR #1109, opened 2016-05-11
Fix fpm to work with recent versions of pip

liger1978:
`--no-install` option has been removed from recent versions of pip.  Instead `pip download` should be used.

### DIFF
--- a/lib/fpm/package/python.rb
+++ b/lib/fpm/package/python.rb
@@ -131,7 +131,7 @@ class FPM::Package::Python < FPM::Package
                  "--build-directory", target, want_pkg)
     else
       logger.debug("using pip", :pip => attributes[:python_pip])
-      safesystem(attributes[:python_pip], "install", "--no-deps", "--no-install", "--no-use-wheel", "-i", attributes[:python_pypi], "-U", "--build", target, want_pkg)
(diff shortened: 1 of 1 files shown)


jordansissel:
Can you make this behavior depend on the version of pip that is available?

liger1978:
@jordansissel I don't know. I may try. There seem to be three versions of the syntax according to the [changelog](https://github.com/pypa/pip/blob/develop/CHANGES.txt):

  - < 7.0.0: `pip install --no-install` (only versions fpm currently works with)
  - >= 7.0.0: `pip install --download`
  - >= 8.0.0: `pip download`

Given that pip can easily update itself (`pip install pip --upgrade`), is it too much to ask that users run a recent version?

jordansissel:
@liger1978 given my experience as a sysadmin, it is often not possible to upgrade things like pip. How about this, can you add a comment like `# TODO: Support older versions of pip` so some future contributor can know that we'd like to support more than just pip 8?

liger1978:
I guess I assumed other sysadmins use dedicated build VMs as I do, but fair enough if this assumption is unwarranted!  Of course I will do as you ask.  Closing this pull request

PR #1113, opened 2016-05-14
'group' field may be absent in a pacman package, which prevented a valid build

slact:
Fixed Pull Request https://github.com/jordansissel/fpm/pull/1039


lib/fpm/package/pacman.rb
```diff
@@ -164,6 +164,10 @@ def input(pacman_pkg_path)
     end
 
     self.dependencies = control["depend"] || self.dependencies
+    
+    if attributes[:no_auto_depends?]
```
jordansissel: Was this an intended inclusion in this PR?

jordansissel: I think I understand why this was added. LGTM.

PR #1115, opened 2016-05-17
Drop duplicate .conf suffix

### LINKED ISSUE #861: --deb-upstart always adds .conf suffix
In  #479, romabysen suggested that the --deb-upstart option should work like the following:

--deb-upstart something.conf -> /etc/init/something.conf

This is still slightly broken: 

--deb-upstart something.conf -> /etc/init/something.conf.conf 

It adds the .conf suffix even if the file ends with .conf. 

Cheers,
Andy

josegonzalez:
Closes #861


jordansissel:
:heart:

PR #1120, opened 2016-05-26
Add --python-setup-py-arguments flag

cure:
This flag can be used to pass arbitrary arguments to setup.py.

This is helpful to package python modules that have a poorly written
setup.py that requires special arguments.

For example, this allows building ruamel.yaml which requires 

  --single-version-externally-managed

to be given to setup.py


jordansissel:

josephholsten:
fixed with rebase in #1376. Thanks!

PR #1121, opened 2016-05-30
example of usage for --rpm-attr

mkopta:


jordansissel:
Lovely. Thank you! :)

PR #1122, opened 2016-05-31
last_comment is depricated replaced with last_description

LStuker:
last_comment is deprecated with Rake 11.1.0 and will be removed with Rake 12.

more info read:
https://github.com/ruby/rake/issues/116


jordansissel:
Thanks!

PR #1126, opened 2016-06-01
Change the way the package relationship attributes are evaluated

TomyLobo:
The change to the recommends line fixes --deb-recommends for me.

The rest of the changes are untested.


jordansissel:
I agree with this change, but I think the commit message confuses me - it feels like this really fixes a bug (empty lists). "Change the way" is a bit confusing. I'll merge this by hand and update the commit message. <3

jordansissel:
Merged as 0b104c260501cc7f09e85e8418fa2fdfb7a7e039 into master. Thank you!

TomyLobo:
I didn't fully know what i was doing there tbh, i just made everything look the way it worked for the "recommends" line :)

jordansissel:
@TomyLobo no worries! Teamwork solves all :)

PR #1131, opened 2016-06-08
Feature virtualenv from requirements

nicobrevin:
I have a use case where I wanted to package up ansible with some extra dependencies, so using just a python package as an index isn't sufficient.

This adds support for building from requirements.txt, instead of an package name.  Plus I added some specs, as best I could, to the existing package code - there are some quirks in there that I might propose changing with a second PR....

You can test it by creating a requirements.txt anywhere and using that as the input, like so:

    echo 'paramiko' > some-package/requirements.txt
    fpm -s virtualenv -t deb requirements.txt # creates virtualenv-some-package_1.0_amd64.deb


jordansissel:
I like the idea! I wonder if --requirements is necessary, though. Could we have fpm observe that the input is a file named 'requirements.txt' and go into that 'requirements.txt' mode, instead of needing the `--requirements` flag?

jordansissel:
I'm open to merging as-is and then moving to remove the `--requirements` flag later. Thoughts?

jordansissel:

jordansissel:
Merged as 66e59e79e8e8965adde57869343899ae428876ed, d0eaf779211c9843ad5226ea6f6deb401cf1a484.

Made my change to make `--requirements` automatically detected (if the file given is a requirements.txt path) here: 7a7d057109446c56d956a3fdaebe500a06968694

jordansissel:
Thank you for making fpm more awesome!

nicobrevin:
Sorry, missed all your comments, looks like you worked it our for yourself.

Yes, the --requirements flag be optional where there might be ambiguity.

Thanks for merging!

PR #1132, opened 2016-06-10
Lazily require xz at the last possible moment to avoid dependency on liblzma.so.5

### LINKED ISSUE #1090: Latest fpm 1.5.0 fails to install on CentOS6: "ruby-xz requires Ruby version >= 1.9.3"
Maybe there's an easy workaround? I tried first installing an older version of ruby-xz  `gem install ruby-xz -v 0.1.1` but that didn't help.
```
$ docker run -ti centos:centos6 bash
# yum groupinstall -y "Development tools"
# yum install -y make gcc gawk ruby-devel rubygems
# gem install fpm
ERROR:  Error installing fpm:
	ruby-xz requires Ruby version >= 1.9.3.
```

To workaround, use older version of fpm on CentOS6:
```
# gem install fpm -v 1.4.0
Successfully installed fpm-1.4.0
1 gem installed
```

ketan:
Fixes #1090

### DIFF
--- a/lib/fpm/package/freebsd.rb
+++ b/lib/fpm/package/freebsd.rb
@@ -3,7 +3,6 @@ require "fpm/package"
 require "fpm/util"
 require "digest"
 require "fileutils"
-require "xz"
 
 class FPM::Package::FreeBSD < FPM::Package
   SCRIPT_MAP = {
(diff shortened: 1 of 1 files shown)


ketan:
Can we get a 1.6.1 out once this is merged, please?

jordansissel:
Sounds good to me :)

jordansissel:
fpm 1.6.1 published.

ketan:
Many thanks foe this awesome gem :)

PR #1134, opened 2016-06-15
Fix for ignoring --rpm-tag option on command line

specnazzz:
The `--rpm-tag` option was ignored when used on command line.


jordansissel:
Thank you :)

PR #1139, opened 2016-06-21
fix --workdir not having an effect on the build and staging dirs, fixes #1124

### LINKED ISSUE #1124: --workdir option not taken into account
We creates big data packages (~3-6Gb) and we tried tu use `--workspace` option to put working directory elsewhere since `/` is filled up with our data.
But this don't seem to work: it always uses `/tmp`.
Interrupting FPM with _^Z_ and looking in the specified directory will show an empty one while `/tmp/package-*` are created. The option is read from command line since the message `Setting workdir {:workdir=>"/home/pp//tmp", :level=>:info}` is printed.

luto:
This PR uses `Stud::Temporary.directory` instead of `::Dir.mktmpdir` to generate temporary directories for the "build"- and "staging"-dirs. This means that both directories now respect the `--workdir`-argument. Further details on this issue can be found in #1124.

In case there are any questions, please feel free to ask them. If you want anything changed about this PR, please just let me know and I'll get it done. Thank you.

### DIFF
--- a/lib/fpm/package.rb
+++ b/lib/fpm/package.rb
@@ -8,6 +8,7 @@ require "socket" # stdlib, for Socket.gethostname
 require "shellwords" # stdlib, for Shellwords.escape
 require "erb" # stdlib, for template processing
 require "cabin" # gem "cabin"
+require "stud/temporary"
 
 # This class is the parent of all packages.
 # If you want to implement an FPM package type, you'll inherit from this.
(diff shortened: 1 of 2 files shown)


eodgooch:
+1

luto:
Is there anything else you need to review or merge this pull request? @hatt, who also swiftly reviewed my other PR, already gave their :+1: :)

luto:
thank you!

PR #1140, opened 2016-06-22
freebsd pkg manifest needs "arch" over "abi" now

fetep:
Without an "arch" field, pkg just dumps core :(

"arch" is the new "abi" now, I guess:
https://github.com/freebsd/pkg/commit/95b86714ed4cea7c75e92499a2307f9141d3687d

Format description from here:
https://github.com/freebsd/pkg/blob/a3a2a2863d598702051a9e888d5194623c82ce51/libpkg/pkg_elf.c#L155-L159

### DIFF
--- a/lib/fpm/package/freebsd.rb
+++ b/lib/fpm/package/freebsd.rb
@@ -12,17 +12,19 @@ class FPM::Package::FreeBSD < FPM::Package
     :after_remove       => "post-deinstall",
   } unless defined?(SCRIPT_MAP)
 
-  def self.default_abi
-    abi_name = %x{uname -s}.chomp
-    abi_version = %x{uname -r}.chomp.split(".")[0]
-    abi_arch = %x{uname -m}.chomp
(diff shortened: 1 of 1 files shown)


hatt:
Any reason to even keep the `--arch` flag anymore? In RPM and DEB we use the `architecture` method to format the main flag FPM has. Looks like BSD was just using a separate flag because it was called ABI.

fetep:
The `architecture` method sounds fine; I'll poke at this soon and adjust.

jordansissel:
+1 on using the existing `-a`/`--architecture` flag (attributes[:architecture] from your package code).

Freeaqingme:
@fetep Were you already able to take a look at this? I just spent a few hours trying to figuring out why pkgng was segfaulting on me. Please let me know if there's anything I can do to help.

hatt:
An associated PR for this has been merged and closed. If @Freeaqingme or @fetep would like to pull down the latest HEAD and confirm, that'd be great.

PR #1141, opened 2016-06-22
Erase release candidate spec from deps when converting from Gem to RPM

jhull123:
This is the second coming of https://github.com/jordansissel/fpm/pull/1105, which I closed after horking the rebase.

---

A while ago I encountered an issue when building RPMs for Nokogiri and its dependencies. The issue is that Nokogiri lists a release candidate version of mini_portile2 as a dependency.

When I build this RC version of mini_portile2 the resulting RPM's version will not include the RC suffix (2.0.0 instead of 2.0.0.rc2). When I build Nokogirl, the RPM dependencies will include mini_portile2 v2.0.0.rc2. Thus, Nokogiri will fail to install (via yum) since the mini_portile2 dependency cannot be satisfied.

I made small tweak to FPM to erase release candidate suffixes from RPM dependencies in order to prevent this situation.

### DIFF
--- a/lib/fpm/package/rpm.rb
+++ b/lib/fpm/package/rpm.rb
@@ -302,6 +302,11 @@ class FPM::Package::RPM < FPM::Package
       end.flatten
     end
 
+    # Erase troublesome release candidate declarations.
+    self.dependencies = self.dependencies.collect do |dep|
+      dep.gsub(/.rc\d*/, '')
+    end
+
   setscript = proc do |scriptname|
       script_path = self.attributes[scriptname]
       # Skip scripts not set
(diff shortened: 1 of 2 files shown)


hatt:
I read over the original PR again and wouldn't it make more sense to keep the pre-release version identifier in the dependency package names rather than removing it in the dependency list? It seems more accurate that way.

RPM however doesn't like versions that aren't numerical so the general solution is to use the release tag. We could split the version on any character other than a hyphen, underscore, period, or number and then trim any trailing with the last match going into the release tag if it isn't set by the user. See https://fedoraproject.org/wiki/Packaging:NamingGuidelines#Release_Tag

jordansissel:
I think the way we solve prerelease rpms in Logstash is to use a ~ between
version and prerelease label in the version field, like fpm -v 1.2.3~foo

jordansissel:

Issue #1143, opened 2016-06-26
rpm-attr file permission for parent directory are ignored


LStuker:
If I specify rpm-attr for a parent directory it gets ignored in the spec file.

For example:

```bash
fpm -s dir -t rpm -n "myapp" -v 1.0 
--rpm-attr '0755,appuser,appgrp:"/var/lib/myapp/relation"
--rpm-attr '0755,appuser,appgrp:/var/lib/myapp/relation/tmp'
--rpm-attr '0755,appuser,appgrp:/var/lib/myapp/relation/data'
/opt /var/lib/myapp /etc/myapp/relation /usr/lib/systemd/system/
```

SPECS file output:
```
%files
%defattr(-,root,root,-)
%config(noreplace) /etc/myapp/relation/config.yml

%dir /opt/myapp/relation/
%dir /opt/myapp/relation/bin
# Reject config files already listed or parent directories, then prefix files
# with "/", then make sure paths with spaces are quoted. I hate rpm so much.
/opt/myapp/relation/bin/myappcmd
/usr/lib/systemd/system/myapp.service
%attr(0755,appuser,appgrp) /var/lib/myapp/relation/data

%attr(0755,appuser,appgrp) /var/lib/myapp/relation/tmp
```

At first I thought fpm doesn't support this because rpm doesn't support it. But I manually edit the SPEC files and the output rpm creates the directory permission as desire.

My code debugging showed that fpm, goes throw all files and check if they are in also definied in the attr hash, if so it would add it. But as subdirectories are never be part of the file list they will never be checked if they are defined in attrs.

https://github.com/jordansissel/fpm/blob/3338aa8f4b5ba90c1314f3c4887b33cee1c36f7e/templates/rpm.erb#L252
```ruby
<% files.each do |path| -%>
<%   path = "/#{path}" -%>
<%   next if config_files.include?(path)-%>
<%   next if directories.include?(path)-%>
<%= rpm_file_entry(path) %>
<% end -%>

[1] pry(#<FPM::Package::RPM>)> files
=> ["etc/myapp/relation/config.yml",
 "opt/myapp/relation/bin/myappcmd",
 "usr/lib/systemd/system/dhive-relation.service",
 "var/lib/myapp/relation/data",
 "var/lib/myapp/relation/tmp"]
```

https://github.com/jordansissel/fpm/blob/3338aa8f4b5ba90c1314f3c4887b33cee1c36f7e/lib/fpm/package/rpm.rb#L190

```ruby
[...]

PR #1144, opened 2016-06-27
RPM: add support to set attr for subdirectories

LStuker:
This pull request brings the feature to set rpm-attr for subdirectories.
#1143


LStuker:
I apologize, I saw that the pull request is not ready to be merged

PR #1146, opened 2016-07-01
prevents json dependency from creeping to version 2.0

mjhoffman65:
json 2.0.1 was published today and requires ruby's version
to be greater than 2.0.


jordansissel:
Sounds fine to me :)

jordansissel:
fpm 1.6.2 published with this patch.

PR #1154, opened 2016-07-05
Fix pacman dependency regex.

### LINKED ISSUE #1153: Dependencies with version requirements is broken in "pacman" mode
Hi,

Adding dependencies using "-d package-name" currently works when using "pacman" as the output type, however it doesn't work as soon as you use some version comparison operators.

For example, this doesn't work with "pacman":

    root@4e9526ad0d2c:/tmp# fpm -s empty -t "pacman" -n package-name -d "base-package-name=1.2.3"
    Some of the dependencies looked like they weren't package names. Such dependency entries only serve to confuse arch. I am removing them. {:removed_dependencies=>["base-package-name=1.2.3"], :fixed_dependencies=>[], :level=>:warn}
    Created package {:path=>"package-name-1.0-1-x86_64.pkg.tar.xz"}

but it works fine in "deb" mode:

    root@4e9526ad0d2c:/tmp# fpm -s empty -t "deb" -n package-name -d "base-package-name=1.2.3"
    Created package {:path=>"package-name_1.0_amd64.deb"}

Without version operators, it works fine as well.

Is this expected ?

Cheers,
Romain

Romain-Geissler-1A:
Fixes #1153.

This is based on https://wiki.archlinux.org/index.php/PKGBUILD#Dependencies which allows either '+', '.', '@' and '_' in either the package name or the version.

### DIFF
--- a/lib/fpm/package/pacman.rb
+++ b/lib/fpm/package/pacman.rb
@@ -58,7 +58,7 @@ class FPM::Package::Pacman < FPM::Package
   end # def config_files
 
   def dependencies
-    bogus_regex = /[^\sA-Za-z0-9><=-]/
+    bogus_regex = /[^\sA-Za-z0-9><=+._@-]/
     # Actually modifies depencies if they are not right
     bogus_dependencies = @dependencies.grep bogus_regex
     if bogus_dependencies.any?


Romain-Geissler-1A:
Ping

hatt:
Looks good to me, regex is fine. A test for valid+invalid characters might be nice though.

jordansissel:
LGTM. +1 on adding a test, but I won't block the PR.

PR #1155, opened 2016-07-07
Fix pacman packages having symlinks.

Romain-Geissler-1A:
Hi,

I am having issues with pacman packages involving symlinks. Symlinks are not keep in the final package, instead, they are copied. Trivial reproducer:

> root@c12e18c630ff:/tmp# mkdir package.src
> root@c12e18c630ff:/tmp# echo "some file" > package.src/some-file
> root@c12e18c630ff:/tmp# ln -s some-file package.src/some-file.symlink
> root@c12e18c630ff:/tmp# ls -l package.src/
> total 4
> -rw-r--r-- 1 root root 10 Jul  7 16:47 some-file
> lrwxrwxrwx 1 root root  9 Jul  7 16:47 some-file.symlink -> some-file
> root@c12e18c630ff:/tmp# fpm -s dir -t pacman -n package-name package.src
> Created package {:path=>"package-name-1.0-1-x86_64.pkg.tar.xz"}
> root@c12e18c630ff:/tmp# tar -tvf package-name-1.0-1-x86_64.pkg.tar.xz
> -rw-r--r-- 0/0             316 2016-07-07 16:48 .MTREE
> -rw-r--r-- 0/0             645 2016-07-07 16:48 .PKGINFO
> drwxr-xr-x 0/0               0 2016-07-07 16:48 package.src/
> -rw-r--r-- 0/0              10 2016-07-07 16:47 package.src/some-file.symlink
> -rw-r--r-- 0/0              10 2016-07-07 16:47 package.src/some-file

This can be fixed by removing the `preserve=true` argument when copying entries.

Cheers,
Romain

### DIFF
--- a/lib/fpm/package/pacman.rb
+++ b/lib/fpm/package/pacman.rb
@@ -245,7 +245,7 @@ class FPM::Package::Pacman < FPM::Package
     Find.find(staging_path) do |path|
       src = path.gsub(/^#{staging_path}/, '')
       dst = build_path(src)
-      copy_entry(path, dst, preserve=true, remove_destination=true)
+      copy_entry(path, dst, remove_destination=true)
       copy_metadata(path, dst)
     end
 


Romain-Geissler-1A:
Ping

hatt:
What version of Ruby are you on? I just tested this against 1.9.3 and 2.0-2.3 both with and without the change and got the same behavior both times (link is copied correctly). If this is an actual bug, I expect it's in `FileUtils#copy_entry` in specific versions or platforms of Ruby. In this case, it may be better to see how this behaves in RPM, DEB, etc and then modify `FPM::Util#copy_entry` to have a link case instead, just make sure to take `nlink` into account from the `File#stat` call to make sure not to mess up hard links.

As far as the `preserve` option goes, from the Ruby [documentation](ruby-doc.org/stdlib-2.3.0/libdoc/fileutils/rdoc/FileUtils.html#method-c-copy_entry), it only handles ownership and mtime and has nothing to do with symlinks. The method will always copy symlinks as is.

The expected behavior (to me) sounds like it should always have broken symlinks. Going by the Arch [documentation](https://wiki.archlinux.org/index.php/creating_packages#package.28.29), symlinks should be broken until installed at which time they point to the correct destination. Running your example with a prefix produces this exact behavior, so it looks fine to me and not a bug.

Romain-Geissler-1A:
I normally use ruby 2.1. I have just tried with different Ubuntu base images with ruby 1.9 and 2.3 as well, and the result is always the one expose in this issue: the symlink is turned into a normal file, which is a copy of the original one. I would have expected the symlink to remain a symlink when tar'ed and then installed as a package. All my tests are made inside a docker container, but I doubt this changes anything.

I don't quite get your last remark. For you it is expected that symlinks are turned into normal files rather than symlinks ?

hatt:
Sorry, I meant I expect it to remain a symlink as happened in my tests, not turn into a file as happened for you.

Digging a little deeper, it looks like the behavior differs on OSX vs Linux for when symlinks are dereferenced. I'll experiment a bit more when I get access to an Ubuntu and CentOS VM later. There are a few bug references around from like 10 years ago but they all seem fixed and I can't tell why this would be alright in OSX but not Ubuntu. Here's the definition that looks to be börking:

```
    def stat
      return @stat if @stat
      if lstat() and lstat().symlink?
        @stat = File.stat(path())
      else
        @stat = lstat()
      end
      @stat
    end

    def lstat
      if dereference?
        @lstat ||= File.stat(path())
      else
        @lstat ||= File.lstat(path())
      end
    end
```

I don't think this PR addresses the actual problem either way though, it looks like the weird way FileUtils and IO parts of Ruby stdlib work based on parameters is somehow doing it right only some of the time.

Issue #1125, opened 2016-05-31
--version fails if rc file is present


andresvia:
```
macs-Mac-mini:~ andres$ mkdir /tmp/fpm-test
macs-Mac-mini:~ andres$ cd /tmp/fpm-test/
macs-Mac-mini:fpm-test andres$ fpm --version
1.6.0
macs-Mac-mini:fpm-test andres$ echo -- --rpm-sign > .fpm 
macs-Mac-mini:fpm-test andres$ fpm --version
Loading flags from rc file .fpm {:level=>:warn}
Additional options: -- --rpm-sign {:level=>:warn}
All flags should be before the first argument (stray flags found: ["--rpm-sign", "--version"] {:level=>:warn}
Missing required -s flag. What package source did you want? {:level=>:warn}
Missing required -t flag. What package output did you want? {:level=>:warn}
Fix the above problems, and you'll be rolling packages in no time! {:level=>:fatal}
macs-Mac-mini:fpm-test andres$ 
```

jordansissel:
Oops! Good catch :)

PR #1162, opened 2016-07-20
Fix for #1125: --version fails when .fpm file exists

drwl:
I wrote a fix for issue #1125. I also wrote 2 tests to make sure that it works. This is my first time doing something like this, so let me know if there's anything I should change or that I can improve upon :).

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -257,7 +257,7 @@ class FPM::Command < Clamp::Command
   # Execute this command. See Clamp::Command#execute and Clamp's documentation
   def execute
     # Short-circuit if someone simply runs `fpm --version`
-    if ARGV == [ "--version" ]
+    unless (ARGV & ["-v", "--version"]).empty?
       puts FPM::VERSION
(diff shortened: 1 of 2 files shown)


lib/fpm/command.rb
```diff
@@ -257,7 +257,7 @@ def initialize(*args)
   # Execute this command. See Clamp::Command#execute and Clamp's documentation
   def execute
     # Short-circuit if someone simply runs `fpm --version`
-    if ARGV == [ "--version" ]
+    unless (ARGV & ["-v", "--version"]).empty?
```
jordansissel: This will print the version and exit if `-v` or `--version` appear anywhere in the flags. I don't think we want that. For example, this should not print the fpm version:


```
% bin/fpm -s empty -t rpm -n example -v 1.0
1.6.2
```

I have an idea on how to fix this, though.

jordansissel:
I made some slight changes based on my comment above and merged it.

Thank you for fixing this! <3

PR #1168, opened 2016-08-02
Update README.md with freebsd in the targets list

kylegato:
Adding freebsd to the README due to recently introduced functionality.


jordansissel:
Wooo :)

PR #1170, opened 2016-08-04
add missing apk require

lloydpick:
I'm not sure if this was left out intentionally (doubt it), however I'm trying to update fpm-cookery to work with APK packages, and without this require it doesn't appear in`FPM::Package.types` array.


jordansissel:
Good catch :)

Issue #564, opened 2013-10-02
Weird permissions on prefix/top-level directory for tar to deb


tpot:
Just tracking down a weird install issue and found that the permissions on the --prefix or top-level directory are set to 0700.  The value of umask does not seem to affect this.

$ fpm -f -s tar -t deb --prefix /opt/OV/HPBsmIntNagios \
    --name hpcs-omi-genint \
    --version 01.10.021-Linux2.6-64 \
    --deb-user nagios --deb-group nagios \
    dist/genint-01.10.021-Linux2.6_64.tar.bz2

$ dpkg-deb --contents hpcs-omi-genint_01.10.021-Linux2.6-64_amd64.deb | head 
drwxr-xr-x nagios/nagios     0 2013-10-02 00:54 ./
drwxr-xr-x nagios/nagios     0 2013-10-02 00:54 ./opt/
drwxr-xr-x nagios/nagios     0 2013-10-02 00:54 ./opt/OV/
drwx------ nagios/nagios     0 2013-10-02 00:54 ./opt/OV/HPBsmIntNagios/
drwxr-xr-x nagios/nagios     0 2013-10-02 00:54 ./opt/OV/HPBsmIntNagios/genint/
drwxr-xr-x nagios/nagios     0 2013-10-02 00:54 ./opt/OV/HPBsmIntNagios/genint/lib/

Interesting leaving off the --prefix has the same result, but the ./ directory is 0700 now:

drwx------ nagios/nagios     0 2013-10-02 00:56 ./
drwxr-xr-x nagios/nagios     0 2013-10-02 00:56 ./genint/
drwxr-xr-x nagios/nagios     0 2013-10-02 00:56 ./genint/lib/

fpm rocks - thanks!

AdamJacobMuller:
I am still seeing this issue with 1.4.0

edgan:
I made a pull request,  #1173 .

PR #1173, opened 2016-08-11
Fixing #564 by doing a chmod 0755 of the prefix directory after the c…

edgan:
…lone

### DIFF
--- a/lib/fpm/package/dir.rb
+++ b/lib/fpm/package/dir.rb
@@ -69,6 +69,7 @@ class FPM::Package::Dir < FPM::Package
       ::Dir.chdir(chdir) do
         begin
           clone(source, destination)
+          FileUtils.chmod(0755, destination)
         rescue Errno::ENOENT => e
           raise FPM::InvalidPackageConfiguration,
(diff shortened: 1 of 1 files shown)


hatt:
I'm not sure if it's better to comment here or on #564 but since this is the PR, will keep it here. I think because Tar sources come from a Tmpdir, it's inheriting the permissions used somehow. `Dir::Tmpdir` defaults to 0700 (see `Dir.mktmpdir` in `tmpdir.rb` in Ruby core.

The clone method which looks to be bork is just calling copy which in turn sets 0755, or at least it should. I'm not sure the PR really fixes the source of the issue and why this is happening but it's a valid enough workaround.

edgan:
My thought was the it could be argued that in general a 0700 on a temp directory isn't a bad policy. The code then directly clones the temp directory. So why not just fix it after the clone.

hatt:
I think that makes sense but also wonder why the call to `copy(:mode => 0755)` isn't being applied properly...

jordansissel:

PR #1182, opened 2016-08-23
Fix #1179 Remove .packlist files and empty parent directories from CPAN packages

### LINKED ISSUE #1179: Remove .packlist files from CPAN modules
Every converted CPAN module ends up with a `.packlist` file with useless junk in it, e.g.:

```
/tmp/package-cpan-staging20160819-10-1uhegjk/usr/local/bin/analyze_apache_logs
/tmp/package-cpan-staging20160819-10-1uhegjk/usr/local/share/man/man3/Apache::Log::Parser.3pm
/tmp/package-cpan-staging20160819-10-1uhegjk/usr/local/share/perl5/Apache/Log/Parser.pm
```
I propose that these are automatically removed in the same way as `perllocal.pod`.

liger1978:
@liger1978

`.packlist` files are useless junk containing irrelevant build information. In `noarch` packages they end up being the only file in the `lib64` directory when converting to e.g. RPMs (breaking the standard for `noarch` packages). This patch removes them from the package and also removes all empty parent directories of the unlinked `.packlist` file resulting in no more pollution of the `lib64` directory on the target system.

This is the second iteration of this patch after further testing.  This version breaks out of the `do` loop when a non-empty directory is found (more efficient) and sorts the directory contents before comparison to find empty dirs (annoyingly some platforms list contents as `[".", ".."]` and others as `["..", '.']`).

### DIFF
--- a/lib/fpm/package/cpan.rb
+++ b/lib/fpm/package/cpan.rb
@@ -3,6 +3,7 @@ require "fpm/package"
 require "fpm/util"
 require "fileutils"
 require "find"
+require "pathname"
 
 class FPM::Package::CPAN < FPM::Package
   # Flags '--foo' will be accessable  as attributes[:npm_foo]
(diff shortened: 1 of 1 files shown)


hatt:
Looks good to me.

jordansissel:
@liger1978 thank you for working on this :)
@hatt Thanks for reviewing!

PR #1185, opened 2016-08-31
Fix excludes

aeppert:
Excludes need to look at the staging_path, not the installdir.

### DIFF
--- a/lib/fpm/package.rb
+++ b/lib/fpm/package.rb
@@ -386,7 +386,7 @@ class FPM::Package
       installdir = staging_path
     end
 
-    Find.find(installdir) do |path|
+    Find.find(staging_path) do |path|
       match_path = path.sub("#{installdir.chomp('/')}/", '')
 
       attributes[:excludes].each do |wildcard|


jordansissel:
Sorry for the delays in review, and thank you for helping improve fpm!

This seems like a fairly significant behavioral change. I have two primary concerns so far:

1) It's unclear what was being fixed. If it needs to look at staging_path, what behavior is this fixing? What didn't work before that works with this patch?
2) There are no tests for this, so it's hard to absorb the change and keep it working in the future.

aeppert:
1. The excludes were, flatly, not working. We have been using fpm for ~2 years and suddenly needed to explicitly exclude a few entries from an otherwise fine directory structure generated via `make install` in one of our artifacts. However, when trying to exclude, it was failing, completely. 

2. A test for this would be trivial - Create a file test directory for packaging and include a few files a, b, c, d. Create an exclude file to ignore b,c. Enumerate the resulting package and verify b,c do not exist in the package.

In our case, our problems manifested with RPM, however, I assume the rest suffer the same problem.

PR #1188, opened 2016-09-02
Add long flag options for chdir, input, and output

### LINKED ISSUE #1187: fpms common command-line options should have long (e.g. --output_type) version, too.
`fpm` is so much fire-and-forget that I use it in some scripts and it basically just-works(tm). But on revisiting the scripts, I wonder about what e.g. the option `-s` or `-t` was about. Most options like '-f' have their speaking alternative, like '--force', which I will always use in scripts so that I understand directly what happens. Unfortunately, two often used options do not have a long version.

It would be great if these "short" version would be accompanied by respective long versions. I propose:

    option "-t", "OUTPUT_TYPE"
    -> option ["-t", "--output-type"], "OUTPUT_TYPE"

    option "-s", "INPUT_TYPE",
    -> option ["-s", "--input-type"], "INPUT_TYPE"

Give me a go and I branch and merge request. As a bonus I would add (and this one is more complicated :) ):

    -C CHDIR -> --chdir

hatt:
The flags for chdir, input, and output aren't immediately intuitive as to what they're short for. This commit adds long versions and a spec stub for chdir. Resolves #1187.

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -45,13 +45,13 @@ class FPM::Command < Clamp::Command
     return lines.join("\n")
   end # def help
 
-  option "-t", "OUTPUT_TYPE",
+  option ["-t", "--output-type"], "OUTPUT_TYPE",
     "the type of package you want to create (deb, rpm, solaris, etc)",
     :attribute_name => :output_type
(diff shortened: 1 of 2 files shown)


jordansissel:
LGTM.

PR #1196, opened 2016-09-28
Use global architecture switch for FreeBSD ABI

hatt:
Alternative implementation of #1140. I'm not super happy with the way the casing works but I think this can extend to cover platforms like ARM, MIPS, SPARC, etc down the track. There is a pseudo-authoritative list of valid architecture and wordsize options [here](https://github.com/freebsd/pkg/blob/203e665dcff3d669cbcc6fc26a44323d1fddd958/libpkg/private/elf_tables.h#L83) but I've only really kept the x86 parts of it. Since i386 and x86_64 both use the same architecture (x86), I've just set this as is for now.

Overall I feel like this implementation is ugly but I can't think of a nicer way to do it. This is at least more consistent with what the other package architecture mutators do, like for RPM or Deb.

### DIFF
--- a/lib/fpm/package/freebsd.rb
+++ b/lib/fpm/package/freebsd.rb
@@ -12,18 +12,6 @@ class FPM::Package::FreeBSD < FPM::Package
     :after_remove       => "post-deinstall",
   } unless defined?(SCRIPT_MAP)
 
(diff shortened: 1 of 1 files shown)


jordansissel:
Wooo <3

PR #1199, opened 2016-10-05
Correctly create device nodes on Linux

dividedmind:
Pointer that used to be passed to xmknod was
FFI::MemoryPointer.new(dev), which created an array of `dev`
size. Correctly construct the pointer instead.

Also use Stat#rdev, not #dev (which is the device where the
node is in, not which it points to).

### DIFF
--- a/lib/fpm/util.rb
+++ b/lib/fpm/util.rb
@@ -252,7 +252,9 @@ module FPM::Util
     case %x{uname -s}.chomp
     when 'Linux'
       # bits/stat.h #define _MKNOD_VER_LINUX  0
-      rc = xmknod(0, path, mode, FFI::MemoryPointer.new(dev))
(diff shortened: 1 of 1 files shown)


jordansissel:
Is there a test we can add to cover this? otherwise, I'm not sure how to reproduce any failure that this patch fixes.

dividedmind:
This is kinda hard, because you need root to actually execute any of this
code.

But try making a package from dir to rpm which includes device files and
you'll see the problem.
5 paź 2016 19:20 "Jordan Sissel" <notifications@github.com> napisał(a):

Is there a test we can add to cover this? otherwise, I'm not sure how to
reproduce any failure that this patch fixes.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/jordansissel/fpm/pull/1199#issuecomment-251827294>, or mute
the thread

PR #1209, opened 2016-10-15
Fix failing virtualenv builds

nicobrevin:
Seems like #1124 made tmp dirs long enough to trip a kernel-defined size limit of shebangs.  This change removes the reliance on the shebangs in the temporary virtualenvs.


nicobrevin:
I think I might clean a few more things up then PR again

PR #1210, opened 2016-10-15
Bugfix virtualenv various

nicobrevin:
Looks like #1124 broke some of the virtualenv features, this fixes that, plus another bug, and a blind attempt at getting travis to run the python virtualenv tests


jordansissel:
Specs failed on OSX, but this seems to happen often enough.

LGTM.

Issue #1212, opened 2016-10-17
fpm should support bz2 format of data


shalq:
# fpm  --iteration 37.56 --verbose --debug -s deb -t deb linux-image-4.4.0-37_4.4.0-37.56_amd64.deb

Invalid package configuration: Unknown compression type 'gz'  {:level=>:error, :file=>"fpm/command.rb", :line=>"495", :method=>"rescue in execute"}
# ar t /opt/redstone/builds/workload/ubuntu-xenial.wiki/linux-image-4.4.0-37_4.4.0-37.56_amd64.deb

debian-binary
control.tar.gz
data.tar.bz2

To fix this issue, fpm should support bz2 .

jordansissel:
This only affects `-s deb` right?

jordansissel:
#1213 merged.

PR #1213, opened 2016-10-17
deb data file should support bz2 format

shalq:
Fix issue 1212 https://github.com/jordansissel/fpm/issues/1212


shalq:
recheck

jordansissel:
Thanks for this!

This feature would benefit from having tests to ensure it continues to work in the future, but I wont' require it for this PR.

PR #1218, opened 2016-10-23
make virtualenv packages truly standalone

nzjrs:
this branch contains two improvements to allow using fpm to distribute end user python applications _and_ their dependencies together, in one virtualenv based package.


nzjrs:
ping? this is a pretty large increase in functionality for a handful of lines of code

jordansissel:
Code looks good. I wish for tests, but I won't block on it.

jordansissel:
Thank you very much for this improvement! <3

PR #1235, opened 2016-11-25
pleaserun: support chdir parameter

cfstras:
Added support for --pleaserun-chdir to specify the working directory of a
created service.


jordansissel:
Thank you :)

PR #1241, opened 2016-11-28
Fixes #1236 Fixed version number for metacpan search

### LINKED ISSUE #1236: Failing to download a CPAN module
FPM Fails to download Rex-Repositorio CPAN module when converting.  Will investigate.

liger1978:
The new way of searching metacpan with an http `POST` consistently requires numerical version number (i.e. no `v` prefix).  This patch tested against 73 CPAN modules: https://gitlab.com/harbottle/epmel/builds/6722395

### DIFF
--- a/lib/fpm/package/cpan.rb
+++ b/lib/fpm/package/cpan.rb
@@ -309,17 +309,16 @@ class FPM::Package::CPAN < FPM::Package
                  :distribution => distribution,
                  :version => cpan_version)
 
-    # default to latest versionunless we specify one
+    # default to latest version unless we specify one
     if cpan_version.nil?
-      self.version = metadata["version"]
+      self.version = "#{metadata["version"]}"
     else
-      if metadata["version"] =~ /^v\d/
-        self.version = "v#{cpan_version}"
-      else
-        self.version = cpan_version
-      end
+      self.version = "#{cpan_version}"
     end
-
+    
+    # remove 'v' prefix from version if it is there
+    self.version.sub!(/^v/, '')
+    
     # Search metacpan to get download URL for this version of the module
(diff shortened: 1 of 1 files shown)


liger1978:
It's only the OSX build failing travis-ci (as usual)...

lib/fpm/package/cpan.rb
```diff
@@ -309,17 +309,16 @@ def download(metadata, cpan_version=nil)
                  :distribution => distribution,
                  :version => cpan_version)
 
-    # default to latest versionunless we specify one
+    # default to latest version unless we specify one
     if cpan_version.nil?
-      self.version = metadata["version"]
+      self.version = "#{metadata["version"]}"
```
jordansissel: this can just be `self.version = metadata["version"]` I think.

jordansissel:
Lovely thank you! I made one comment for a style change, but it's minor and I am happy to merge.

jordansissel:
@liger1978 how would you characterize this change, a bug fix? new feature?

lib/fpm/package/cpan.rb
liger1978: It was just belt and braces to ensure self.version was a string before attempting the substitution below.

liger1978:
@jordansissel A bug

lib/fpm/package/cpan.rb
jordansissel: Gotcha. My preferred way to do this is to call `.to_s` on the object: `metadata["version"].to_s`

No change required. Thanks for the explanation :)

jordansissel:
@liger1978 ok cool, noted for the release notes <3

PR #1242, opened 2016-11-28
First stab at rpm docs

nicobrevin:
Hey, this look like the sort of thing you're after?  If so, I can go and do this for a bunch of stuff I have some experience with.

### DIFF
--- a/docs/source/rpm.rst
+++ b/docs/source/rpm.rst
@@ -1,4 +1,23 @@
 `rpm` - RPM Packages
 ====================
 
-Hello
+Synopsis::
+
+  fpm -s rpm [other options] path-to-rpm
+
+Using 'rpm' as a source lets you treat an existing package as a source for
+building a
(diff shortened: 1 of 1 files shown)


docs/source/rpm.rst
```diff
@@ -1,4 +1,23 @@
...
+Synopsis::
+
+  fpm -s rpm [other options] path-to-rpm
+
+Using 'rpm' as a source lets you treat an existing package as a source for
+building a
+new one.  This can be useful for converting packages between formats or
+for "editing" upstream packages.
+
+Strip out docs under `/usr/share/doc`::
```
jordansissel: How about making this a separate section, like adding a header:

```
Example Uses
------------
```

jordansissel:
:+1:  on this early work.

nicobrevin:
Yeah ok, sounds good, I'll do that and add some docs for the virtualenv
stuff as well

PR #1246, opened 2016-11-30
Some docs for virtualenv support

nicobrevin:
Another one.

PS I'm not particularly happy with the inconsistencies in the virtualenv support, it looks like someone else's conventions have been baked in to the way the options are structured, instead of just using prefix </whinge>.  I've thought about changing it, but this will probably make lots of people sad.  Thoughts?

### DIFF
--- a/docs/source/virtualenv.rst
+++ b/docs/source/virtualenv.rst
@@ -1,0 +1,40 @@
+`virtualenv` - Python virtual environments
+==========================================
+
+Synopsis::
+
+  fpm -s virtualenv [other options] EGG_SPEC|requirements.txt
+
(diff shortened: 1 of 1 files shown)


josegonzalez:
@nicobrevin worth discussing in a separate issue as to what should be changed/updated/fixed.

nzjrs:
It's worth documenting that the virtualenv support doesn't quite do what one would expect - put the app *and* dependencies into one virtualenv package.

I added that feature here https://github.com/jordansissel/fpm/pull/1218

grote:
Can't this be closed?

kenyon:
Should be closed, this doc was added in #1262.

PR #1249, opened 2016-12-01
Revert "symlinks need full dest path for FileUtils.copy_entry"

### LINKED ISSUE #1135: [REGRESSION] Copying of symbolic links to debian packages is broken
When copying symbolic links to debian packages, a directory is created and the symlink is placed inside it.

```
/tmp ✔ $ echo test > test
/tmp ✔ $ ln -s test link
/tmp ✔ $ ls -l test link
lrwxrwxrwx 1 luca luca 4 Jun 17 19:06 link -> test
-rw-rw-r-- 1 luca luca 5 Jun 17 19:05 test
/tmp ✔ $ fpm -sdir -tdeb -n test -v 1 ./test=/dest/test ./link=/dest/link
Created package {:path=>"test_1_amd64.deb"}
/tmp ✔ $ dpkg --contents test_1_amd64.deb
drwx------ 0/0               0 2016-06-17 19:06 ./
drwxrwxr-x 0/0               0 2016-06-17 19:06 ./usr/
drwxrwxr-x 0/0               0 2016-06-17 19:06 ./usr/share/
drwxrwxr-x 0/0               0 2016-06-17 19:06 ./usr/share/doc/
drwxrwxr-x 0/0               0 2016-06-17 19:06 ./usr/share/doc/test/
-rw-r--r-- 0/0             152 2016-06-17 19:06 ./usr/share/doc/test/changelog.Debian.gz
drwxrwxr-x 0/0               0 2016-06-17 19:06 ./dest/
drwxrwxr-x 0/0               0 2016-06-17 19:06 ./dest/link/
lrwxrwxrwx 0/0               0 2016-06-17 19:06 ./dest/link/link -> test
-rw-rw-r-- 0/0               5 2016-06-17 19:05 ./dest/test
/tmp ✔ $ gem list fpm

*** LOCAL GEMS ***

fpm (1.4.0)
```

Note the `./dest/link/link -> test`, it should be `./dest/link -> test`.

This used to work. Unfortunately I don't know exactly in which version, but we had a script that used to work like this and on a recent install it doesn't work anymore.

nemanja-boric-sociomantic:
This reverts commit 0de9c3aa0c152ace64df7982f2d09b20730635e4.

The reason for reverting is that this disallows the same usage
pattern as expected from rsync:

my-sym-link.so=/usr/lib/my-sym-link.so

The patch that this reverts will create directory, and put the
symlink inside `/usr/lib/my-sym-link.so/my-sym-link.so
which is very surprising and it doesn't follow the same
pattern as file copying is doing.

Fixes #1135

### DIFF
--- a/lib/fpm/package/dir.rb
+++ b/lib/fpm/package/dir.rb
@@ -154,7 +154,7 @@ class FPM::Package::Dir < FPM::Package
         copy(source, destination)
       end
     elsif fileinfo.symlink?
-      copy(source, File.join(destination, source))
+      copy(source, destination)
     else
       # Copy all files from 'path' into staging_path
       Find.find(source) do |path|


nemanja-boric-sociomantic:
Can somebody point me how to write the test for this project, so we don't end up in regression?

nemanja-boric-sociomantic:
This introduces a regression where the `dir.input` is not fed with `a=b` syntax. I'll reopen another PR with this fixed.

PR #1253, opened 2016-12-05
In case of a=b syntax where a is symlink, don't create directory

nemanja-boric-sociomantic:
Previously, if the symlink is included into the package and specified
with

my-sym-link.so=/usr/lib/my-sym-link.so

fpm would create directory, and put the
symlink inside `/usr/lib/my-sym-link.so/my-sym-link.so
which is very surprising and it doesn't follow the same
pattern as file copying is doing.


nemanja-boric-sociomantic:
I've added the test specifications for symlinks. Also the failure doesn't seems to be related to this PR, but something else is broken.

jordansissel:
Thank you for this improvement :)

Issue #1264, opened 2016-12-26
Dependent gem json v1.8.3 breaks fpm gem install on ruby 2.4


NeoCN:
when ```gem install fpm``` on ruby 2.4：

```
gem install fpm
Building native extensions.  This could take a while...
ERROR:  Error installing fpm:
	ERROR: Failed to build gem native extension.

    current directory: /usr/local/lib/ruby/gems/2.4.0/gems/json-1.8.3/ext/json/ext/generator
/usr/local/opt/ruby/bin/ruby -r ./siteconf20161226-62014-1huvp4l.rb extconf.rb
creating Makefile

current directory: /usr/local/lib/ruby/gems/2.4.0/gems/json-1.8.3/ext/json/ext/generator
make "DESTDIR=" clean

current directory: /usr/local/lib/ruby/gems/2.4.0/gems/json-1.8.3/ext/json/ext/generator
make "DESTDIR="
compiling generator.c
generator.c:861:25: error: use of undeclared identifier 'rb_cFixnum'
    } else if (klass == rb_cFixnum) {
                        ^
generator.c:863:25: error: use of undeclared identifier 'rb_cBignum'
    } else if (klass == rb_cBignum) {
                        ^
...
6 warnings and 2 errors generated.
make: *** [Makefile:242: generator.o] Error 1

make failed, exit code 2

Gem files will remain installed in /usr/local/lib/ruby/gems/2.4.0/gems/json-1.8.3 for inspection.
Results logged to /usr/local/lib/ruby/gems/2.4.0/extensions/x86_64-darwin-16/2.4.0/json-1.8.3/gem_make.out
```

skyzyx:
Running into this at this exact moment. Building on CentOS 7.3 with Ruby 2.4.0, I have a shorter error.

```
$ cat /usr/local/lib/ruby/gems/2.4.0/extensions/x86_64-linux/2.4.0-static/json-1.8.3/gem_make.out

current directory: /usr/local/lib/ruby/gems/2.4.0/gems/json-1.8.3/ext/json/ext/generator
/usr/local/bin/ruby -r ./siteconf20161228-21714-1t3vbie.rb extconf.rb
creating Makefile

current directory: /usr/local/lib/ruby/gems/2.4.0/gems/json-1.8.3/ext/json/ext/generator
make "DESTDIR=" clean

current directory: /usr/local/lib/ruby/gems/2.4.0/gems/json-1.8.3/ext/json/ext/generator
[...]

PR #1266, opened 2016-12-30
Update Travis config with Ruby 2.3 and 2.4.

skyzyx:
* Adds known-good Ruby minor versions to the `.travis.yml` file for 2.3.x and 2.4.x.
* The Travis checks in this PR will fail until #1264 is fixed. This is, literally, TDD.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ rvm:
   - 2.0.0
   - 2.1.1
   - 2.2.1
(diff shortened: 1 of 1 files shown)


skyzyx:
ping @jordansissel

jordansissel:
I am ok with merging this, but be aware that I don't watch Travis reports
for this projects due to false positives and other noise.

-Jordan

PR #1268, opened 2017-01-05
Change Gem dependency due to upstream rename

### LINKED ISSUE #1197: Use minitar instead of archive-tar-minitar
archive-tar-minitar is being dropped in Debian in favor of minitar which is more recent. What do you think of upgrading?

Reference: https://groups.google.com/forum/#!topic/linux.debian.devel/vQkQvvn1Pek

In addition, when trying to package fpm as a native debian package, I have a problem with this dependency (because the transitional ruby-archive-tar-minitar package doesn't provide any ruby source)

aaronjcopley:
Closes #1197

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -59,8 +59,8 @@ You should see output that looks like this::
     Successfully installed ffi-1.9.14
     Fetching: childprocess-0.5.9.gem (100%)
     Successfully installed childprocess-0.5.9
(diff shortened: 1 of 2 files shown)


halostatue:
The upstream has only just renamed with `minitar` 0.6; the previous version was a fork (retroactively permitted).

aaronjcopley:
I probably could have phrased it better, but do you see what the original person meant with #1197?

Issue #880, opened 2015-03-31
fpm builds in all dependencies for a npm module into the package


ngrennan-inflection:
I found "fpm -s npm -t rpm gulp-imagemin" builds a 17mb rpm of gulp-imagemin and all it's dependencies. This is insanity.

Alternatively, if I exclude all the dependencies, "fpm -s npm -t rpm -x lib/node_modules/gulp-imagemin/node_modules gulp-imagemin", I get a 5k rpm.

fpm needs to do proper dependency management for npm modules.

Yes, I found this in the code. Here is that feature request.

   # npm installs dependencies in the module itself, so if you do
    # 'npm install express' it installs dependencies (like 'connect')
    # to: node_modules/express/node_modules/connect/...
    #
    # To that end, I don't think we necessarily need to include 
    # any automatic dependency information since every 'npm install'
    # is fully self-contained. That's why you don't see any bother, yet,
    # to include the package's dependencies in here.
    #
    # It's possible someone will want to decouple that in the future,
    # but I will wait for that feature request.

ngrennan-inflection:
rpm -qlp node-gulp-imagemin-2.2.1-1.x86_64.rpm | grep lib/node_modules/gulp-imagemin/node_modules | sed 's/node_modules/\nnode_modules/g' | sort | uniq | grep '\.js$' | cut -f2 -d/ | sort | uniq | wc -l
224

It recursively pulls in 224 npm modules.

PR #1286, opened 2017-01-19
allow npm dependencies to be decoupled from the target package

dspinoz:
Hope this helps for #880, I have found it quite useful in reducing package sizes and decoupling the dependencies.
I was also looking at implementing a solution that allowed the dependency packages to be built as additional, but separate, packages. This would require some additional changes that I hope to address in another pull request.
fpm is awesome by the way! :+1:

### DIFF
--- a/lib/fpm/package/npm.rb
+++ b/lib/fpm/package/npm.rb
@@ -17,6 +17,30 @@ class FPM::Package::NPM < FPM::Package
   option "--registry", "NPM_REGISTRY",
     "The npm registry to use instead of the default."
 
+  option "--nodeps", :flag, "Should the target package include dependencies",
+    :default => false
+    
+  option "--set-dependencies", :flag, "Set dependencies on the target package",
+    :default => false
+    
+  option "--dependencies-op", "OPERATOR", "Operator to use for dependencies version matching",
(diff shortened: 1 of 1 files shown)


jordansissel:
Hmm.. I wonder if we could turn all of these new flags into a single flag.

When reading your proposal, it feels like we can summarize this behavior into two paths:

1) Should the node_modules dependency be internal? (ie: there's a node_modules directory in the package)
2) or, Should the dependencies be external? (meaning: the package lists them as dependencies like the `--depends` flag does)

Maybe the flag could be `--external-dependencies`, defaulting false, and when true will exhibit the new behavior you're adding (that we list dependencies in package metadata instead of including them in the package as files)

Thoughts?

dspinoz:
Hi! Yes I agree with your proposal, that sounds good to me. When I get a chance I will update my branch as appropriate.

In the end, for my specific use case, I didn't use this new functionality as "external-dependencies" sent me deep into dependency hell. I do still wonder if things like `npm shrinkwrap` could have helped - perhaps this is something to consider for this implementation? (sorry, I don't think I'm experienced enough to know best practices with npm/node in production)

PR #1291, opened 2017-02-17
Update installing.rst

PabloCastellano:
Apart from gcc and make you need other dependencies.
Without build-essentials I had an error complaining about "stdio.h not found"


jordansissel:
Thank you! This is sure to help future users who would have otherwise stumbled. <3

PR #1299, opened 2017-03-01
Update fpm.rb

faraazahmad:
Remove the explicit requiring of all the packages. Instead, add a small snippet to automatically require all the packages in the package directory.

### DIFF
--- a/lib/fpm.rb
+++ b/lib/fpm.rb
@@ -1,20 +1,8 @@
 require "fpm/namespace"
 
 require "fpm/package"
-require "fpm/package/dir"
-require "fpm/package/gem"
-require "fpm/package/deb"
-require "fpm/package/npm"
-require "fpm/package/rpm"
-require "fpm/package/tar"
-require "fpm/package/cpan"
-require "fpm/package/pear"
-require "fpm/package/empty"
-require "fpm/package/puppet"
-require "fpm/package/python"
-require "fpm/package/osxpkg"
-require "fpm/package/solaris"
-require "fpm/package/p5p"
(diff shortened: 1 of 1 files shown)


jordansissel:
I think I understand the proposal, but I am not sure what the benefit is - Can you tell me why you want this change?

faraazahmad:
It reduces the number of lines of code and you don't have to worry about ```require```ing new files explicitly. They'll be automatically included. If it's not a helpful feature, you can close the issue. ✌️

jordansissel:
In testing this patch, I find it doesn't appear to work?

```
% pwd
/home/jls/projects/fpm

% bin/fpm --help
/home/jls/projects/fpm/lib/fpm.rb:5:in `open': No such file or directory @ dir_initialize - fpm/package (Errno::ENOENT)
```

faraazahmad:
Can you show the code? You may have to use ```require_relative``` instead of ```require```

faraazahmad:
I think I'll have to ```require_relative```

josephholsten:
I'm not sure we really want this, for the same reasons `autoload` is being deprecated: https://bugs.ruby-lang.org/issues/5653

jordansissel:
Interesting thread on ruby-lang. I've never wanted Autoload, but it's cool that Autoload is thread-safe in JRuby, but maybe still not safe in MRI? FPM itself doesn't use threads right now, so maybe it's not an issue.

Personally, a long list of requires doesn't bother me much. I think we could close this or fix it to work -- either is fine with me.

jordansissel:

Issue #1307, opened 2017-03-28
Python packages: FPM ignores dependencies markers (PEP0508)


penguinolog:
In setup.py use:
```install_requires=[
        'six',
        'funcsigs >= 1.0; python_version == "2.7"',
        'typing >= 3.5; python_version == "3.4"'
    ]
```
During build markers should be evaluated and not built, if not enforced.
Currently all requirements is added to result requirements.

PR #1308, opened 2017-03-28
Process PEP0508 markers on python package requirements

penguinolog:
Process PEP0508 markers on python package requirements

Closes-bug: #1307

### DIFF
--- a/lib/fpm/package/pyfpm/get_metadata.py
+++ b/lib/fpm/package/pyfpm/get_metadata.py
@@ -47,6 +47,11 @@ class get_metadata(Command):
 
     def process_dep(self, dep):
         deps = []
+        if hasattr(dep, 'marker') and dep.marker:
+            # PEP0508 marker present
+            if not dep.marker.evaluate():
+                return deps
+
         if dep.specs:
             for operator, version in dep.specs:
                 deps.append("%s %s %s" % (dep.project_name,
--- a/spec/fixtures/python/requirements.txt
+++ b/spec/fixtures/python/requirements.txt
@@ -1,2 +1,4 @@
 rtxt-dep1 > 0.1
 rtxt-dep2 == 0.1
+rtxt-dep3; python_version == "2.0"
+rtxt-dep4; python_version > "2.0"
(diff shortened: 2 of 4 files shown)


penguinolog:
No any changes over tests has been providden,
https://packaging.pypa.io/en/latest/markers/ was used as marker API description and https://setuptools.readthedocs.io/en/latest/pkg_resources.html#requirements-parsing as record API description.

Tested on example from bug description.

jordansissel:
I confess that I don't quite understand what this fixes. For disclosure, I don't know what these 'markers' are or what problem is fixed by this patch. I read the parent issue (#1307) and did not reach much clarity.

This is ok, though. 

Code looks OK, but I'd like it if someone else with more recent python experience especially around this PEP-0508 area would review this.

Alternately, adding tests might help this get merged more quickly as the tests could help ensure this fix remains fixed.

penguinolog:
@jordansissel Tests updated: added 2 requirements, 1 of which should be ignored.

pohmelie:
@jordansissel 
This patch affect optional dependencies.
Real life example: python 3.5 introduced new module `typing`, which is backported as separate module to pypi. Dependency for third package may looks like:
```
install_requires=[
    'typing; python_version < "3.5"',
]
```
And installing this third package on python 3.5 or greater means, that this package have no external dependencies. The opposite case is when install on python 2.7 (for example), this third package should have `typing` dependency in result package metainformation.

Vote for this to be implemented with both hands! (can't review code, since I'm not familiar with ruby)

penguinolog:
Main patch is implemented on python helper side and on Ruby side was updated unit tests only.

pohmelie:
@penguinolog 
Do you mean this feature in upstream already?

penguinolog:
@pohmelie no, I just explain languages of this PR

papermakkusu:
Yo!
Code looks fine to me but I do not get the idea behind these changes. No surpriises here scince I'm not familiar with the code.

jordansissel:
Given the feedback so far, I'm content to merge. The Travis-CI failures seem unrelated to this PR.

PR #1309, opened 2017-03-28
Add rubygems to CentOS install line

jesusbagpuss:
For CentOS7, I needed to install rubygems.
Not sure if it is included in e.g. ruby-devel for previous versions?


jordansissel:
Thank you for improving fpm!

Issue #1313, opened 2017-03-31
Creating zip from dir prepends random tmpdir to packaged file paths.


pillarsdotnet:
When using fpm with source `dir` and target `zip`, a random tempdir is prepended to packaged file paths.

To demonstrate:

    mkdir foo
    echo test > foo/bar
    fpm -s dir -t zip -n foo -p foo.zip foo
    unzip -v foo.zip

Expected output:

    Archive:  foo.zip
     Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
    --------  ------  ------- ---- ---------- ----- --------  ----
           0  Stored        0   0% 2017-03-31 18:42 00000000  foo/
           5  Stored        5   0% 2017-03-31 18:42 3bb935c6  foo/bar
    --------          -------  ---                            -------
           5                5   0%                            2 files

Actual output:

    Archive:  foo.zip
     Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
    --------  ------  ------- ---- ---------- ----- --------  ----
           0  Stored        0   0% 2017-03-31 18:45 00000000  tmp/package-dir-staging-
    43e99d3c99bf3e340d26126f2a1002b341d844c5454e7a2442cbcf15cd0d/
           0  Stored        0   0% 2017-03-31 18:45 00000000  tmp/package-dir-staging-
    43e99d3c99bf3e340d26126f2a1002b341d844c5454e7a2442cbcf15cd0d/foo/
           5  Stored        5   0% 2017-03-31 18:42 3bb935c6  tmp/package-dir-staging-
    43e99d3c99bf3e340d26126f2a1002b341d844c5454e7a2442cbcf15cd0d/foo/bar
    --------          -------  ---                            -------
           5                5   0%                            3 files

pillarsdotnet:
Opened [PR#1314](https://github.com/jordansissel/fpm/pull/1314)

pillarsdotnet:
PR was merged nearly two years ago.

PR #1314, opened 2017-03-31
Fix zipfile target.

pillarsdotnet:
See https://github.com/jordansissel/fpm/issues/1313

### DIFF
--- a/lib/fpm/package/zip.rb
+++ b/lib/fpm/package/zip.rb
@@ -41,7 +41,7 @@ class FPM::Package::Zip < FPM::Package
     output_check(output_path)
     realpath = Pathname.new(output_path).realdirpath.to_s
     ::Dir.chdir(staging_path) do
-      safesystem("zip", "-9ry", realpath, ".")
+      safesystem("zip", "-9r", realpath, ".")
     end
   end # def output
 end # class FPM::Package::Tar
--- a/lib/fpm/package/zip.rb
+++ b/lib/fpm/package/zip.rb
@@ -36,28 +36,12 @@ class FPM::Package::Zip < FPM::Package
     dir.cleanup_build
   end # def input
 
-  # Output a tarball.
-  #
-  # If the output path ends predictably (like in .tar.gz) it will try to obey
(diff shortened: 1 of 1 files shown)


pillarsdotnet:
Dunno why travis-ci is failing, but a trivial change (adding a line to README.rst) also fails.

lib/fpm/package/zip.rb
```diff
@@ -36,28 +36,12 @@ def input(input_path)
...
-        return "-j"
-      when /\.tar\.gz$|\.tgz$/
-        return "-z"
-      when /\.tar\.xz$/
-        return "-J"
-      else
-        return nil
+    realpath = Pathname.new(output_path).realdirpath.to_s
+    ::Dir.chdir(staging_path) do
+      safesystem("zip", "-9ry", realpath, ".")
```
jordansissel: This is a behavioral change (adding -r and -y flags). What is the expected impact on users? I imagine the `-y` flag to build the zip will not work when extracting the resulting zip on Windows?

jordansissel:
Travis is failing due to an integration problem between osx and fpm's test suite. I ignore this particular error.

lib/fpm/package/zip.rb
jordansissel: The reason I mention this is that the PR and linked issue focus on fixing the directory path problem, and this is fixed by other parts of the patch (the two previous lines (`realpath` and `Dir.chdir`), and adding `-9ry` seems unrelated to the bug fix.

pillarsdotnet: the "-r" is necessary; the "-y" is helpful on unix but (...testing...) sadly unsupported on Windows.  I'll re-roll the patch with out the "-y" flag.

bazimov:
Anything blocking this MR, we are seeing same issue on zip functionality. This MR fixes that issue.

jordansissel:
I'd like to see more tests (who wouldn't! hehe) for the zip output, but I won't block this PR on that.

Thank you for helping make fpm more awesome :)

PR #1320, opened 2017-04-03
Fix OSX Python tests

robyoung:
The test was failing because the calculated staging path is incorrect
with brew. If done in a virtualenv it is correct.


robyoung:
Finally, go the tests passing 😌

robyoung:
Is there anything I can do to this PR to help it get accepted?

jordansissel:
There are only so many hours in the day.

PR #1321, opened 2017-04-07
docs/installing: Fix build deps on Debian-derived systems

ge-fa:
The meta package is called `build-essential` not `build-essentials`.


jordansissel:
LGTM. Thank you <3

Issue #1287, opened 2017-01-20
upstart init.d shim 


dodophoenix:
Due to inserting the shim as file named xxx.conf 
latest debian und ubuntu show xxx.conf as service name. 
    
    service --status-all
    [ - ] ...
    [ - ]  xxx.conf`

Running it by invoking 
`service xxx.conf start`
 ends up in
`initctl: Unknown job: xxx.conf`

This also infects the auto-completion in bash.

I would suggest to omit the .conf suffix when storing the shim to `etc/init.d`
@see https://github.com/jordansissel/fpm/blob/master/lib/fpm/package/deb.rb#L504

PR #1325, opened 2017-04-20
fix wrong init.d shim naming

vbakayev:
see #1287


vbakayev:
problem introduced in this commit: a24cabb4

itzg:
It doesn't appear this made it into 1.8.1. I was wondering if a release could be cut for this solution? 

Thanks.

georgealton:
I'm encountering the same issue, running 1.8.1

Issue #1176, opened 2016-08-19
Disabling changelog ?


ferrieux:
Maybe I'm misising the obvious, but in dir mode, since --deb-changelog and --rpm-changelog have been added, when those options are *not* passed,  it seems FPM insists on creating a dummy changelog anyway, with a target in /usr/share/doc.

What if I really want control over the files and directories my package will create ? I don't want an /usr/share/doc to be created, nor even a changelog for that matter...

I tried "--deb-no-changelog" and  "--deb-changelog none" to no avail.

jordansissel:
@ferrieux I agree with your problem and proposal. As for implementation, a new flag seems OK to me. Thoughts?

The history behind why there is a dummy changelog file is because some folks (I don't remember who?) complained that `lintian` was reporting their fpm-created packages to be invalid because of a missing changelog file. Links: https://github.com/jordansissel/fpm/commit/c6f66dd0 and #784.

hatt:
Any reason not to have say `--changelog` as a singular flag and default it to a dummy and then offer `--no-changelog` for anyone who doesn't want it? Seems alright given RPM and Deb both support it along with most other output formats. I suppose for anything that doesn't support changelogs the flag could just be ignored.

TBH I never got why anyone cared about Lintian since it seemed like people that subscribed to Debian's policies just used Debian tools to begin with.

ferrieux:
I'm okay with *any* solution, provided it works; so, thanks in advance for implementing one :)

The only unease I feel is in the past: when the Lintian guys came asking for a way to add a changelog, why not just *add* a specific to that effect for them, *without* changing the default behavior ? Backwards compat is cool !!! . But that's too late, so never mind ;-)

jordansissel:
@ferrieux the reason for the change was that `lintian`'s default behavior changed to make it an error if there's no changelog, and #784 mentions that this behavior change in `lintian` caused the fpm test suite to fail (https://github.com/jordansissel/fpm/blob/master/spec/fpm/package/deb_spec.rb#L366-L376).

It's a grey area, but my feeling was that a test that used to pass but then failed is a bug, not a feature, and we should fix bugs and not treat bad behavior as a feature that needs backwards compat. As said, it's a grey area for sure ;P

Alexander-Shukaev:
Would be nice to have this feature finally.

Alexander-Shukaev:
I also don't see how one can change the default location of it.  It seems strange that `--prefix` has no influence on it.

PR #1331, opened 2017-05-19
Adds --deb-no-changelog flag

paulhamby:
This is my attempt at solving #1176. I'm not too crazy about the implementation (and I'm pretty sure it's not very idiomatic), but I thought it might be a point to start from.

I'm not sure why the diff looks so strange, btw..when I do a "diff -bd" the diff seem fairly straightforward. Perhaps my editor is doing something funky?

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -83,6 +83,9 @@ class FPM::Package::Deb < FPM::Package
 
   option "--group", "GROUP", "The group owner of files in this package", :default => 'root'
 
+  option "--no-changelog", :flag,
+    "Do not add a changelog by default for Debian packages."
+
   option "--changelog", "FILEPATH", "Add FILEPATH as debian changelog" do |file|
     File.expand_path(file)
   end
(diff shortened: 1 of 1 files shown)


lib/fpm/package/deb.rb
```diff
@@ -83,6 +83,9 @@ class FPM::Package::Deb < FPM::Package
 
   option "--group", "GROUP", "The group owner of files in this package", :default => 'root'
 
+  option "--no-changelog", :flag,
```
jordansissel: I prefer positive terms (remove the word 'no')

I think this should just be `--changelog` (a positive term) with `:default => true` and users can disable with `--no-deb-changelog`.

jordansissel:
I added a comment preferring positive term (removal of the 'no' word). The benefit is that this should make the flag be negatable with `--no-deb-changelog` instead of having users pass `--deb-no-changelog` (the `no` is my preference to be in front).

FooBarQuaxx:
I have a use case for this change. If the original author is not interested on making the requested changes, then I can volunteer to work on that.

lib/fpm/package/deb.rb
FooBarQuaxx: @jordansissel is this comment still valid since as `changelog` option is present already.

PR #1338, opened 2017-06-05
Be consistent in letting newer Perl module versions meet the requirement, if not otherwise specified

rjbell4:
Note that this is the same comparison a few lines later (line 179).  Without this change, I would never get all the Perl modules aligned perfectly.  As far as I can determine, this is what CPAN is enforcing, anyway.


jordansissel:
Can you add a comment above the line you changed to indicate your reasoning? (I see it in the PR description, but for future authors/readers it might be helpful to say this in the code comment as well)

rjbell4:
Sure.  Done.

jordansissel:
@rjbell4 awesome, thank you :)

PR #1339, opened 2017-06-05
Do not remove the leading "v" from version dependencies

rjbell4:
This is how the version of those dependencies themselves are being provided.  I ran into this with Kafka::Librd and Alien::Librdkafka.  Removing this code allowed the versioning to be correctly checked.

### DIFF
--- a/lib/fpm/package/cpan.rb
+++ b/lib/fpm/package/cpan.rb
@@ -316,9 +316,6 @@ class FPM::Package::CPAN < FPM::Package
       self.version = "#{cpan_version}"
     end
     
-    # remove 'v' prefix from version if it is there
-    self.version.sub!(/^v/, '')
-    
     # Search metacpan to get download URL for this version of the module
     metacpan_search_url = "http://api.metacpan.org/v0/release/_search"
     metacpan_search_query = '{"query":{"match_all":{}},"filter":{"term":{"release.name":"' + "#{distribution}-#{self.version}" + '"}}}'


jordansissel:
Interesting... This was added in 93dacaf0 (for #1239). It's hard to tell if this is working since this is still using the now-gone v0 api. 

Any tips for testing this?

rjbell4:
I should have `git blame`d it to learn more.  FYI, for anyone else reading, it's #1236.

It's interesting to think if this change is in part due to v1 of the API ... though I think it _may_ have been failing before v0 went away.  The downloading does seem to be working with the v1 API, at least, so if that's what the reason was, perhaps it no longer applies?

This is the module in question that was causing me problems: https://metacpan.org/pod/Alien::Librdkafka

(which I needed for Kafka::Librd)

rjbell4:
FWIW, the following command worked on Ubuntu 16.10:
`bin/fpm -s cpan -t deb Rex::Repositorio`
And the following worked on SLES12 SP1:
`bin/fpm -s cpan -t rpm Rex::Repositorio`

Since the problem was originally reported with Rex-Repositorio, I'm going to guess that this means that there is no regression with this change.

jordansissel:
@rjbell4 Thanks for the extra testing. I think this is good to merge.

liger1978:
#1236 Was mine. It was certainly needed at the time.

jordansissel:
@liger1978 any chance you can try master and see if it works for you? No pressure.

liger1978:
For converting `Beanstalk::Client` to rpm, I get:
```
/usr/local/share/gems/bundler/gems/fpm-32d526fb14cf/lib/fpm/package/cpan.rb:102:in `+': no implicit conversion of Float into String (TypeError)
	from /usr/local/share/gems/bundler/gems/fpm-32d526fb14cf/lib/fpm/package/cpan.rb:102:in `block in input'
	from /usr/local/share/gems/bundler/gems/fpm-32d526fb14cf/lib/fpm/package/cpan.rb:101:in `each'
	from /usr/local/share/gems/bundler/gems/fpm-32d526fb14cf/lib/fpm/package/cpan.rb:101:in `input'
	from build.rb:229:in `cpan2rpm'
	from build.rb:666:in `block in <main>'
	from build.rb:649:in `each'
	from build.rb:649:in `<main>'
```
Line 102 is here: https://github.com/jordansissel/fpm/blob/master/lib/fpm/package/cpan.rb#L102

I think error relates to the new capabilities functionality in #1340, rather than this issue though.

liger1978:
Changing `self.version` to `"#{self.version}"` in line 102 should fix it.

liger1978:
Opened #1343 for this.  Will attempt a fix.

PR #1340, opened 2017-06-05
Use capabilities in Perl dependencies

rjbell4:
This is a somewhat notable change to depend upon (and provide) Perl "capabilities" such as `perl(Capture::Tiny)`, instead of `perl-Capture-Tiny`.  Truthfully, I don't know how widespread this is, but it was absolutely _critical_ to get things running on my SLES system.  (Dealing with `Test2`, which presents itself as `Test-Simple`, which is also the name of a core module, was particulary problematic until I full switched over to dependencies)


jordansissel:
Seems like an OK change. I think this best belongs in the rpm `converted_from` but I am OK waiting until someone reports a problem.

Merging!

PR #1342, opened 2017-06-05
Translate Perl version dependencies

rjbell4:
Some CPAN modules have things like `use 5.008001`, which shows up in their dependencies.  However, that's not the version format used by the perl RPMs, which would be `5.8.1` (or newer).  This detects such situations and re-formats the version string.

### DIFF
--- a/lib/fpm/package/cpan.rb
+++ b/lib/fpm/package/cpan.rb
@@ -150,6 +150,10 @@ class FPM::Package::CPAN < FPM::Package
        found_dependencies.each do |dep_name, version|
           # Special case for representing perl core as a version.
           if dep_name == "perl"
+            m = version.match(/^(\d)\.(\d{3})(\d{3})$/)
+            if m
+               version = m[1] + '.' + m[2].sub(/^0*/, '') + '.' + m[3].sub(/^0*/, '')
+            end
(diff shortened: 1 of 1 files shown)


lib/fpm/package/cpan.rb
```diff
@@ -150,6 +150,10 @@ def input(package)
        found_dependencies.each do |dep_name, version|
           # Special case for representing perl core as a version.
           if dep_name == "perl"
+            m = version.match(/^(\d)\.(\d{3})(\d{3})$/)
+            if m
+               version = m[1] + '.' + m[2].sub(/^0*/, '') + '.' + m[3].sub(/^0*/, '')
```
jordansissel: will this work for all target packages? or just rpm?

If it focuses specific to rpm, then I think this can be added to the `converted_from` method in rpm.rb instead? Thoughts?

rjbell4: Well, I have a bit of ignorance here, but my Ubuntu machine has Perl `5.22.2` installed, not 5.022002`, and I would presume that a comparison of similar x.y.z versions would be better there, as well.

jordansissel: ehh. I'm uncertain but willing to try this.

PR #1354, opened 2017-06-20
spec: stud is runtime dependency

glensc:
```
$ grep -r require.*stud .
./lib/fpm/package.rb:require "stud/temporary"
./spec/fpm/package/dir_spec.rb:require "stud/temporary"
./spec/fpm/package/deb_spec.rb:require "stud/temporary"
./spec/fpm/package/pacman_spec.rb:require "stud/temporary"
./spec/fpm/package/rpm_spec.rb:require "stud/temporary" # gem 'stud'
./spec/fpm/util_spec.rb:require "stud/temporary"
./spec/fpm/command_spec.rb:require "stud/temporary"
```

introduced to `lib/fpm/package.rb` in 4547540 via #1139 for #1124

### DIFF
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -56,10 +56,11 @@ Gem::Specification.new do |spec|
   # For sourcing from pleaserun 
   spec.add_dependency("pleaserun", "~> 0.0.24") # license: Apache 2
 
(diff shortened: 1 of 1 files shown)


jordansissel:
👍

PR #1355, opened 2017-06-20
remove archive-tar-minitar as a dependency

dmmartins:
`archive-tar-minitar` is not used on `fpm` and version 0.5.2, pinned on current version of `fpm`, has [vulnerabilities](https://github.com/atoulme/minitar/issues/5).


jordansissel:
Fair point. I grepped around and it seems we use the rubygems (Gem::...) tar writer for tar stuff in some places, and in other places we shell out to call `tar` directly.

LGTM

PR #1356, opened 2017-06-22
If SOURCE_DATE_EPOCH is set, produce identical output every time

### LINKED ISSUE #1232: support reproducible packages
See the overview at https://reproducible-builds.org/ for this security effort.
Giving binary identical sources fpm produce binary different packages.

I guess the reason are different timestamps in generated control files.
Tested with -s dir and targets rpm, deb and osxpkg.

What I did in my source to fix this was this:
```
newest=`find $srcdir -print0 | xargs -0r ls -t | head -n1`
find $srcdir -depth -newer "$newest" -print0 | \
    xargs -0r touch --no-dereference -r "$newest"
```
debian does it similarily to touch their tar members.

So generated control files should always get a unique timestamp depending on the newest source, and not depending on the current time. The final package itself may have the current timestamp, just the members not.

locale-specific sort ordering while walking the source should also be changed to use the C locale.
same for locale specific timestamps.

dankegel:
Partly fixes #1232

Some work will be required in every package format handler;
for now, all that's been done is the minimum needed to achieve
reproducibility when running the canonical example:
```
   fpm -s gem -t deb json
```

See https://reproducible-builds.org/specs/source-date-epoch/
and https://wiki.debian.org/ReproducibleBuilds/StandardEnvironmentVariables
for why this is an environment variable, and not a commandline option.

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -233,6 +233,17 @@ class FPM::Command < Clamp::Command
     "copying, downloading, etc. Roughly any scratch space fpm needs to build " \
     "your package.", :default => Dir.tmpdir
 
+  option "--source-date-epoch", "SOURCE_DATE_EPOCH",
+    "Integer unix timestamp to use for generated files. " \
+    "See https://reproducible-builds.org/specs/source-date-epoch",
+    :environment_variable => "SOURCE_DATE_EPOCH"
+
+  option "--build-path-prefix-map", "BUILD_PATH_PREFIX_MAP",
+    "If SOURCE_DATE_EPOCH is given without BUILD_PATH_PREFIX_MAP, " \
+    "use a fixed package build directory to avoid nondeterminism. " \
+    "See https://reproducible-builds.org/specs/build-path-prefix-map",
+    :environment_variable => "BUILD_PATH_PREFIX_MAP"
+
   parameter "[ARGS] ...",
     "Inputs to the source package type. For the 'dir' type, this is the files" \
     " and directories you want to include in the package. For others, like " \
--- a/lib/fpm/package.rb
+++ b/lib/fpm/package.rb
@@ -1,6 +1,7 @@
 require "fpm/namespace" # local
 require "fpm/util" # local
 require "pathname" # stdlib
+require "fileutils"
 require "find"
 require "tmpdir" # stdlib
 require "backports" # gem 'backports'
@@ -142,6 +143,13 @@ class FPM::Package
       attributes[attribute] = value
     end
 
+    # Obey Reproducible Builds spec, see https://reproducible-builds.org/
+    # If set, individual package formats should set all timestamps to it
+    # and be extra-careful to avoid randomness in output.
+    if ENV.include?('SOURCE_DATE_EPOCH')
+      attributes[:source_date_epoch] = ENV['SOURCE_DATE_EPOCH']
+    end
+
     @name = nil
     @architecture = "native"
     @description = "no description given"
@@ -249,8 +257,20 @@ class FPM::Package
                                   "creating #{self.type} packages")
   end # def output
 
+  def temporary_directory(path)
+    if attributes[:source_date_epoch].nil?
+      return Stud::Temporary.directory("package-#{type}-staging")
+    else
+      # Choose a predictable build path; see https://reproducible-builds.org/docs/build-path/
+      root = ENV["TMP"] || ENV["TMPDIR"] || ENV["TEMP"] || "/tmp"
+      dir = File.join(root, 'fpm', 'reproducible-builds.org', attributes[:source_date_epoch], path)
+      FileUtils.mkdir_p(dir, :mode => 0700)
+      return dir
+    end
+  end
+
   def staging_path(path=nil)
-    @staging_path ||= Stud::Temporary.directory("package-#{type}-staging")
+    @staging_path ||= temporary_directory("package-#{type}-staging")
 
     if path.nil?
       return @staging_path
@@ -260,7 +280,7 @@ class FPM::Package
   end # def staging_path
 
   def build_path(path=nil)
-    @build_path ||= Stud::Temporary.directory("package-#{type}-build")
+    @build_path ||= temporary_directory("package-#{type}-build")
 
     if path.nil?
       return @build_path
--- a/lib/fpm/package.rb
+++ b/lib/fpm/package.rb
@@ -143,13 +143,6 @@ class FPM::Package
       attributes[attribute] = value
     end
 
-    # Obey Reproducible Builds spec, see https://reproducible-builds.org/
-    # If set, individual package formats should set all timestamps to it
-    # and be extra-careful to avoid randomness in output.
-    if ENV.include?('SOURCE_DATE_EPOCH')
-      attributes[:source_date_epoch] = ENV['SOURCE_DATE_EPOCH']
-    end
-
     @name = nil
     @architecture = "native"
     @description = "no description given"
@@ -258,10 +251,11 @@ class FPM::Package
   end # def output
 
   def temporary_directory(path)
-    if attributes[:source_date_epoch].nil?
-      return Stud::Temporary.directory("package-#{type}-staging")
+    if attributes[:source_date_epoch].nil? or not attributes[:build_path_prefix_map].nil?
+      # Normal, secure case
+      return Stud::Temporary.directory(path)
     else
-      # Choose a predictable build path; see https://reproducible-builds.org/docs/build-path/
+      # Remove this case once compilers implement https://wiki.debian.org/ReproducibleBuilds/BuildPathProposal
       root = ENV["TMP"] || ENV["TMPDIR"] || ENV["TEMP"] || "/tmp"
       dir = File.join(root, 'fpm', 'reproducible-builds.org', attributes[:source_date_epoch], path)
       FileUtils.mkdir_p(dir, :mode => 0700)
--- a/lib/fpm/package/gem.rb
+++ b/lib/fpm/package/gem.rb
@@ -239,7 +239,7 @@ class FPM::Package::Gem < FPM::Package
       # different output on successive runs.
       Find.find(installdir) do |path|
         if path =~ /.*(gem_make.out|Makefile)$/
-          logger.info("Removing %s" % path)
+          logger.info("Removing build file %s to reduce nondeterminism" % path)
           File.unlink(path)
         end
       end
--- a/lib/fpm/package/gem.rb
+++ b/lib/fpm/package/gem.rb
@@ -2,6 +2,7 @@ require "fpm/namespace"
 require "fpm/package"
 require "rubygems"
 require "fileutils"
+require "find"
 require "fpm/util"
 require "yaml"
 
@@ -229,6 +230,18 @@ class FPM::Package::Gem < FPM::Package
     if attributes[:gem_version_bins?] and File.directory?(bin_path)
       (::Dir.entries(bin_path) - ['.','..']).each do |bin|
         FileUtils.mv("#{bin_path}/#{bin}", "#{bin_path}/#{bin}-#{self.version}")
+      end
+    end
+
+    if not attributes[:source_date_epoch].nil?
+      # Remove generated Makefile and gem_make.out files, if any; they
+      # are not needed, and may contain generated paths that cause
+      # different output on successive runs.
+      Find.find(installdir) do |path|
+        if path =~ /.*(gem_make.out|Makefile)$/
+          logger.info("Removing %s" % path)
+          File.unlink(path)
+        end
       end
     end
   end # def install_to_staging
--- a/lib/fpm/util.rb
+++ b/lib/fpm/util.rb
@@ -277,7 +277,7 @@ module FPM::Util
     ["tar", "gtar", "gnutar"].each do |tar|
       opts=[]
       score=0
-      ["-sort=name", "--mtime=@0"].each do |opt|
+      ["--sort=name", "--mtime=@0"].each do |opt|
         system("#{tar} #{opt} -cf fpm-dummy.tar.tmp fpm-dummy.tmp > /dev/null 2>&1")
         if $?.exitstatus == 0
           puts("tar_cmd: #{tar} #{opt} succeeded")
--- a/lib/fpm/util.rb
+++ b/lib/fpm/util.rb
@@ -290,11 +290,9 @@ module FPM::Util
       ["--sort=name", "--mtime=@0"].each do |opt|
         system("#{tar} #{opt} -cf fpm-dummy.tar.tmp fpm-dummy.tmp > /dev/null 2>&1")
         if $?.exitstatus == 0
-          puts("tar_cmd: #{tar} #{opt} succeeded")
           opts << opt
           score += 1
         end
-        puts("tar_cmd: #{tar} #{opt} failed")
       end
       if score > bestscore
         best=tar
--- a/lib/fpm/util.rb
+++ b/lib/fpm/util.rb
@@ -238,6 +238,8 @@ module FPM::Util
   def ar_cmd
     return @@ar_cmd if defined? @@ar_cmd
 
+    @@ar_cmd_deterministic = FALSE
+
     # FIXME: don't assume current directory writeable
     FileUtils.touch(["fpm-dummy.tmp"])
     ["ar", "gar"].each do |ar|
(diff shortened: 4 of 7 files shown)


templates/deb/changelog.erb
```diff
@@ -2,4 +2,4 @@
 
   * Package created with FPM.
 
- -- <%= maintainer %>  <%= Time.now.strftime("%a, %d %b %Y %T %z") %>
+ -- <%= maintainer %>  <%= (if attributes[:source_date_epoch].nil? then Time.now() else Time.at(attributes[:source_date_epoch].to_i) end).strftime("%a, %d %b %Y %T %z") %>
```
lamby: ``"%a`` is locale-specific.

lib/fpm/package.rb
```diff
@@ -249,8 +257,20 @@ def output(path)
                                   "creating #{self.type} packages")
   end # def output
 
+  def temporary_directory(path)
+    if attributes[:source_date_epoch].nil?
+      return Stud::Temporary.directory("package-#{type}-staging")
+    else
+      # Choose a predictable build path; see https://reproducible-builds.org/docs/build-path/
+      root = ENV["TMP"] || ENV["TMPDIR"] || ENV["TEMP"] || "/tmp"
```
lamby: This seems somewhat regrettable (ie. it's not actually "predictable" as the comment claims IMHO) but I can see how there is no real workaround for fpm as it cannot, for example, guarantee the availability of, say, ``/tmp/buildd/whatever``.

lib/fpm/package.rb
```diff
@@ -249,8 +257,20 @@ def output(path)
...
+      dir = File.join(root, 'fpm', 'reproducible-builds.org', attributes[:source_date_epoch], path)
+      FileUtils.mkdir_p(dir, :mode => 0700)
```
lamby: I think this is an insecure use of a tempdir?

lib/fpm/package/gem.rb
```diff
@@ -231,6 +232,18 @@ def install_to_staging(gem_path)
...
       end
     end
+
+    if not attributes[:source_date_epoch].nil?
+      # Remove generated Makefile and gem_make.out files, if any; they
+      # are not needed, and may contain generated paths that cause
+      # different output on successive runs.
+      Find.find(installdir) do |path|
+        if path =~ /.*(gem_make.out|Makefile)$/
+          logger.info("Removing %s" % path)
```
wasosa: How about also logging the reason for removing these files?

wasosa:
Nice getting the ball rolling on reproducible builds here.

lib/fpm/package.rb
dankegel: That's why I asked for review on the mailing list.
What else can I do?  Without this, diffoscope reports a plethora of nasty differences
on 'fpm -s gem -t deb json', e.g.
```
├── data.tar.gz
│   ├── data.tar
│   │   ├── ./var/lib/gems/2.3.0/extensions/x86_64-linux/2.3.0/json-2.1.0/json/ext/generator.so
│   │   │   ├── readelf --wide --notes {}
│   │   │   │ @@ -1,5 +1,5 @@
│   │   │   │  
│   │   │   │  Displaying notes found at file offset 0x000001c8 with length 0x00000024:
│   │   │   │    Owner                 Data size        Description
│   │   │   │    GNU                  0x00000014        NT_GNU_BUILD_ID (unique build ID bitstring)
│   │   │   │ -    Build ID: 438800c00798505fc92d9302ebf4ecfcd5aa5f57
│   │   │   │ +    Build ID: 6364fd6a8cad33df79bd85b47faa3617e13304fe
│   │   │   ├── readelf --wide --debug-dump=info {}
│   │   │   │ @@ -2,60 +2,60 @@
│   │   │   │  
│   │   │   │    Compilation Unit @ offset 0x0:
│   │   │   │     Length:        0x7605 (32-bit)
│   │   │   │     Version:       4
│   │   │   │     Abbrev Offset: 0x0
│   │   │   │     Pointer Size:  8
│   │   │   │   <0><b>: Abbrev Number: 1 (DW_TAG_compile_unit)
│   │   │   │ -    <c>   DW_AT_producer    : (indirect string, offset: 0x361): GNU C11 5.4.0 20160609 -mtune=generic -march=x86-64 -g -O2 -fstack-protector-strong -fPIC -fstack-protector-strong
│   │   │   │ +    <c>   DW_AT_producer    : (indirect string, offset: 0x2d1): GNU C11 5.4.0 20160609 -mtune=generic -march=x86-64 -g -O2 -fstack-protector-strong -fPIC -fstack-protector-strong
│   │   │   │      <10>   DW_AT_language    : 12        (ANSI C99)
│   │   │   │      <11>   DW_AT_name        : (indirect string, offset: 0x1147): generator.c
│   │   │   │ -    <15>   DW_AT_comp_dir    : (indirect string, offset: 0xae): /tmp/package-gem-staging-7f3a2a248718996b858ad69aa865ca5dc46125339c4d6f163bedf10f0b24/var/lib/gems/2.3.0/gems/json-2.1.0/ext/json/ext/generator
│   │   │   │ +    <15>   DW_AT_comp_dir    : (indirect string, offset: 0xe4a): /tmp/package-gem-staging-6d7699fb2e5ccddffbb0ca5b66855e3971f9db4da3984f8ef7845da10088/var/lib/gems/2.3.0/gems/json-2.1.0/ext/json/ext/generator
```

lib/fpm/package/gem.rb
dankegel: Good idea.

lib/fpm/package.rb
jordansissel: fpm already has a setting for this: `--workdir WORKDIR`

We can add an environment variable if you believe you need one, but `--workdir` is intended for this purpose, at least, as far as I can tell.

templates/deb/changelog.erb
dankegel: %a ceased being locale-specific in ruby 1.9....? 
https://idiosyncratic-ruby.com/57-what-the-time.html says
```
"%a | Weekday Name (Abbreviated)
Locale-independent (English) three-letter name of weekday"
```

Verified:
```
$ export DEBEMAIL=dank@kegel.com
$ export DEBFULLNAME="Dan Kegel"
$ export SOURCE_DATE_EPOCH=123
$ LANG=de_DE.UTF-8
$ date
Do 22. Jun 15:42:00 PDT 2017
$ ruby -Ilib bin/fpm --verbose -s gem -t deb json
$ dpkg-deb -x *.deb tmp
$ zmore tmp/usr/share/doc/rubygem-json/changelog.gz 
rubygem-json (2.1.0) whatever; urgency=medium

  * Package created with FPM.

 -- Dan Kegel <dank@kegel.com>  Wed, 31 Dec 1969 16:02:03 -0800
```

lib/fpm/package.rb
dankegel: --workdir, alas, only sets TMP; it doesn't prevent the nondeterministic suffix.

jordansissel: Can we make --workdir do what you need, then?

lib/fpm/package.rb
dankegel: I 'solved' this by also sensing BUILD_PATH_PREFIX_MAP and using the old secure path if it's set.

lib/fpm/package.rb
jordansissel: I may not be understanding this scenario, btw. Temporary directories, work directories, etc, are only need to exist during `fpm`'s run -- when `fpm` exits, they should be automatically removed.

Can you help me understand what about temporary paths is important to this 'repeatable build' goal?

dankegel:
I think all the issues raised during review have been resolved.  
On Ubuntu 17.04, rspec is passing as well as it did before my changes
(filed https://github.com/jordansissel/fpm/issues/1357 for the one pre-existing failure).

lib/fpm/package.rb
```diff
@@ -142,6 +143,13 @@ def initialize
       attributes[attribute] = value
     end
 
+    # Obey Reproducible Builds spec, see https://reproducible-builds.org/
+    # If set, individual package formats should set all timestamps to it
+    # and be extra-careful to avoid randomness in output.
+    if ENV.include?('SOURCE_DATE_EPOCH')
```
dankegel: Done.

lib/fpm/package.rb
dankegel: If you run the command "fpm -s gem -t deb json" twice in a row, it generates two different debs.
In the reproducible builds use case, we want it to generate two identical debs, regardless of who
runs the build when or where, as long as the OS is the same release / version / architecture. 
(Otherwise e.g. clicking 'Force Rebuild' in buildbot generates failed uploads because reprepro rejects
the new deb with the same version and different checksum.)

One particular problem is that the C toolchain often embeds the absolute path to the build
directory in the binary, say, in the DW_AT_comp_dir field.
Until we teach the compiler to obey BUILD_PATH_PREFIX_MAP,
we can work around this by using a deterministic build path
( see https://reproducible-builds.org/docs/build-path/ ).

So, this has nothing to do with leaving a clean build machine after running fpm, but everything to do with bit-for-bit identical output packages given bit-for-bit identical input packages.

lib/fpm/package.rb
dankegel: We can remove temporary_directory() and --build-path-prefix-map,
and go back to the original code once BUILD_PATH_PREFIX_MAP is implemented by old compilers.

dankegel:
Travis is unhappy with this PR because it's running on Ubuntu 12.04.
The --sort option wasn't added to gnu tar until circa Ubuntu 16.04.

I guess the test should skip if tar doesn't support --sort ...

lib/fpm/package.rb
jordansissel: Ahh, that helps a bit. Are there known casess with your `fpm -s gem -t deb json` example where temporary directories (by file content or file path) are showing up somehow in the resulting .deb?

dankegel: Yes, the example from the next hunk down shows it.  Here it is with more context:
```
$ fpm -s gem -t deb json; mkdir run1; mv *.deb run1
$ sleep 1
$ fpm -s gem -t deb json; mkdir run1; mv *.deb run2
$ diffoscope run1/rubygem-run2/json_2.1.0_amd64.debrubygem-json_2.1.0_amd64.deb 
...
├── data.tar.gz
│   ├── data.tar
│   │   ├── ./var/lib/gems/2.3.0/extensions/x86_64-linux/2.3.0/json-2.1.0/json/ext/generator.so
│   │   │   ├── readelf --wide --notes {}
│   │   │   │ @@ -1,5 +1,5 @@
│   │   │   │  
│   │   │   │  Displaying notes found at file offset 0x000001c8 with length 0x00000024:
│   │   │   │    Owner                 Data size        Description
│   │   │   │    GNU                  0x00000014        NT_GNU_BUILD_ID (unique build ID bitstring)
│   │   │   │ -    Build ID: 438800c00798505fc92d9302ebf4ecfcd5aa5f57
│   │   │   │ +    Build ID: 6364fd6a8cad33df79bd85b47faa3617e13304fe
│   │   │   ├── readelf --wide --debug-dump=info {}
│   │   │   │ @@ -2,60 +2,60 @@
│   │   │   │  
│   │   │   │    Compilation Unit @ offset 0x0:
│   │   │   │     Length:        0x7605 (32-bit)
│   │   │   │     Version:       4
│   │   │   │     Abbrev Offset: 0x0
│   │   │   │     Pointer Size:  8
│   │   │   │   <0><b>: Abbrev Number: 1 (DW_TAG_compile_unit)
│   │   │   │ -    <c>   DW_AT_producer    : (indirect string, offset: 0x361): GNU C11 5.4.0 20160609 -mtune=generic -march=x86-64 -g -O2 -fstack-protector-strong -fPIC -fstack-protector-strong
│   │   │   │ +    <c>   DW_AT_producer    : (indirect string, offset: 0x2d1): GNU C11 5.4.0 20160609 -mtune=generic -march=x86-64 -g -O2 -fstack-protector-strong -fPIC -fstack-protector-strong
│   │   │   │      <10>   DW_AT_language    : 12        (ANSI C99)
│   │   │   │      <11>   DW_AT_name        : (indirect string, offset: 0x1147): generator.c
│   │   │   │ -    <15>   DW_AT_comp_dir    : (indirect string, offset: 0xae): /tmp/package-gem-staging-7f3a2a248718996b858ad69aa865ca5dc46125339c4d6f163bedf10f0b24/var/lib/gems/2.3.0/gems/json-2.1.0/ext/json/ext/generator
│   │   │   │ +    <15>   DW_AT_comp_dir    : (indirect string, offset: 0xe4a): /tmp/package-gem-staging-6d7699fb2e5ccddffbb0ca5b66855e3971f9db4da3984f8ef7845da10088/var/lib/gems/2.3.0/gems/json-2.1.0/ext/json/ext/generator
```

spec/fpm/package/deb_spec.rb
```diff
@@ -358,13 +358,23 @@ def dpkg_field(field)
...
       # Check prerequisites
-      puts("Using " + `which #{tar_cmd}` + `#{tar_cmd} --version`)
-      puts("Exit status of '#{tar_cmd} -cf /dev/null --mtime=@0 --sort=name /dev/null' is #{$?.exitstatus}")
+      # FIXME: ar_cmd and tar_cmd should tell us this
+      puts("Using ar " + `which #{ar_cmd[0]}` + `#{ar_cmd[0]} --version`)
+      FileUtils.rm_f("foo.ar")
+      system("#{ar_cmd.join(' ')} foo.ar /dev/null && env LC_TIME=C TZ=GMT ar tv foo.ar | grep 1970 > /dev/null")
+      FileUtils.rm_f("foo.ar")
+      if $?.exitstatus != 0
+        skip("This system doesn't seem to have an ar that can omit timestamps")
```
jordansissel: Seems like this should be in FPM, not the tests. Otherwise a user could request a repeatable-build and think fpm is doing the right thing but the `ar` isn't capable and we end up lying to the user, right?

lib/fpm/command.rb
```diff
@@ -233,6 +233,17 @@ def help(*args)
     "copying, downloading, etc. Roughly any scratch space fpm needs to build " \
     "your package.", :default => Dir.tmpdir
 
+  option "--source-date-epoch", "SOURCE_DATE_EPOCH",
+    "Integer unix timestamp to use for generated files. " \
+    "See https://reproducible-builds.org/specs/source-date-epoch",
+    :environment_variable => "SOURCE_DATE_EPOCH"
+
+  option "--build-path-prefix-map", "BUILD_PATH_PREFIX_MAP",
+    "If SOURCE_DATE_EPOCH is given without BUILD_PATH_PREFIX_MAP, " \
```
jordansissel: This description seems to be about what happens if you do not provide this flag, but not what happens if you do provide it.

spec/fpm/package/deb_spec.rb
dankegel: It is in fpm, in the ar_cmd and tar_cmd functions.  Unfortunately, those don't pass back the boolean.  I should refactor that so I can avoid duplicating the test.

lib/fpm/util.rb
```diff
@@ -226,24 +232,73 @@ def safesystemout(*args)
     return stdout_r_str
   end # def safesystemout
 
+  # Get an array containing the recommended 'ar' command for this platform
+  # and the recommended options to quickly create/append to an archive
+  # without timestamps or uids (if possible).
+  def ar_cmd
```
jordansissel: This seems to default to calling `ar` in a way that doesn't use timestamps or uids. This is a behavior change,but perhaps not a problem.

However, if the user is asking for repeatable builds (with whatever flags) on a system who's `ar` is not capable of doing this, then this `ar_cmd` will fall back to one that is not repeatable.

My main goal; If a user is asking for deterministic builds, and the system cannot produce it, we should abort and error. Looking at this `ar_cmd` it is unclear if that is what is happening.

lib/fpm/command.rb
dankegel: That's because I wrote it before I figured out what it'd do :-)   I'll add that soon.  Bit hard to test without a compiler that supports it, so maybe I'll have to build one.

jordansissel: Ahh, makes sense. Is this purely about compilers? Which compiler(s)? Would it be useful to instruct users on which software this affected? (gcc, for example, but not javac perhaps?)

lib/fpm/util.rb
dankegel: Done.

dankegel:
I have yet to implement true support for BUILD_PATH_PREFIX_MAP (which, I think, involves appending our work directory to that environment variable).  Once I do that, and folks are happy, I'll squash and submit a clean MR.

lib/fpm/command.rb
dankegel: The set of folks who need to know are small, and anyone who invokes this feature will probably know to consult the doc at https://reproducible-builds.org/ in case of trouble.
This tool itself can't keep track of the state of play in which other tools support reproducible builds, and the doc already links to that site; adding much more detail probably wouldn't be worth it.

dankegel:
Closing in favor of https://github.com/jordansissel/fpm/pull/1360

PR #1358, opened 2017-06-25
Performance optimization for apk packages

jandelgado:
This patch dramatically improves performance when building larger alpine apk packages.


jordansissel:
I haven't tested, but this change looks fine to me.

For posterity and any curious readers, the reason this improves performance is the previous operation `+=` concatenates two strings and stores the result -- this creates a new string every time, making the previous value something needing to be garbage collected. The `<<` operator appends directly to the string without wasting much memory.

Issue #1225, opened 2016-11-09
[deb] support activate-noawait and interest-noawait


philtay:
Currently `--deb-activate` and `--deb-interest` options allow to specify `activate` and `interest` triggers only. Please add `--deb-activate-noawait` and `--deb-interest-noawait` options to support `activate-noawait` and `interest-noawait` triggers too. More info on deb-triggers [here](http://man7.org/linux/man-pages/man5/deb-triggers.5.html). It looks like a trivial PR, but I don't speak Ruby.

nthykier:
Very much a good idea.  The average user of triggers generally want the "-noawait" variant and notably it avoids the entire problem of "trigger-cycles".  I would go so far to as to recommend people to use the noawait triggers by default in the documentation.

philtay:
ping @jordansissel

jordansissel:
The link to more information on triggers looks like a lot of nice details.
Thanks!

I personally don't use triggers, so I don't have much self interest in this
feature. If someone comes up with a proposal, I am open to patches for this.

ppoilbarbe:
I have done it... It's not quite difficult even if I do not understand all the magic under the hood of Ruby (I did just mimic what have been done for --deb-interest and --deb-activate).
Do you prefer a patch or a merge request ?

jordansissel:
#1359 merged

PR #1359, opened 2017-06-26
Added --deb-interest-noawait and --deb-activate-noawait (#1225)

ppoilbarbe:
Added these two options to allow the debian functionalities associated with them


jordansissel:
I read a little bit about debian triggers and this code seems to look like the right thing. Beyond that I am unsure how to evaluate it.

Given it's an opt-in flag users have to set, I'm OK merging this as-is. Code LGTM.

PR #1360, opened 2017-06-27
Implement reproducible gem -> deb conversion

dankegel:
Partial fix for https://github.com/jordansissel/fpm/issues/1232
Ready for review and merging.

It fetches the source epoch from the changelog for about a third of the gems I tried.

Native gems still have some randomness on Darwin, but I'm not going to worry about that;
docs/source/gem.rst mentions that it's only been verified to produce identical results on Linux.

The following test passes locally on ubuntu 16.04:
```
 mkdir run1
 ruby -Ilib bin/fpm -s gem -t deb --gem-stagingdir /tmp/gem1.tmp --source-date-epoch-from-changelog json
 mv *.deb run1

 mkdir run2
 ruby -Ilib bin/fpm -s gem -t deb --gem-stagingdir /tmp/gem1.tmp --source-date-epoch-from-changelog json
 mv *.deb run2

 cmp run1/*.deb run2/*.deb
```

### DIFF
--- a/lib/fpm/package/gem.rb
+++ b/lib/fpm/package/gem.rb
@@ -275,33 +275,44 @@ class FPM::Package::Gem < FPM::Package
   # Supports formats suggested by http://keepachangelog.com and https://github.com/tech-angels/vandamme
   # as well as other similar formats that actually occur in the wild.
   # Build it in pieces for readability, and allow version and date in either order.
+  # Whenever you change this, add a row to the test case in spec/fpm/package/gem_spec.rb.
+  # Don't even try to handle dates that lack four-digit years.
   # Building blocks:
   P_RE_LEADIN    = '^[#=]{0,3}\s?'
   P_RE_VERSION_  = '[\w\.-]+\.[\w\.-]+[a-zA-Z0-9]'
   P_RE_SEPARATOR = '\s[-=/(]?\s?'
-  P_RE_DATE      = '(?<date>\w+ \d{1,2}(?:st|nd|rd|th)?,\s\d{4}|\d{4}-\d{2}-\d{2})'
-  P_RE_VERSION      = "\\[?(?:Version |v)?(?<version>#{P_RE_VERSION_})\\]?"
+  P_RE_DATE1     = '\d{4}-\d{2}-\d{2}'
+  P_RE_DATE2     = '\w+ \d{1,2}(?:st|nd|rd|th)?,\s\d{4}'
+  P_RE_DATE3     = '\w+\s+\w+\s+\d{1,2},\s\d{4}'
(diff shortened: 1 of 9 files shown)


dankegel:
Pfft.  Also found a mac problem locally. will upload fixes once it's happy.

lib/fpm/package/gem.rb
```diff
@@ -47,6 +47,22 @@ class FPM::Package::Gem < FPM::Package
 
   option "--version-bins", :flag, "Append the version to the bins", :default => false
 
+  option "--stagingdir", "STAGINGDIR",
+    "The directory where fpm installs the gem temporarily before conversion. " \
+    "Normally a random subdirectory of workdir."
+
+  # Override parent method
+  def staging_path(path=nil)
+    @gem_staging_path ||= attributes[:gem_stagingdir] || Stud::Temporary.directory("package-#{type}-staging")
```
jordansissel: Long-term I would prefer a simpler user experience for deterministic builds (perhaps a single flag `--deterministic-build`) that would make these kinds of decisions for the user. For now, this is good :)

lib/fpm/util.rb
```diff
@@ -226,26 +232,90 @@ def safesystemout(*args)
     return stdout_r_str
   end # def safesystemout
 
+  # Get an array containing the recommended 'ar' command for this platform
+  # and the recommended options to quickly create/append to an archive
+  # without timestamps or uids (if possible).
+  def ar_cmd
+    return @@ar_cmd if defined? @@ar_cmd
+
+    @@ar_cmd_deterministic = FALSE
```
jordansissel: TIL Ruby provides a constant `FALSE` that means `false`. This should probably be lowercase `false` (for style) but there is no negative impact to the user, so it's cool.

jordansissel: Ruby 2.4 warns about this constant, so I'll remove it after merging:

```
../home/jls/projects/fpm/lib/fpm/util.rb:284: warning: constant ::FALSE is deprecated
/home/jls/projects/fpm/lib/fpm/util.rb:301: warning: constant ::TRUE is deprecated
/home/jls/projects/fpm/lib/fpm/util.rb:241: warning: constant ::FALSE is deprecated
/home/jls/projects/fpm/lib/fpm/util.rb:255: warning: constant ::TRUE is deprecated
```

jordansissel:
Did a light review and ran the test suite. Tests passing and overall the code is good. Thank you for writing docs :)

lib/fpm/package/gem.rb
dankegel: Well, yeah... but long-term this option won't be needed (once compilers obey https://reproducible-builds.org/specs/build-path-prefix-map/ ), so if we're going to have an intermediate workaround, let's make it simple, transparent, and scary so nobody uses it by accident.

PR #1364, opened 2017-06-28
Add --deb-after-purge FILE

alxwr:
Makes Debian's `postrm purge` hook usable.

Adds a new CLI argument `--deb-after-remove-purge` and extends the already prepared (thanks a lot!) `postrm_upgrade.sh.erb` template.

Thanks for `fpm`!

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -18,6 +18,7 @@ class FPM::Package::Deb < FPM::Package
     :after_install      => "postinst",
     :before_remove      => "prerm",
     :after_remove       => "postrm",
+    :after_purge        => "postrm",
   } unless defined?(SCRIPT_MAP)
 
   # The list of supported compression types. Default is gz (gzip)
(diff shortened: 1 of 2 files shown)


alebsack:
Would love to see this. +1

lib/fpm/package/deb.rb
```diff
@@ -18,6 +18,7 @@ class FPM::Package::Deb < FPM::Package
     :after_install      => "postinst",
     :before_remove      => "prerm",
     :after_remove       => "postrm",
+    :after_remove_purge => "postrm",
```
jordansissel: `apt` calls this 'purge' not "remove purge" -- thoughts on just having this called `after_purge` ?

alxwr: I named it that way because `purge` is a special case of `remove`, which happens _after_ remove. "after remove: purge" (`purge` is even a part of the `postrm` script.) 

That said, I've got no problem renaming the CLI parameter and the variables. :-) I'm OK with `after_purge`, but `on_purge` would be more accurate IMHO.

Please let me known which variant you prefer, and I'll edit the PR.

alxwr:
I went ahead and changed it to `--after-purge` / `after_purge`. Please let me know if the current state of the PR is OK for you. :-)

BTW: Thank you very much for sharing `fpm`!

jordansissel:
Code looks good. Thank you! :)

PR #1376, opened 2017-07-21
Add --python-setup-py-arguments flag which can be used to pass arbitrary arguments to setup.py.

josephholsten:
updated #1120

### DIFF
--- a/lib/fpm/package/python.rb
+++ b/lib/fpm/package/python.rb
@@ -73,6 +73,10 @@ class FPM::Package::Python < FPM::Package
     "The python package name to remove from dependency list",
     :multivalued => true, :attribute_name => :python_disable_dependency,
     :default => []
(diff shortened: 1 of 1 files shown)


lib/fpm/package/python.rb
```diff
@@ -310,6 +314,13 @@ def install_to_staging(setup_py)
         flags += [ "build_scripts", "--executable", attributes[:python_scripts_executable] ]
       end
 
+      if !attributes[:python_setup_py_arguments].nil? and !attributes[:python_setup_py_arguments].empty?
```
jordansissel: this won't ever be nil because the default is `[]` (empty array), but there's probably no harm in this nil check.

jordansissel:
I reviewed #1120 and this PR. Seems safe to merge.

PR #1378, opened 2017-07-21
Local project path support for virtualenv target. (re:#947)

josephholsten:
rebased #947, I'm skeptical this actually works.

*Note*: apparently I rebased this badly.

It would also be awesome to have @vi4m review this, I don't really know how all these flags interact.

### DIFF
--- a/lib/fpm/package/virtualenv.rb
+++ b/lib/fpm/package/virtualenv.rb
@@ -15,6 +15,10 @@ class FPM::Package::Virtualenv < FPM::Package
   option "--package-name-prefix", "PREFIX", "Name to prefix the package " \
   "name with.", :default => "virtualenv"
 
+  option "--local-project-path", "DIRECTORY",
+  "Local python project path",
+  :default => ""
+
   option "--install-location", "DIRECTORY", "DEPRECATED: Use --prefix instead." \
     "  Location to which to install the virtualenv by default.",
     :default => "/usr/share/python" do |path|
@@ -135,7 +139,15 @@ class FPM::Package::Virtualenv < FPM::Package
       target_args << package
     end
 
-    pip_args = [python_exe, pip_exe, "install", "-i", attributes[:virtualenv_pypi]] << extra_index_url_args << find_links_url_args << target_args
+    # Installing from local project path has an advantage to now having to upload
+    # python package to pypi server just to make distro package.
+    if attributes[:virtualenv_local_project_path]
+      logger.debug("Installing from local project path.")
+      pip_args = [python_exe, pip_exe, "install", "-e", "file://" + attributes[:virtualenv_local_project_path]] << extra_index_url_args << find_links_url_args
+    else
+      logger.debug("Installing from PyPi server.")
+      pip_args = [python_exe, pip_exe, "install", "-i", attributes[:virtualenv_pypi]] << extra_index_url_args << find_links_url_args << target_args
+    end
     safesystem(*pip_args.flatten)
 
     if attributes[:virtualenv_setup_install?]
--- a/lib/fpm/package/virtualenv.rb
+++ b/lib/fpm/package/virtualenv.rb
@@ -143,7 +143,7 @@ class FPM::Package::Virtualenv < FPM::Package
     # python package to pypi server just to make distro package.
     if attributes[:virtualenv_local_project_path]
       logger.debug("Installing from local project path.")
-      pip_args = [python_exe, pip_exe, "install", "-e", "file://" + attributes[:virtualenv_local_project_path]] << extra_index_url_args << find_links_url_args
+      pip_args = [python_exe, pip_exe, "install", "file://" + attributes[:virtualenv_local_project_path]] << extra_index_url_args << find_links_url_args
     else
       logger.debug("Installing from PyPi server.")
       pip_args = [python_exe, pip_exe, "install", "-i", attributes[:virtualenv_pypi]] << extra_index_url_args << find_links_url_args << target_args
--- a/lib/fpm/package/virtualenv.rb
+++ b/lib/fpm/package/virtualenv.rb
@@ -143,7 +143,7 @@ class FPM::Package::Virtualenv < FPM::Package
     # python package to pypi server just to make distro package.
     if attributes[:virtualenv_local_project_path]
       logger.debug("Installing from local project path.")
-      pip_args = [python_exe, pip_exe, "install", "file://" + attributes[:virtualenv_local_project_path]] << extra_index_url_args << find_links_url_args
+      pip_args = [python_exe, pip_exe, "install", attributes[:virtualenv_local_project_path]] << extra_index_url_args << find_links_url_args
     else
       logger.debug("Installing from PyPi server.")
       pip_args = [python_exe, pip_exe, "install", "-i", attributes[:virtualenv_pypi]] << extra_index_url_args << find_links_url_args << target_args


lib/fpm/package/virtualenv.rb
```diff
@@ -135,7 +139,15 @@ def input(package)
       target_args << package
     end
 
-    pip_args = [python_exe, pip_exe, "install", "-i", attributes[:virtualenv_pypi]] << extra_index_url_args << find_links_url_args << target_args
+    # Installing from local project path has an advantage to now having to upload
+    # python package to pypi server just to make distro package.
+    if attributes[:virtualenv_local_project_path]
```
josephholsten: default is `""`, which is true. should check for emptiness instead, or change default.

josephholsten:
needs a test

lib/fpm/package/virtualenv.rb
jordansissel: you could remove the `:default => ""` which would imply the default is nil, making this check `if ...` valid because `nil` is falsey in Ruby.

Thoughts?

Either a falsey check (with no default (nil)) or `.empty?` is fine with me.

jordansissel:
Assuming I'm understanding this correctly, I have a proposal (tell me if I am off-base)

This PR adds a new flag to enable tell fpm a local project path, but what if instead of a new flag, we do what the `python` and `gem` support does in fpm which is to check if the given argument (`target_args` in this scope?) is a local path. If it is a local path, we use that the same way as intended by this PR. If it is not a local path, then we use the existing behavior?

The end result would be nearly the same for the user with the exception that we wouldn't need to add a flag to achieve it. Thoughts?

josephholsten:
is it possible for a file check to be ambiguous in that case? there being a `./foo` directory and a `foo` package? is that a concern?

jordansissel:
I think it's possible, but maybe unlikely.

gem source does this: https://github.com/jordansissel/fpm/blob/master/lib/fpm/package/gem.rb#L80-L88

* If the given argument is a file that exists, use that file. If it does not exist as a file, ask rubygems for the gem by name.

python source does this: https://github.com/jordansissel/fpm/blob/master/lib/fpm/package/python.rb#L110-L149

* If the given argument is a directory, assume `<arg>/setup.py`. If the argument is a file named `setup.py`, use that path. Otherwise, ask pypi for the package named `<arg>`.

From what I can remember, both of the above (gem and python) strategies have served users well without being confusing. I might be remembering incorrectly, though.

That said, I think it's probably OK to take a similar strategy here where we use the argument as a path if that path looks like a reasonable thing to tell virtualenv about, and if it isn't, then we fetch things from pypi directly.

Thoughts?

josephholsten:
+🌮

josephholsten:
current failing tests:
```
  1) FPM::Package::Nuget #fix_dependency with normal versions should accept 'MyDependency'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:23:in `block (4 levels) in <top (required)>'

  2) FPM::Package::Nuget #fix_dependency with normal versions should accept 'MyDependency 1'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:29:in `block (4 levels) in <top (required)>'

  3) FPM::Package::Nuget #fix_dependency with normal versions should accept 'MyDependency 1.2'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:33:in `block (4 levels) in <top (required)>'

  4) FPM::Package::Nuget #fix_dependency with normal versions should accept 'MyDependency 1.2.3'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:37:in `block (4 levels) in <top (required)>'

  5) FPM::Package::Nuget #fix_dependency with normal versions should accept 'MyDependency [1.2.3]'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:41:in `block (4 levels) in <top (required)>'

  6) FPM::Package::Nuget #fix_dependency with normal versions should fail on 'MyDependency (1.2.3)'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:45:in `block (4 levels) in <top (required)>'

  7) FPM::Package::Nuget #fix_dependency with normal versions should accept 'MyDependency (,1.2]'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:49:in `block (4 levels) in <top (required)>'

  8) FPM::Package::Nuget #fix_dependency with normal versions should accept 'MyDependency (,1.2.3)'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:53:in `block (4 levels) in <top (required)>'

  9) FPM::Package::Nuget #fix_dependency with normal versions should accept 'MyDependency (1.2,)'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:57:in `block (4 levels) in <top (required)>'

  10) FPM::Package::Nuget #fix_dependency with normal versions should accept 'MyDependency (1.2,2.0)'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:61:in `block (4 levels) in <top (required)>'

  11) FPM::Package::Nuget #fix_dependency with normal versions should accept 'MyDependency [1.2,2.0]'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:65:in `block (4 levels) in <top (required)>'

  12) FPM::Package::Nuget #fix_dependency with allowed versions should accept 'MyDependency 1 2'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:72:in `block (4 levels) in <top (required)>'

  13) FPM::Package::Nuget #fix_dependency with allowed versions should accept 'MyDependency 1.2 (,2.0)'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:76:in `block (4 levels) in <top (required)>'

  14) FPM::Package::Nuget #fix_dependency with allowed versions should accept 'MyDependency 1.2.3' (,2.0)

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:80:in `block (4 levels) in <top (required)>'

  15) FPM::Package::Nuget #fix_dependency with allowed versions should accept 'MyDependency [1.2.3]' (,2.0)

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:84:in `block (4 levels) in <top (required)>'

  16) FPM::Package::Nuget #fix_dependency with allowed versions should fail on 'MyDependency (1.2.3) (,2.0)'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:88:in `block (4 levels) in <top (required)>'

  17) FPM::Package::Nuget #fix_dependency with allowed versions should accept 'MyDependency (,1.2] (,2.0)'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:92:in `block (4 levels) in <top (required)>'

  18) FPM::Package::Nuget #fix_dependency with allowed versions should accept 'MyDependency (,1.2.3) (,2.0)'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:96:in `block (4 levels) in <top (required)>'

  19) FPM::Package::Nuget #fix_dependency with allowed versions should accept 'MyDependency (1.2,) (,2.0)'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:100:in `block (4 levels) in <top (required)>'

  20) FPM::Package::Nuget #fix_dependency with allowed versions should accept 'MyDependency (1.2,2.0) (,3.0)'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:104:in `block (4 levels) in <top (required)>'

  21) FPM::Package::Nuget #fix_dependency with allowed versions should accept 'MyDependency [1.2,2.0] (,3.0)'

     Failure/Error: fail()

     RuntimeError:

     # ./spec/fpm/package/nuget_spec.rb:108:in `block (4 levels) in <top (required)>'
```

beddari:
I fully support @jordansissel proposal and it is perfectly inline with what I'd expect as a user.

jordansissel:
(sorry, didn't mean to delete this branch)

PR #1379, opened 2017-07-21
copy config files external to source

josephholsten:
rebased #860

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,5 +37,8 @@ Gemfile.lock
 .rbx
 .vagrant
 
+# RubyMine
+.idea/
+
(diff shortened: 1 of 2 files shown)


jordansissel:
Thank you for rebasing this!

Looking at the change, it seems like this only covers the rpm target. It's also a stretch (this is ok) to have something besides the `-s` source provide content for a package.

I like the idea, and I would like to see this expanded to all package types.

I think we can merge as-is and work on moving the code so that fpm can do this for all package types later.

josephholsten:
should we stick with the `config_files` name? also will need doc

Issue #653, opened 2014-02-28
FPM Installation problem


deekonda:
Installed FPM via "gem install fpm" command and it happened successful, but after installing, when I tried typing following command i.e "fpm -v" or "fpm -h" get to see following error and couldn't use FPM command.
Note: Am trying on Windows 7 64 bit, tried multiple times to uninstall/re-install Ruby, Devkit and fpm but no luck.

Note: Before installing installed Ruby "ruby 1.9.3p484 (2013-11-22) [i386-mingw32]" and gem version is "1.8.28", not sure if it would help.
# 

C:/Ruby193N/lib/ruby/gems/1.9.1/gems/ffi-1.9.3-x86-mingw32/lib/ffi/library.rb:26
1:in `attach_function': Function '__xmknod' not found in [msvcrt.dll] (FFI::NotF
oundError)
        from C:/Ruby193N/lib/ruby/gems/1.9.1/gems/fpm-1.0.2/lib/fpm/util.rb:16:i
n`rescue in module:Util'
        from C:/Ruby193N/lib/ruby/gems/1.9.1/gems/fpm-1.0.2/lib/fpm/util.rb:12:i
n `<module:Util>'
        from C:/Ruby193N/lib/ruby/gems/1.9.1/gems/fpm-1.0.2/lib/fpm/util.rb:6:in
`<top (required)>'
        from C:/Ruby193N/lib/ruby/site_ruby/1.9.1/rubygems/custom_require.rb:36:
in `require'
        from C:/Ruby193N/lib/ruby/site_ruby/1.9.1/rubygems/custom_require.rb:36:
in`require'
        from C:/Ruby193N/lib/ruby/gems/1.9.1/gems/fpm-1.0.2/lib/fpm/package.rb:2
:in `<top (required)>'
        from C:/Ruby193N/lib/ruby/site_ruby/1.9.1/rubygems/custom_require.rb:36:
in`require'
        from C:/Ruby193N/lib/ruby/site_ruby/1.9.1/rubygems/custom_require.rb:36:
in `require'
        from C:/Ruby193N/lib/ruby/gems/1.9.1/gems/fpm-1.0.2/lib/fpm.rb:3:in`<to
p (required)>'
        from C:/Ruby193N/lib/ruby/site_ruby/1.9.1/rubygems/custom_require.rb:36:
in `require'
        from C:/Ruby193N/lib/ruby/site_ruby/1.9.1/rubygems/custom_require.rb:36:
in`require'
[...]

PR #1380, opened 2017-07-21
Fix handling of single-digit semver constraints

josephholsten:


josephholsten:
@clayne mentioned in #755 

> This section of code was confusing and repetitive when reading it at first to try and figure out what was going on. I vote this to be a cleaner approach:

```ruby
# If attributes provided, use those, otherwise if rpm-use-file-permissions is false, use rpm defaults.
if !attrs[file].nil?
  return sprintf("%%attr(%s) %s\n", attrs[file], file)
elsif !attributes[:rpm_use_file_permissions?]
  return file
end
```

> Then kill line 148 and let it fall through to the rpm-use-file-permissions true case. Since the attribute check is winning in all cases if it's non-nil, there's no reason to be checking it twice. It's precedence is higher in the overall logic regardless of rpm-use-file-permissions being true or not.

josephholsten:
or do we just want #931/#1382?

zipkid:
Hello,
I am running into this issue when packaging aws-sdk gems.
Are there plans to get this merged & released soon?

josephholsten:
@zipkid is this patch working for you currently? Looks like the failing test is a straightforward fix.

@jordansissel any objection to my fixing the failing test and mergind?

zipkid:
@josephholsten i manually added line #653 in lib/fpm/package/deb.rb and it works like a charm.
The fail is indeed an earlier test in the spec.rb that fails on the changes.

josephholsten:
The failing is coming from [spec/fpm/package/deb_spec.rb:227](https://github.com/jordansissel/fpm/blob/master/spec/fpm/package/deb_spec.rb#L227):
```
      it "should have the correct dependencies" do
        original.dependencies.each do |dep|
          insist { input.dependencies }.include?(dep)
        end
      end
```

which is throwing:
```
Expected "semver ~> 3" in ["something > 10", "hello >= 20", "semver >= 3.0", "semver < 4.0"]
```

I'm not entirely sure how this should actually be changed.

vStone:
[0001-This-way-of-checking-the-semver-no-longer-makes-sens.patch.txt](https://github.com/jordansissel/fpm/files/1386646/0001-This-way-of-checking-the-semver-no-longer-makes-sens.patch.txt)

vStone:
@josephholsten Could you rebase this pr and include the patch I added to fix the tests? Maybe we can get @jordansissel to merge it then

josephholsten:
@vStone sorry for the delay, I've `git am`'d your patch and am waiting for build now!

jordansissel:
One related (I think) test failed in Travis:

```
  4) FPM::Package::Deb#output when read with dpkg should have the correct dependency list
     Failure/Error: insist { dpkg_field("Depends") } == "something (>> 10), hello (>= 20), semver (>= 3.0), semver (<< 4.0)"
     Insist::Failure:
       Expected "something (>> 10), hello (>= 20), semver (>= 3.0), semver (<< 4.0)", but got "something (>> 10), hello (>= 20)"
     # ./spec/fpm/package/deb_spec.rb:263:in `block (4 levels) in <top (required)>'
```

Can you check?

PR #1391, opened 2017-07-29
Restore prior --config-files behavior

jordansissel:
PR #1379/#860 introduced a new behavior for --config-files that fetches
config files from the local filesystem. The prior behavior was to
require these files to be provided by the source package (`--source`
flag).

This patch makes a hybrid solution which does, for any given
`--config-files foo` flag:

* does the path `foo` already exist in the staging area? Do nothing.
This means the package source (`-s gem` for example) provided this file.
* or, does the path `foo` exist in the local filesystem? If so, copy it
into the staging area.
* otherwise, consider this to be an error that the --config-files
setting cannot find the file, and fpm will fail with an error.

### DIFF
--- a/lib/fpm/package/rpm.rb
+++ b/lib/fpm/package/rpm.rb
@@ -474,11 +474,18 @@ class FPM::Package::RPM < FPM::Package
 
     # include external config files
     (attributes[:config_files] or []).each do |conf|
-      path = conf
-      dest_conf = File.join(staging_path, path)
-      FileUtils.mkdir_p(File.dirname(dest_conf))
-      FileUtils.cp_r conf, dest_conf
-      File.chmod(0755, dest_conf)
+      dest_conf = File.join(staging_path, conf)
+
+      if File.exist?(dest_conf)
+        logger.debug("Using --config-file from staging area", :path => conf)
+      elsif File.exist?(conf)
+        logger.info("Copying --config-file from local path", :path => conf)
+        FileUtils.mkdir_p(File.dirname(dest_conf))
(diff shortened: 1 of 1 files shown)


jordansissel:
I tested this a few ways:

* `fpm -s empty -t rpm -n example --config-files /etc/zshrc` -- copies /etc/zshrc from the local filesystem. Good!

And for a `-s dir` test, I have this layout:

```
% find /tmp/z
/tmp/z
/tmp/z/foo
/tmp/z/foo/bar
```

* `fpm -f -s dir -t rpm -n example --config-files /foo/bar /tmp/z/=/` -- correctly uses the `foo/bar` path in the package path already (provided by `-s dir`):

```
% rpm -qpc example-1.0-1.x86_64.rpm
/foo/bar
% rpm -qpl example-1.0-1.x86_64.rpm
/foo/bar
```

* `fpm -f -s dir -t rpm -n example --config-files /no/such/path /tmp/z/=/`

```
Force flag given. Overwriting package at example-1.0-1.x86_64.rpm {:level=>:warn}
Failed to find given --config-file {:path=>"/no/such/path", :level=>:error}
/home/jls/projects/fpm/lib/fpm/package/rpm.rb:487:in `block in output': Could not find config file '/no/such/path' in staging area or on host. This can happen if you specify `--config-file '/no/such/path'` but this file does not exist in the source package and also does not exist in filesystem. (RuntimeError)
```

jordansissel:
I'm ok with the manual tests I did, but I want to write tests for this feature. However, it's the weekend, and I'm not spending my time on that today :)

PR #1400, opened 2017-08-09
Replace the default value of the debian package priority option, extra => optional, close #1398

### LINKED ISSUE #1398: The debian package priority extra is deprecated
The debian package priority extra option, 

    --deb-priority PRIORITY       (deb only) The debian package 'priority' value. (default: "extra")

For the default value "_extra_", [this priority is deprecated](https://www.debian.org/doc/debian-policy/ch-archive.html#s-priorities):

> extra
This priority is deprecated. Use the optional priority instead. This priority should be treated as equivalent to optional.
The extra priority was previously used for packages that conflicted with other packages and packages that were only likely to be useful to people with specialized requirements. However, this distinction was somewhat arbitrary, not consistently followed, and not useful enough to warrant the maintenance effort.

Please replace the default value with the `optional` priority instead.

suntong:


AnthonyMastrean:
Fixes #1398

jordansissel:

PR #1403, opened 2017-08-12
Allow suppresing configuration from the source package

lostgoat:
The source package may contain undesired metadata. Through the cli we
can only currently set said metadata to an empty string to delete it.

An empty string and nil behave differently for some configuration
options, e.g. iteration results in a hyphen being appended to the
version string like so, 123-

For consistency, if a configuration option is set to an empty string,
store it as nil.


lostgoat:
Hey,
 
Been really enjoying fpm, its pretty great. Ran into this small issue recently and wanted to post back as some other users might run into a similar issue. I was basically trying to remove the "iteration" portion of a source package.

This is my first time touching ruby, so I'm not sure if this would be the best way to do this.

lib/fpm/package.rb
```diff
@@ -205,7 +205,8 @@ def convert(klass)
     ivars.each do |ivar|
       #logger.debug("Copying ivar", :ivar => ivar, :value => instance_variable_get(ivar),
                     #:from => self.type, :to => pkg.type)
-      pkg.instance_variable_set(ivar, instance_variable_get(ivar))
+      ival = instance_variable_get(ivar)
+      pkg.instance_variable_set(ivar, ival != '' ? ival : nil )
```
AnthonyMastrean: You can use [`#empty?`](https://ruby-doc.org/core-2.2.0/String.html#method-i-empty-3F) and flip the logic around a little. That's about it as far as plain Ruby goes.

```ruby
ival.empty? ? nil : ival
```

AnthonyMastrean:
I bet you could write a failing spec and then have it pass based on this change. That might help :)

> An empty string and nil behave differently for some configuration options... iteration results in a hyphen being appended to the version string like so, 123-

PR #1415, opened 2017-09-27
[Fixes #1402] Update files.rst

### LINKED ISSUE #1402: [Minor docu problem] The URL at https://fpm.readthedocs.io/en/latest/use-cases/files.html in the Makefile is no longer available
Hello,

At https://fpm.readthedocs.io/en/latest/use-cases/files.html

there is a jenkins example.

The wget line:

    wget http://ftp.osuosl.org/pub/hudson/war/1.396/jenkins.war

This URL seems to have moved or perhaps it is not available right now.

Just notifying you in case you may know the proper URL (I tried to look at 
 http://ftp.osuosl.org/pub/hudson/war/ but this one seems to no longer
exist either, not sure if permanent or temporary)

AnthonyMastrean:
Update the Jenkins/WAR example to use the more stable addresses, although they do redirect to mirrors and make the wget example output look nasty, it should be more reliable.

### DIFF
--- a/docs/use-cases/files.rst
+++ b/docs/use-cases/files.rst
@@ -9,12 +9,12 @@ We'll use `make` to script the download, but `make` isn't required if you don't 
 Makefile::
 
     NAME=jenkins
-    VERSION=1.396
+    VERSION=2.80
 
     .PHONY: package
     package:
       rm -f jenkins.war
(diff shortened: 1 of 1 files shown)


jordansissel:
Thank you :)

PR #1417, opened 2017-09-27
[Fixes #1397] It's no longer necessary to explicitly require 'rubygems'

### LINKED ISSUE #1397: Line require "rubygems" can probably safely be ignored by now
I think as of ruby 1.9, require "rubygems" is no longer required because gem is distributed as part of ruby; at the least I am quite sure that this is in ruby 2.x the case so.

People on old ruby can always call require "rubygems" manually on their own before requiring fpm too.

https://github.com/jordansissel/fpm/blob/master/bin/fpm#L3

AnthonyMastrean:
Ruby 1.9 requires Rubygems internally and this is no longer strictly required. It's been this way since 2009, so it's a little hard to come by official documentation. I hope this works...

> Note: For Ruby 1.8 you must require 'rubygems' before requiring any gems.

http://guides.rubygems.org/rubygems-basics/#requiring-code

### DIFF
--- a/bin/fpm
+++ b/bin/fpm
@@ -1,6 +1,5 @@
 #!/usr/bin/env ruby
 
-require "rubygems"
 $: << File.join(File.dirname(__FILE__), "..", "lib")
 require "fpm"
 require "fpm/command"


AnthonyMastrean:
Oh, and, obviously, this lines up with what's in [`.travis.yml`](https://github.com/jordansissel/fpm/blob/master/.travis.yml#L7)

mcandre:
Strange that the build bots are failing for this PR :/

Hmm, I don't see fpm documenting clearly which version of Ruby is required in order to run fpm. Would be good to specify this in the gemspec, and also include a .ruby-version file for the chruby fans out there who work on multiple, variably Ruby versioned projects.

AnthonyMastrean:
Unfortunately, _all_ of the [builds](https://travis-ci.org/jordansissel/fpm/builds) have been failing for some time. At least, it seems like it's just some specs (the same CPAN specs, maybe).

I got my information on supported Rubies from the [Travis CI](https://github.com/jordansissel/fpm/blob/master/.travis.yml#L7) file.

AnthonyMastrean:
See #1437

jordansissel:
I'm OK with this. I don't know how much Ruby 1.8.7 can practically be supported in year 2018 anyway.

PR #1422, opened 2017-10-14
Fixes typo with APK control scripts

charlesportwoodii:
Alpine Linux describes the available control scripts at https://wiki.alpinelinux.org/wiki/Creating_an_Alpine_package.

This patch fixes an issue where `post-install` was registered multiple times, preventing `:after_install`, `:before_install`, `:before_upgrade`, and `:after_upgrade` from running the correct script if multiple options were declared.

The issue appeared in https://github.com/jordansissel/fpm/commit/87ce028644b0bb9913b00ccd814be58056d9c28f.


jordansissel:
I don't' have tests to verify this, but the change looks correct to me. Merging!

PR #1434, opened 2017-11-06
fix: install loop problem in rpm.erb

schmto:
When using install command (from /usr/bin/install) in after_install
shell in creates a endless loop, because it jumps to funtion beginning
of install(). So it's best practise to rename it to _install() to avoid
collision with /usr/bin/install command


jordansissel:
I haven't tested, but the change looks good to me.

schmto:
yes. it works for me. tested on two different hosts.

Issue #1440, opened 2017-11-16
Error creating .deb package with --config-file


NoBodyCam:
I am running into a error creating .DEB packages if I specify --config-file. The error I get is: `Invalid package configuration: Error trying to use 'etc/name/file.conf' as a config file in the package. Does it exist? {:level=>:error, :file=>"fpm/command.rb", :line=>"502", :method=>"rescue in execute"}`

The command I am using is:
`fpm -f --verbose --debug -s python -t deb -n package-name -v 0.0.0 -a all --config-files etc/name/file.conf ./setup.py`

etc/name/file.conf does exist.

The command works if I use `-t rpm` instead of `-t deb`.

PR #1443, opened 2017-11-17
Non ruby programmer attempt to fix config-file issue with deb packages

NoBodyCam:
This is a fix to correct a Issue I am running into (#1440) when creating deb packages with the config-files option.


NoBodyCam:
I do not believe my patch is the cause of the test failures. Looking at the output of the tests I am seeing: `perl failed (exit code 2).`

jordansissel:
I like the proposed change. I'd like to see tests added, but I may write these before merging. If you are interested, please let me know :)

PR #1446, opened 2017-12-16
Require only the used backports

marcandre:
Looking at the .travis.yml, it looks like you are supporting Ruby 1.9.3+.

Assuming this is correct, this PR:

1) specifies that in the gemspec

2) loads the backports that are still in use.

As I'm ready to publish a bunch of backports for Ruby 2.5, I'd like to deprecate the `require 'backports'` (which loads all of the backports). That was, in retrospect, not such a great idea as the number of backports is increasing with every Ruby version, and could potentially be an issue for some code. Moreover, it is wasteful to load the hundreds of backports when only a few are needed.

I wrote a script to run your test suite while making a list of the backports used. Assuming the coverage is good, there's only one used in the library.

Thanks

### DIFF
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -20,6 +20,8 @@ Gem::Specification.new do |spec|
     "management without wasting pointless hours debugging bad rpm specs!"
   spec.license = "MIT-like"
 
+  spec.required_ruby_version = '>= 1.9.3'
+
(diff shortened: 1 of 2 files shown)


jordansissel:
Thank you for this patch and for your continued work on the backports gem :)

I'm ignoring the unrelated failures in Travis.

LGTM

PR #1447, opened 2017-12-21
add xzmt - multithreaded xz compression for rpmbuild [fixes #1419]

### LINKED ISSUE #1419: RFE: can you add support for using multi-threaded compression in rpmbuild
e.g. by defining

`%define _binary_payload w7T.xzdio
`
in the spec file?

Thanks!
   Amnon

amnonbc:
This change will allows the user to specify multi-threaded compression of RPMs.

RPM packaging can take a long time. Most of this time is the compression
of the rpm payload. Compression is compute heavy. But it can be speeded up 
if done on many cores in parallel rather than on a single core.
A 5Mb rpm I package takes 25 minutes using single threaded compression.
This goes down to 3 minutes when run multi-threaded.

Fixes https://github.com/jordansissel/fpm/issues/1419

### DIFF
--- a/lib/fpm/package/rpm.rb
+++ b/lib/fpm/package/rpm.rb
@@ -25,6 +25,7 @@ class FPM::Package::RPM < FPM::Package
   COMPRESSION_MAP = {
     "none" => "w0.gzdio",
     "xz" => "w9.xzdio",
+    "xzmt" => "w9T.xzdio",
     "gzip" => "w9.gzdio",
     "bzip2" => "w9.bzdio"
   } unless defined?(COMPRESSION_MAP)


SEJeff:
👍 confirmed working

jordansissel:
>  A 5Mb rpm I package takes 25 minutes using single threaded compression.

I try to understand the slowness of `xz` here. I'm unable to reproduce this "25 minutes" time, though perhaps I am not using the same environment?

I tested with fpm v1.9.3, and two files (one zeroes, one random bytes):

* `dd if=/dev/zero bs=4k count=$((5<<20 / 4<<10)) of=5mb.zero`
* `dd if=/dev/urandom bs=4k count=$((5<<20 / 4<<10)) of=5mb.random`

```
du -hs 5mb*
5.0M    5mb.random
5.0M    5mb.zero
```

Environment:

* Fedora 27
* Dual Xeon E5-2680v2 w/ NVMe storage; linux (fpm) runs in a hyper-v vm.

Base command:

`% time fpm -f -s dir -t rpm -n example 5mb.random 5mb.zero`

Results:
| flags | execution time | .rpm size |
|-|-|-|
| defaults | 0.808s | 5.1mb |
| `--rpm-compression xz` | 2.163s | 5.1mb | 

Conclusion:

xz vs default is only 1.355s difference. Seconds, not minutes?

jordansissel:
I confess I am unable to reproduce the "25 minutes" claim, and I'm unable (in my test) to reproduce xzmt being faster (8.3x faster as reported in the description).

I'm fine merging this, but I would like to understand the problem a bit better.

jordansissel:
@amnonbc Thank you for making FPM better! :)

amnonbc:
Apologies for not replying earlier. (I was travelling, with no access to internet).
Yes, there was a typo in my original description. The rpm payload was 5 Gb, rather 
than 5 Mb. And at this scale, the benefit of doing the compression in parallel is significant.
Thanks for merging this in!
I'll add some benchmarks here next week when I have access to my linux box.

jordansissel:
Ahh! This makes more sense now, hah. Thank you for clarifying :)

amnonbc:
I tried to benchmark this with 5Gb of data on a 14 core Haswell E5-2697 v3 running at  2.60GHz
```
dd if=/dev/zero bs=4k count=$((5<<30 / 4<<10)) of=5gb.zero
dd if=/dev/urandom bs=4k count=$((5<<30 / 4<<10)) of=5gb.random
```

```
time fpm -f -s dir -t rpm  --rpm-compression xz -n examplexzmt6 5gb.random 5gb.zero
```

| flags        | execution time           | rpm size  |
| ------------- |-------------|-----|
|  `--rpm-compression xz`    | 29m3s | 2.6Gb |
|  `--rpm-compression xzmt`  |  3m2s      |   2.6Gb |
|  `--rpm-compression xzmt6` | 1m13s      |   2.6Gb |

I added in the xzmt6 option which reduces the xz compression level to 6 (which is xz's default compression level.

The result show the expected speedup when parallelizing the work of compression across several cores. The test does not really allow us to see a realistic difference in compression levels, as our compressed data is 50% uncompressible and 50% trivially compressible.

amnonbc:
To try to make the test more realistic, I re-ran the tests on on the gcc-7.2.0 binaries (which weigh 1.5Gb)

| flags        | execution time           | rpm size  |
| ------------- |-------------:|-----:|
|  `--rpm-compression gzip`    | 6m10s | 508Mb |
|  `--rpm-compression xz`  |  12m14s      |   261Mb |
|  `--rpm-compression xzmt`  |  1m53      |   263Mb |
|  `--rpm-compression xzmt6` | 36s      |   275Mb |

The results are unsurprising:
- xz gives better compression than gzip (which is why fedora has been packaged using it since F12)
- If you use lots of cores to do your compression, the compression will finish earlier
- If you use a less agressive level of XZ compression, it will get done faster, but the rpm will be bigger (though still a great deal smaller than gzip level 7 compression)

centminmod:
would definitely like to have multi-threaded compression support.

I did some benchmarks at https://community.centminmod.com/threads/compression-comparison-benchmarks-zstd-vs-brotli-vs-pigz-vs-bzip2-vs-xz-etc.12764/ and the difference can be huge !

bviktor:
Does this actually work for anyone? I tried modifying `/usr/lib/rpm/macros`

```
%_source_payload w3T8.xzdio
%_binary_payload w3T8.xzdio
```

I tried

```
--rpm-compression=xzmt
```

I tried

```
--rpm-rpmbuild-define "_binary_payload w3T8.xzdio"
```

I also tried `--rpm-compression xzmt6` but fpm will error out right away with no such option. No matter what I do, rpmbuild always uses just 1 core.

FPM 1.11.0, Ruby 2.5, RPM 4.14.1 on Ubuntu 18.04. But the same happens in my CentOS 7 docker containers.

bviktor:
Oh nevermind, I guess I just didn't give it enough time. I'll do some more testing but I guess it'll work.

image

In any case, I think it'd be great to either allow for specifying compression level or use a saner default, like 6. Level 9 is a complete waste of time.

https://www.rootusers.com/gzip-vs-bzip2-vs-xz-performance-comparison/

amnonbc:
:+1

sevenfish:
@centminmod 
In latest version fpm 1.17, things changed.
In "/var/lib/gems/3.2.0/gems/fpm-1.17.0/lib/fpm/package/rpm.rb" file,  "xzmt" => "T.xzdio", in COMPRESSION_MAP, and the payload_compression is defined as 
` return "w#{attributes[:rpm_compression_level]}" + COMPRESSION_MAP[attributes[:rpm_compression]`

So for --rpm-compression xzmt to actually take effect, it needs to be used together with --rpm-compression-level 8. Only then will the payload_compression function return w8T.xzdio, (using %{getncpus} threads)

PR #1455, opened 2018-01-09
Resolve deb-creation bug with docker

FrankPetrilli:
Without this, I receive an error on attempting to create a deb, since it calls `tar` with the `--user` flag among others which busybox tar does not have.


FrankPetrilli:
@dmuth, calling your attention to this in particular.

PR #1484, opened 2018-04-13
Dockerfile: Follow best practices

allanlewis:
In order to follow best practice for Dockerfiles, this commit:

* Removes leading and trailing blank lines.
* Pins the base image to a specific version of Alpine; the latest, 3.7, is used.
* Tells `apk` to not cache any package lists.
* Puts each dependency on its own line.
* Does some minor reformatting for readability.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,16 @@
-
 #
 # To build this Docker image: docker build -t fpm .
 #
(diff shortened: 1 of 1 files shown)


allanlewis:
Thanks, @jordansissel!

PR #1488, opened 2018-04-18
Fixes typo; claims to make an rpm, but really deb

artburkart:
I just clicked the edit button on the readthedocs sites, so I apologize if this wasn't the correct way to make a PR for this site.


jordansissel:
Thank you both. Merging.

PR #1489, opened 2018-04-23
Changed the mode of templates file for deb packages to 0644

dreamlax:
If "--deb-templates" is used, fpm uses the wrong mode for this file. Since it is not an executable file, it should not have the executable bits set. The lintian tool complains (with an error) that the mode is incorrect, and that it should not be executable. Changing the mode to 0644 resolves this.


jordansissel:
Thanks for helping improve fpm! 

Long term, I'd like to see this scenario have some test coverage, but there's no need to stall this PR waiting for tests. For context, there's some existing tests which verify lintian is error-free, but I don't think these tests use --deb-templates.

Issue #1149, opened 2016-07-04
[Query] Can fpm support package formats that bundle dependencies? (AppImage, Flatpak, Guix, gx, Snappy)


kwill:
Some new(ish) package formats bundle (most) dependencies in the package itself. (Their common goals seem to include multi-distro support and "solving the dependency problem".) Is this approach supportable?

These formats include:
- AppImage http://appimage.org
- Flatpak http://flatpak.org
- Guix https://www.gnu.org/software/guix/manual/html_node/Packaging-Guidelines.html [1]
- gx https://github.com/whyrusleeping/gx [2]
- Snappy https://snapcraft.io

If this is possible, support to/from each format presumably warrants separate tickets? Which would be the best resource to point the maintainers of AppImage et al to, if they wanted to contribute to fpm themselves?

[1] [2] Guix and gx have other implementation details that would require individual investigation.

kwill:
AppImage support is already ticketed: https://github.com/jordansissel/fpm/issues/1133

kwill:
See also https://github.com/mikix/deb2snap

hatt:
I've been investigating Snap and AppImage over the past few days for FPM. As far as dependencies, it would depend on how easily the source gives up it's dependencies. Ruby gems would be a relatively simple one while Perl modules not so much. The main thing I see is that all these formats are tailored to Linux and require compilation steps, which is a little bit annoying for cross-platform compatibility. I don't think anyone expects to build for Linux on OSX though so that's probably not a big deal.

kwill:
Snap / Snappy support is ticketed at https://github.com/jordansissel/fpm/issues/1181

elopio:
I'd be happy to help on the snapcraft side. My ruby is not strong though.
@hatt from your experiments, did you see anything that could make it easier for you to support snaps?

hatt:
@elopio actually snap is the easiest of the formats I looked at for this - the biggest issues are as I detailed above figuring out dependencies for building, and then a nice way to get extra metadata (ports, interfaces, plugins, etc) into the snap definition.

[...]

PR #1490, opened 2018-04-23
Introduce support for snap packages

### LINKED ISSUE #1181: Produce 'snap' packages
The snap format is a self-contained package with dependencies that works across newer releases of Ubuntu, Debian, Arch, Fedora etc. It feels a bit like 'build and run in a subdirectory of $HOME' because even if the app runs as root it is sandboxed to write in only designated directories. There are 'system' and 'user' writable directories.

There are likely to be two particular requirements that need consideration. One is to determine and dictate where the app writes. Many apps will respect either config or build-time options to direct them to write in a particular place. Others need a patch applied. The other issue is that apps which need to interact outside the sandbox can have additional metadata added to the package to describe extensions of the sandbox that allow such interactions.

kyrofa:
[Snaps][1] are self-contained squashfs images. This PR resolves #1181 by adding basic support for
using them as inputs and outputs.

Typically Snapcraft is used to create snaps, but #1149 indicated a desire to not use Snapcraft, so the snap is created natively, using mksquashfs at the end to actually create the squashfs image. This may not work for all use-cases (e.g. when the source doesn't already have its dependencies bundled).

[1]: https://snapcraft.io/


kyrofa:
Older Ruby versions don't seem to have yaml safe_load... but the tests on those versions seem to be failing for other reasons as well. In fact, the entire suite seems to having trouble.

kyrofa:
Hey there, any chance for a review?

lib/fpm/package/snap.rb
```diff
@@ -0,0 +1,130 @@
...
+        "name" => self.name,
+        "version" => self.version,
+        "summary" => summary,
+        "description" => description,
+        "architectures" => [self.architecture],
+        "confinement" => self.attributes[:snap_confinement],
+        "grade" => self.attributes[:snap_grade],
+      }
+
+      unless attributes[:snap_apps].nil? or attributes[:snap_apps].empty?
```
jordansissel: (I haven't used snap before, so I'm glancing at the docs while reviewing this PR, let me know if I'm confused)

The default snap type is `app`, and the way I read this is that if the user does not specify `--snap-yaml` flag, then they will generate an `app` snap without any apps? Would this kind of snap make sense to a snap user to have an app without any apps defined?

lib/fpm/package/snap.rb
```diff
@@ -0,0 +1,130 @@
...
+      if remainder.any?
+        description = remainder.join("\n")
+      end
+
+      yaml_data = {
+        "name" => self.name,
+        "version" => self.version,
+        "summary" => summary,
+        "description" => description,
+        "architectures" => [self.architecture],
```
jordansissel: If you think it is appropriate, you could convert rpm's `noarch` concept to `all` similar to how it is done in fpm for [rpm](https://github.com/jordansissel/fpm/blob/master/lib/fpm/package/rpm.rb#L231-L234) and [deb](https://github.com/jordansissel/fpm/blob/master/lib/fpm/package/deb.rb#L221-L223). This mainly to help users who float between package formats day to day. 

It is OK to to say "no" to this suggestion :)

kyrofa: Ah, good idea!

lib/fpm/package/snap.rb
kyrofa: You're right, it would be rare for an app snap to have no apps and be deemed useful. Ideally I could figure out a way to have them nicely specified via CLI arguments, but they're complex enough I couldn't come up with anything that made me happy (at least in the first rev). Should I make the snap.yaml arg required, then, you think?

jordansissel:
I'm content merging this as-is. Upon reviewing my comments, I think they are mostly small suggestions and nothing that merits further delay. Additionally, the comments I made, if resolved, could (I believe) be done without breaking compatibility, so there's not much to worry about.

AnthonyMastrean:
Will there be updates to the readme, wiki, or command-line help?

jordansissel:
I agree those additions are welcome :)

I hadn’t done them yet and they weren’t a part of this PR.

PR #1492, opened 2018-05-07
Generate .changes file for Debian

### LINKED ISSUE #467: Signed debian package with .changes file and source
Hi,

Would it be possible for fpm to create a signed .deb packages?

For ex this output from a debuild:

apache-tomcat_6.0.37-2_amd64.changes
apache-tomcat_6.0.37-2.debian.tar.gz
apache-tomcat_6.0.37-2_amd64.deb  
apache-tomcat_6.0.37-2.dsc  
apache-tomcat-6.0.37.tar.gz

This could then be sent to a debian dak server for verification and addition to the repository.

Thanks,
Pierre

NetSpida:
Issue #467 calls for creation of .debain.tar.gz, .dsc and .changes files.

This Patch adds generation of .changes file for Debian packages. 
.changes files (containing all files relevant to a package with their checksums) are used by some repository management tools when importing packages. 
Generating a .debian.tar.gz (which would contain the Debian specifics for this package, for example the rules file (basically, Makefile), the control file (specifying dependencies) and Debian specific patches) or a .dsc (which would specify the source package) might not be relevant for the use-case of fpm, though. 
Fixes #467.

The command line option --(no-)deb-generate-changes allows to select to generate the .changes file, the option --deb-dist allows to set a distribution (like sid) for the package.

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -50,6 +50,8 @@ class FPM::Package::Deb < FPM::Package
     value
   end
 
+  option "--dist", "DIST-TAG", "Set the deb distribution."
+
   # Take care about the case when we want custom control file but still use fpm ...
   option "--custom-control", "FILEPATH",
     "Custom version of the Debian control file." do |control|
@@ -87,6 +89,10 @@ class FPM::Package::Deb < FPM::Package
   option "--changelog", "FILEPATH", "Add FILEPATH as debian changelog" do |file|
     File.expand_path(file)
   end
+
+  option "--generate-changes", :flag,
+    "Generate PACKAGENAME.changes file.",
+    :default => false
 
   option "--upstream-changelog", "FILEPATH", "Add FILEPATH as upstream changelog" do |file|
     File.expand_path(file)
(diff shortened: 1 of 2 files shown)


jordansissel:
I think #467 is requesting signed packages, so I am not sure if this completely solves that request, but it looks like it's a necessary step? Can you help me understand the workflow -- For example, if I run `bin/fpm -s dir -t deb -n example --deb-generate-changes my files here`, what is the next step for you to sign it?

For some context on #467, at work, we generate debs with fpm (logstash, beats, kibana) and I think we sign those deb packages without any `.changes` files. To accomplish this, we invoke `dpkg-sig --sign <gpg-key_name> <the-package.deb>`. That said, it's possible you're signing things differently than I do at work.

lib/fpm/package/deb.rb
```diff
@@ -50,6 +50,8 @@ class FPM::Package::Deb < FPM::Package
     value
   end
 
+  option "--dist", "DIST-TAG", "Set the deb distribution."
```
jordansissel: Can you set a default value: `:default => "unstable"`

My hope is that this improves documentation (the `--help` shows you the default), and further, it would let you remove one line of code:

(This line below can be removed if you set a default value)
`       distribution = self.attributes[:deb_dist] ? "#{self.attributes[:deb_dist]}" : "unstable"; ` further down in this file.

jordansissel:
Did one manual test ✅ beyond the test suite:

```
% bin/fpm -s dir -t deb -n example --deb-generate-changes ~/.zshrc
Created changes {:path=>"example_1.0_amd64.changes"}
Created package {:path=>"example_1.0_amd64.deb"}
```

Then I briefly reviewed the `.changes` file with my eyes and it looked ok (caveat: I have no recent experience with .changes files, so it's hard to evaluate further).

Issue #1271, opened 2017-01-11
Providing several --deb-systemd files should be supported


Jeger:
If you add several --deb-systemd files they seem to get packaged correctly, but the `postinst` and `prerm` scripts only contain code referring to the last one.

timcoote:
Is there any work around for this? Can I provide a hacked version of postinst somehow (eg generate with one service, then manually add in the rest)?

timcoote:
fwiw, I'm creating multiple packages - one for each service, with a dependency DAG between them.

ronaldevers:
Also running in to this. Thinking to just add additional prerm/postinst scripts manually...

ronaldevers:
Hmm, actually, that does not work either. It is generating a bogus `postinst` script in the `control.tar.gz` in the deb now:

```
#!/bin/sh
after_upgrade() {
    :

systemctl --system daemon-reload >/dev/null || true
if ! systemctl is-enabled my-service >/dev/null 
then
    systemctl enable my-service >/dev/null || true
    systemctl start my-service >/dev/null || true
else
    systemctl restart my-service >/dev/null || true
fi
}

after_install() {
    :
#!/bin/sh
after_upgrade() {
    :
...
```

It seems to interleave the output of multiple `--after-install` options or the `--after-install` conflicts with `--deb-systemd`.

Issue #1363, opened 2017-06-28
Provide a way not to auto enable and start systemd units


nemanja-boric-sociomantic:
In case template unit files are added to `-deb-systemd`, fpm will try to enable and start this service, which is not possible. There should be an attribute to prevent auto-enable & start of the service, and leave it to the user. Perhaps just the `systemctl reload` should be mandatory.

cam-fulcrum:
Yeah, unfortunately this overrides the `exit 101` trick too https://jpetazzo.github.io/2013/10/06/policy-rc-d-do-not-start-services-automatically/

jordansissel:
Agreed.

To add this setting, you'd want to look in the [templates directory](https://github.com/jordansissel/fpm/blob/master/templates/deb/) for things like [this](https://github.com/jordansissel/fpm/blob/master/templates/deb/postinst_upgrade.sh.erb#L35-L37) (where systemd start/enable are invoked)

Issue #1284, opened 2017-01-18
Debian post install scripts should use deb-systemd-invoke instead of systemctl


jdelic:
When using `fpm` with `--deb-systemd` the following incantations will be automatically added to the post-install script:

```
systemctl --system daemon-reload >/dev/null || true
if ! systemctl is-enabled myservice >/dev/null
then
    systemctl enable myservice >/dev/null || true
    systemctl start myservice >/dev/null || true
else
    systemctl restart myservice >/dev/null || true
fi
```

All instances of `systemctl enable|start|restart` should be replaced with `deb-systemd-invoke` which is a wrapper that respects `policy-rc.d` and was specifically added to Debian for that purpose.

jensschroer:
Is deb-systemd-invoke available for other platforms that use the deb format as well? For example ubuntu or yocto?

cappaberra:
At least for Ubuntu...

The deb-systemd-invoke perl script is available in many flavors of Ubuntu land:  trusty, xenial, yakkety, zesty, and artful.  The package name is init-system-helpers:  https://packages.ubuntu.com/xenial/init-system-helpers

jhermann:
BTW, this easily makes things work if that helper is not available:

    sysctl=$(command which deb-systemd-invoke || echo systemctl)
    $sysctl …

jordansissel:
@jhermann this assumes `which` is available (it may not on some systems), but I agree with your idea. You could implement `which` in shell (check each $PATH entry for an executable with that name) and this should work.

cam-fulcrum:
Ok, so as [Debian](https://wiki.debian.org/Shell) and [Ubuntu[(https://wiki.ubuntu.com/DashAsBinSh) both use [dash](https://linux.die.net/man/1/dash) as the default `/bin/sh`, we can use the `command -v` builtin instead.

```
sysctl=$(command -v deb-systemd-invoke || echo systemctl)
$sysctl...
```

cam-fulcrum:
Ok, so as [Debian](https://wiki.debian.org/Shell) and [Ubuntu[(https://wiki.ubuntu.com/DashAsBinSh) both use [dash](https://linux.die.net/man/1/dash) as the default `/bin/sh`, we can use the `command -v` builtin instead.

```
sysctl=$(command -v deb-systemd-invoke || echo systemctl)
$sysctl...
```

PR #1498, opened 2018-05-24
Allow auto-enable and auto-start for deb-systemd services after install

c-ameron:
Hi, this PR adds in two new options to enable/disable and autostart systemd services

`--[no-]deb-systemd-enable` to enable the systemd service after install
`--[no-]deb-systemd-auto-start` to auto start the service after install

Additionally, any service start/stop/restart will use the `deb-systemd-invoke` wrapper if installed on the system

This should solve https://github.com/jordansissel/fpm/issues/1363 and https://github.com/jordansissel/fpm/issues/1284

Even though you can now disable a service from starting by default, I think its important to still allow for using [deb-systemd-invoke](http://manpages.ubuntu.com/manpages/trusty/man1/deb-systemd-invoke.1p.html) to keep in line with debian standards.

Let me know if you need anything changed! Thanks :)

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -181,6 +181,10 @@ class FPM::Package::Deb < FPM::Package
     next File.expand_path(file)
   end
 
+  option "--systemd-enable", :flag , "Enable service on install or upgrade", :default => true
+
+  option "--systemd-auto-start", :flag , "Start service after install or upgrade", :default => true
+
   option "--systemd-restart-after-upgrade", :flag , "Restart service after upgrade", :default => true
 
   option "--after-purge", "FILE",
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -181,9 +181,9 @@ class FPM::Package::Deb < FPM::Package
     next File.expand_path(file)
   end
 
-  option "--systemd-enable", :flag , "Enable service on install or upgrade", :default => true
+  option "--systemd-enable", :flag , "Enable service on install or upgrade", :default => false
 
-  option "--systemd-auto-start", :flag , "Start service after install or upgrade", :default => true
+  option "--systemd-auto-start", :flag , "Start service after install or upgrade", :default => false
 
   option "--systemd-restart-after-upgrade", :flag , "Restart service after upgrade", :default => true
 
@@ -444,7 +444,7 @@ class FPM::Package::Deb < FPM::Package
       attributes[:deb_systemd] << name
     end
 
-    if script?(:before_upgrade) or script?(:after_upgrade) or not attributes[:deb_systemd].empty?
+    if script?(:before_upgrade) or script?(:after_upgrade) or attributes[:deb_systemd].any?
       puts "Adding action files"
       if script?(:before_install) or script?(:before_upgrade)
         scripts[:before_install] = template("deb/preinst_upgrade.sh.erb").result(binding)
@@ -452,7 +452,7 @@ class FPM::Package::Deb < FPM::Package
       if script?(:before_remove) or not attributes[:deb_systemd].empty?
         scripts[:before_remove] = template("deb/prerm_upgrade.sh.erb").result(binding)
       end
-      if script?(:after_install) or script?(:after_upgrade) or not attributes[:deb_systemd].empty?
+      if script?(:after_install) or script?(:after_upgrade) or attributes[:deb_systemd].any?
         scripts[:after_install] = template("deb/postinst_upgrade.sh.erb").result(binding)
       end
       if script?(:after_remove)
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -432,6 +432,7 @@ class FPM::Package::Deb < FPM::Package
       raise "#{name}: tar is insufficient to support source_date_epoch."
     end
 
+    attributes[:deb_systemd] = []
     attributes.fetch(:deb_systemd_list, []).each do |systemd|
       name = File.basename(systemd, ".service")
       dest_systemd = staging_path("lib/systemd/system/#{name}.service")
@@ -439,19 +440,19 @@ class FPM::Package::Deb < FPM::Package
       FileUtils.cp(systemd, dest_systemd)
       File.chmod(0644, dest_systemd)
 
-      # set the attribute with the systemd service name
-      attributes[:deb_systemd] = name
+      # add systemd service name to attribute
+      attributes[:deb_systemd] << name
     end
 
-    if script?(:before_upgrade) or script?(:after_upgrade) or attributes[:deb_systemd]
+    if script?(:before_upgrade) or script?(:after_upgrade) or not attributes[:deb_systemd].empty?
       puts "Adding action files"
       if script?(:before_install) or script?(:before_upgrade)
         scripts[:before_install] = template("deb/preinst_upgrade.sh.erb").result(binding)
       end
-      if script?(:before_remove) or attributes[:deb_systemd]
+      if script?(:before_remove) or not attributes[:deb_systemd].empty?
         scripts[:before_remove] = template("deb/prerm_upgrade.sh.erb").result(binding)
       end
-      if script?(:after_install) or script?(:after_upgrade) or attributes[:deb_systemd]
+      if script?(:after_install) or script?(:after_upgrade) or not attributes[:deb_systemd].empty?
         scripts[:after_install] = template("deb/postinst_upgrade.sh.erb").result(binding)
       end
       if script?(:after_remove)
--- a/templates/deb/postinst_upgrade.sh.erb
+++ b/templates/deb/postinst_upgrade.sh.erb
@@ -9,9 +9,9 @@ after_upgrade() {
 <% end -%>
 
 <%# if any systemd services specified, loop through and start them -%>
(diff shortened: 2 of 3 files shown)


c-aamm:
I've updated this PR to fix https://github.com/jordansissel/fpm/issues/1271

We pass an array of service names to the post install script and start them individually instead of only the last service in  `deb_systemd_list`. Let me know if this implementation fits the style with the rest of fpm 😃

ilederman:
I came across this today and it would be a great feature to have

c-aamm:
Hi @jordansissel - any chance this could please be merged? I've just updated it to properly remove any services that get installed. Thanks!

jordansissel:
I will review and test soon. thank you for making fpm better :)

ilederman:
Hi, any updates to this?

jordansissel:
Noted! I will merge this before the next release.

templates/deb/postinst_upgrade.sh.erb
```diff
@@ -8,18 +8,26 @@ after_upgrade() {
 <%=  script(:after_upgrade) %>
 <% end -%>
 
-<% if attributes[:deb_systemd] -%>
+<%# if any systemd services specified, loop through and start them -%>
+<% unless attributes[:deb_systemd].empty? -%>
```
jordansissel: For future style note, I prefer to avoid this kind of usage where a double-negative is used. I find it hard to parse (in my head). By double-negative I mean `unless` being a negative of `if` and `.empty?` being the negative of `.any?`. No change required at this time for this line.

"unless empty" can be written without negatives as "if any": `if attributes[:deb_systemd].any?`

In general, I prefer to avoid `unless` in most places with exception to an inline use `foo(..) unless condition` because I find `unless` a bit of a mental twist to read.

templates/deb/postinst_upgrade.sh.erb
```diff
@@ -8,18 +8,26 @@ after_upgrade() {
...
 systemctl --system daemon-reload >/dev/null || true
-if ! systemctl is-enabled <%= attributes[:deb_systemd] %> >/dev/null 
+sysctl=$(command -v deb-systemd-invoke || echo systemctl)
+<% attributes[:deb_systemd].each do |service| -%>
+if ! systemctl is-enabled <%= service %> >/dev/null
```
jordansissel: Did you mean `"$sysctl"` here instead of `systemctl` ? Also, for variable names, `sysctl` may confuse future maintainers as `sysctl` is a different tool from `systemctl`. This comment applies to any invocations of `systemctl` where `$sysctl` seems the appropriate replacement.

templates/deb/postinst_upgrade.sh.erb
```diff
@@ -31,10 +39,18 @@ after_install() {
...
-<% if attributes[:deb_systemd] -%>
+<%# if any systemd services specified, loop through and start them -%>
+<% unless attributes[:deb_systemd].empty? -%>
 systemctl --system daemon-reload >/dev/null || true
-systemctl enable <%= attributes[:deb_systemd] %> >/dev/null || true
-systemctl start <%= attributes[:deb_systemd] %> >/dev/null || true
+sysctl=$(command -v deb-systemd-invoke || echo systemctl)
+<% attributes[:deb_systemd].each do |service| -%>
+<% if attributes[:deb_systemd_enable?]-%>
+systemctl enable <%= service %> >/dev/null || true
```
jordansissel: did you mean `$sysctl` instead of `systemctl` here?

templates/deb/postinst_upgrade.sh.erb
```diff
@@ -8,18 +8,26 @@ after_upgrade() {
...
 systemctl --system daemon-reload >/dev/null || true
-if ! systemctl is-enabled <%= attributes[:deb_systemd] %> >/dev/null 
+sysctl=$(command -v deb-systemd-invoke || echo systemctl)
+<% attributes[:deb_systemd].each do |service| -%>
+if ! systemctl is-enabled <%= service %> >/dev/null 
 then
-    systemctl enable <%= attributes[:deb_systemd] %> >/dev/null || true
-    systemctl start <%= attributes[:deb_systemd] %> >/dev/null || true
+<% if attributes[:deb_systemd_enable?]-%>
+    systemctl enable <%= service %> >/dev/null || true
```
jordansissel: did you mean `$sysctl` instead of `systemctl` here?

templates/deb/prerm_upgrade.sh.erb
```diff
@@ -8,10 +8,12 @@ before_remove() {
...
 <% end -%>
 
-<%# Removing the systemd service-%>
-<% if attributes[:deb_systemd] -%>
-systemctl stop <%= attributes[:deb_systemd] %> >/dev/null || true
-systemctl disable <%= attributes[:deb_systemd] %> >/dev/null || true
+<%# Stop and remove any systemd services that were installed-%>
+<% unless attributes[:deb_systemd].empty? -%>
+<% attributes[:deb_systemd].each do |service| -%>
+systemctl stop <%= service %> >/dev/null || true
```
jordansissel: did you mean `$sysctl` instead of `systemctl` here?

templates/deb/prerm_upgrade.sh.erb
```diff
@@ -8,10 +8,12 @@ before_remove() {
 <%=  script(:before_remove) %>
 <% end -%>
 
-<%# Removing the systemd service-%>
-<% if attributes[:deb_systemd] -%>
-systemctl stop <%= attributes[:deb_systemd] %> >/dev/null || true
-systemctl disable <%= attributes[:deb_systemd] %> >/dev/null || true
+<%# Stop and remove any systemd services that were installed-%>
+<% unless attributes[:deb_systemd].empty? -%>
+<% attributes[:deb_systemd].each do |service| -%>
```
jordansissel: Did you want to use the same "try deb-systemd-invoke or systemctl" check here as your other scripts do?

lib/fpm/package/deb.rb
```diff
@@ -428,26 +432,27 @@ def output(output_path)
...
       File.chmod(0644, dest_systemd)
 
-      # set the attribute with the systemd service name
-      attributes[:deb_systemd] = name
+      # add systemd service name to attribute
+      attributes[:deb_systemd] << name
     end
 
-    if script?(:before_upgrade) or script?(:after_upgrade) or attributes[:deb_systemd]
+    if script?(:before_upgrade) or script?(:after_upgrade) or not attributes[:deb_systemd].empty?
```
jordansissel: Similar style comment (no change required) - double negatives can be hard to read, and `or attributes[...].any?` to me is clearer than `not attributes[...].empty?`

lib/fpm/package/deb.rb
```diff
@@ -181,6 +181,10 @@ class FPM::Package::Deb < FPM::Package
     next File.expand_path(file)
   end
 
+  option "--systemd-enable", :flag , "Enable service on install or upgrade", :default => true
+
+  option "--systemd-auto-start", :flag , "Start service after install or upgrade", :default => true
```
jordansissel: I would prefer this default to false. Otherwise, we're breaking behavior from a previous release where services were not automatically started. Users wishing to have a service auto start should be required to use this flag.

lib/fpm/package/deb.rb
```diff
@@ -181,6 +181,10 @@ class FPM::Package::Deb < FPM::Package
     next File.expand_path(file)
   end
 
+  option "--systemd-enable", :flag , "Enable service on install or upgrade", :default => true
```
jordansissel: I would prefer this default to false. Otherwise, we're breaking behavior from a previous release where services were not automatically started. Users wishing to have a service enabled out of the box should be required to use this flag.

templates/deb/postinst_upgrade.sh.erb
c-ameron: nope, `deb-systemd-invoke` doesn't have an `enable` option, so `systemctl` is fine to be used http://manpages.ubuntu.com/manpages/bionic/man1/deb-systemd-invoke.1p.html

templates/deb/postinst_upgrade.sh.erb
c-ameron: same comment as above

templates/deb/postinst_upgrade.sh.erb
c-ameron: good find, I've changed it `debsystemd` instead

c-ameron:
@jordansissel Thanks for your help with the PR, sorry its taken a while to get onto this.
I've updated the code per your comments

- The enabling and autostart is off by default
- Use `debsystemd` as a command wrapper for _deb-systemd-invoke or systemctl_ instead of `$sysctl`
- Style changes per your comments

Please let me know if there's anything else I can do 😄

lib/fpm/package/deb.rb
c-ameron: I've defaulted both to false

templates/deb/postinst_upgrade.sh.erb
c-ameron: I've changed this to suit your style :)

jordansissel:
> sorry its taken a while to get onto this

No worries. There's only so many hours in a day. :)

vutny:
Awesome! Thank you @c-ameron and @jordansissel .

Issue #1499, opened 2018-05-29
Please add option definitions for  pleaserun 


hbarrett:
Although the[ underlying functionality](https://github.com/jordansissel/fpm/commit/37cc459dd8088f9fa70df117d718e163ab2578a3) is present, most pleaserun option definitions are not, so most of the options cannot be used.
Please add the option definitions in pleaserun.rb.
For example, adding  `option "--user", "USER", "The user to use for executing this program."` to line 21 will allow users to utilize --pleaserun-user

Issue #1449, opened 2017-12-27
--pleaserun-user & others aint work


iggu:
As stated [here](http://fpm.readthedocs.io/en/latest/changelog.html) in changelog, fpm starting from version 1.9.1 supports –pleaserun-* set of options. But actually, none of [them](9dc67031e2213c338a0ecd86c1d0197458c1b6b0) actually work (at least in version 1.9.3), neither in form --pleaserun-user, nor as --user.

Do they actually should work? If so, how actually should they be used? They are not mentioned anywhere, not at readthedocs pages, not in --help output.

PR #1500, opened 2018-05-29
Added options for pleaserun

hbarrett:
To resolve [issue 1449](https://github.com/jordansissel/fpm/issues/1449), and my own [issue 1499](https://github.com/jordansissel/fpm/issues/1499)
These were copy pasted from pleaserun --help

### DIFF
--- a/lib/fpm/package/pleaserun.rb
+++ b/lib/fpm/package/pleaserun.rb
@@ -18,6 +18,26 @@ class FPM::Package::PleaseRun < FPM::Package
 
   option "--name", "SERVICE_NAME", "The name of the service you are creating"
   option "--chdir", "CHDIR", "The working directory used by the service"
+  option "--user", "USER", "The user to use for executing this program."
+  option "--group", "GROUP", "The group to use for executing this program."
+  option "--umask", "UMASK", "The umask to apply to this program"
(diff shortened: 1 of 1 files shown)


jordansissel:
I haven't tested yet, but the code looks good. I will try to make time to test this week
.

sbrl:
Is there any update on this? This would be a very useful feature, as I'd like to avoid running services as root when using pleaserun.

sbrl:
Is there any update on this, @jordansissel? This would be a very useful feature, as I'd like to avoid running services as root when using pleaserun.

zgazak:
I'm also very curious about this one.  It looks like --pleaserun-user (or --user) has already been added, but I can't invoke from CLI.

zgazak:
Actually I'm going to back off on my comment.  I was using gem installed version 1.14.2, to which --pleasrun-user isn't yet added.  Installing from source I *can* invoke --pleaserun-user, but I don't see the result I expect.  I count this up to my own ignorance with package building, and I'm going to attempt to switch to using FPM and pleaserun independently until I'm not a danger to myself and others anymore.  Love both FPM and pleaserun, by the way!  Made an impossible task workable for me.

PR #1510, opened 2018-07-04
Perl CPAN Support, Fix Broken PERL5LIB & Add cpanm Verbose Option

wbraswell:
We always need to properly set PERL5LIB, even if CPAN tests are disabled, so this is the only valid way to handle it:

```
 make = [ "env", "PERL5LIB=#{build_path("cpan/lib/perl5")}", "make" ]
```

Thus, I have simply deleted the other incorrect lines and only left this one line remaining.

Also, I added the ability to produce verbose output from `cpanm` via the `--cpan-verbose` option, which is important because sometimes the process will seem to hang during long builds with no output produced, and with this option at least we can see what is happening in the background.

### DIFF
--- a/lib/fpm/package/cpan.rb
+++ b/lib/fpm/package/cpan.rb
@@ -25,6 +25,9 @@ class FPM::Package::CPAN < FPM::Package
   option "--test", :flag,
     "Run the tests before packaging?", :default => true
 
+  option "--verbose", :flag,
+    "Produce verbose output from cpanm?", :default => true
+
   option "--perl-lib-path", "PERL_LIB_PATH",
     "Path of target Perl Libraries"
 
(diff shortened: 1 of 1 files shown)


lib/fpm/package/cpan.rb
```diff
@@ -25,6 +25,9 @@ class FPM::Package::CPAN < FPM::Package
   option "--test", :flag,
     "Run the tests before packaging?", :default => true
 
+  option "--verbose", :flag,
+    "Produce verbose output from cpanm?", :default => true
```
jordansissel: Thoughts on this being false by default instead of true?

wbraswell: oops that's a bug, stand by for a new pull request!

wbraswell: (I was literally JUST fixing this as you made the comment, haha!)

jordansissel:
The code change looks OK to me. I'm testing it now.

jordansissel:
@wbraswell I see you closed this, was that on purpose or did you still want this reviewed for merging?

wbraswell:
@jordansissel 
replaced by
https://github.com/jordansissel/fpm/pull/1511

PR #1516, opened 2018-07-04
Version bump to 1.10.2

jordansissel:
@wbraswell found another bug while testing 1.10.1, so it's time for a quick patch to fix that :)


RX14:
Please can you make a github tag for this release, thanks.

jordansissel:
@RX14 oops, thank you for raising. 1.10.2 is now tagged. :)

PR #1518, opened 2018-07-04
Perl CPAN Support, Broken tar --strip-components Command, Fix

wbraswell:
For handling CPAN tarballs with leading "./Foo/" directory structures, we must employ a `tar --transform` regex instead of a `tar --strip-components` command.

I also edited the output format of the failed commands in util.rb safesystem so developers can actually copy/paste the commands for debugging.  (Feel free to do the same to safesystemout if you like.)


wbraswell:
Closing this pull request, will make another pull request which includes this plus other commits.

Issue #1540, opened 2018-08-10
Alternate compression methods in .deb control files not supported


mvastola:
When editing a debian package, `fpm` appears to assume that the control archive is always stored at `control.tar.gz`.  As a result, it runs something like:
```bash
ar p [PACKAGE] control.tar.gz | tar -zxf - -C /tmp/package-deb-build-[IDENTIFIER]/control
```

However, the [specifications](https://manpages.debian.org/unstable/dpkg-dev/deb.5.en.html) for the debian package format -- as documented in `deb(5)` -- state:
> The second required member is named control.tar. It is a tar archive containing the package control information, either not compressed (supported since dpkg 1.17.6), or compressed with gzip (with .gz extension) or xz (with .xz extension, supported since 1.17.6) [...]

Indeed, Ubuntu seems to now be distributing all it's packages (at least as of bionic) with the `data` and `control` members as `.tar.xz` files.  This results in a "process failed" error every time you use `fpm` to read such a file.

PR #1542, opened 2018-08-10
Auto-detects compression (if any) of elements of .deb packages.

mvastola:
Fixes jordansissel/fpm#1540

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -22,7 +22,7 @@ class FPM::Package::Deb < FPM::Package
   } unless defined?(SCRIPT_MAP)
 
   # The list of supported compression types. Default is gz (gzip)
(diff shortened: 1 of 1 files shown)


mvastola:
Note: CI checks are failing, but all issues present predate this commit.

jordansissel:
Merged w/ slight modification to the commit message.

@mvastola Thank you for improving FPM :)

ayufan:
Can we release a new version of FPM? As it seems that using `fpm 1.11.0` is broken unless `--debian-compression gzip` is used. It seems to produce `control.tar.gz` in a format that is defined `debian-compression` instead of using `gzip` as extension states.

This should be using `controltar` instead of `control.tar.gz`: https://github.com/jordansissel/fpm/blob/v1.11.0/lib/fpm/package/deb.rb#L800

PR #1552, opened 2018-09-10
FIX: Proper indentation for changes

maveonair:
Set the expected indentation format in changes according to the documentation: https://www.debian.org/doc/debian-policy/ch-source.html#s-dpkgchangelog


jordansissel:
Thank you for this improvement!

Can you clarify something for me -- does this fix a bug? The answer will determine how I write the changelog entry for this change.

jordansissel:
(No tests were broken by this change, and it looks correct, so I merge it)

maveonair:
@jordansissel Thank you very much and here is the explanation

This pull-request fixes the indentation for the changes list according to the expected format. This fix was necessary so that the generated .changes file can be properly read and used by Debian tooling such as `dput`.

PR #1553, opened 2018-09-20
Fix bug #649 - Dependency & Provides issues converting from RPM to DEB.

### LINKED ISSUE #649: rpm to deb result unusable
I'm not a deb user and don't have access to a suitable machine. However, I do have a valid, working RPM which I converted to DEB:

```
 # fpm -t deb -s rpm ZeTool-14.02.24-2.noarch.rpm
no value for epoch is set, defaulting to nil {:level=>:warn}
Downcasing provides 'ZeTool' because deb packages  don't work so good with uppercase names {:level=>:warn}
Debian tools (dpkg/apt) don't do well with packages that use capital letters in the name. In some cases it will automatically downcase them, in others it will not. It is confusing. Best to not use any capital letters at all. I have downcased the package name for you just to be safe. {:oldname=>"ZeTool", :fixedname=>"zetool", :level=>:warn}
Created deb package {:path=>"/var/tmp/zetool_14.02.24-2_all.deb"}
```

When a user tried to install this DEB on Ubuntu 12.04, he reported,

```
 # sudo dpkg -i zetool_14.02.24-2_all.deb 
dpkg: error processing zetool_14.02.24-2_all.deb (--install):
 parse error, in file '/var/lib/dpkg/tmp.ci/control' near line 8 package 'zetool':
 `Depends' field, invalid package name `/bin/bash': must start with an alphanumeric
Errors were encountered while processing:
 zetool_14.02.24-2_all.deb
```

darkelfin:
Fixes #649 
Removes any Dependencies or Provides entries that are invalid when converting from rpm to deb.


lib/fpm/package/deb.rb
```diff
@@ -710,6 +712,18 @@ def fix_dependency(dep)
     end
   end # def fix_dependency
 
+  def validate_dependencies()
+    self.dependencies.each do |dep|
+      if (dep !~ /^[A-Z]/ && dep !~ /^[a-z]/)
+        logger.warn("Dependency '#{dep}' is not a valid deb dependency " \
```
jordansissel: fpm will [convert to lowercase[(https://github.com/jordansissel/fpm/blob/master/lib/fpm/package/deb.rb#L234-L239) the package name if it contains uppercase letters.

I worry that removing the dependency isn't the correct action. Do you have any thoughts on lowercasing the dependency instead?

lib/fpm/package/deb.rb
```diff
@@ -710,6 +712,18 @@ def fix_dependency(dep)
...
+              "name and will be removed.")
+        self.dependencies.delete(dep)
+      end
+    end.flatten
+    # Remove rpmlib dependencies leftover from the rpm.
+    self.dependencies.delete_if { |d| d.start_with?('rpmlib') }
```
jordansissel: This might benefit from being implemented in [converted_from](https://github.com/jordansissel/fpm/blob/master/lib/fpm/package/deb.rb#L234-L239) since it's specific to rpm->deb conversion.

lib/fpm/package/deb.rb
```diff
@@ -724,9 +738,20 @@ def fix_provides(provides)
                    "debs don't like underscores")
       provides = provides.gsub("_", "-")
     end
+    
     return provides.rstrip
   end
 
+  def validate_provides()
+    self.provides.each do |provides|
```
jordansissel: Similar comment to above with 'requires' -- thoughts on lowercasing instead of deleting?

jordansissel:
Thanks for working to improve fpm :)

I left some comments in the change and would like your feedback.

darkelfin:
@jordansissel I've basically moved everything under the converted_from function as you suggested as these are exclusive to rpm -> deb conversion..
The main fix removes things like `/bin/sh` being listed as a dependency (which was the original issue, this is just a bit of a better fix). As it could be difficult to determine the actual package name on different systems.
The removal of the rpmlib dependencies is the same but again moved to the converted_from function.
Look forward to your feedback, thanks!
(sorry for late response, been a little busy)

lib/fpm/package/deb.rb
darkelfin: See comment below.

lib/fpm/package/deb.rb
darkelfin: See comment below.

lib/fpm/package/deb.rb
darkelfin: See comment below.

Issue #1406, opened 2017-09-02
FPM virtualenv packaging not working with Python 3.6


fgimian:
Hi guys, hope you're doing great! 😄 

The FPM virtualenv extension first performs an update and install of both `pip` and `distribute`.  However, installation of `distribute` fails on Python 3.6 (due to a problem in distribute).  To my knowledge, distribute is not actually all that useful anymore (as it is also listed as a legacy package on PyPI).

Here's sample output on CentOS 7 using Python 3.6 from the IUS YUM repository and the latest FPM release:

```
fots > ~ $ fpm --verbose --input-type virtualenv --output-type rpm Flask
Setting workdir {:workdir=>"/tmp", :level=>:info}
Requirement already up-to-date: pip in /tmp/package-virtualenv-build-dead82952af5ad484714f9f4cfb239635b116711834e708f5cedff8eb4d7/usr/share/python/Flask/lib/python3.6/site-packages {:level=>:info}
Collecting distribute {:level=>:info}
  Using cached distribute-0.7.3.zip {:level=>:info}
    Complete output from command python setup.py egg_info: {:level=>:info}
    Traceback (most recent call last): {:level=>:info}
      File "<string>", line 1, in <module> {:level=>:info}
      File "/tmp/pip-build-fx3jcul7/distribute/setuptools/__init__.py", line 2, in <module> {:level=>:info}
        from setuptools.extension import Extension, Library {:level=>:info}
      File "/tmp/pip-build-fx3jcul7/distribute/setuptools/extension.py", line 5, in <module> {:level=>:info}
        from setuptools.dist import _get_unpatched {:level=>:info}
      File "/tmp/pip-build-fx3jcul7/distribute/setuptools/dist.py", line 7, in <module> {:level=>:info}
        from setuptools.command.install import install {:level=>:info}
      File "/tmp/pip-build-fx3jcul7/distribute/setuptools/command/__init__.py", line 8, in <module> {:level=>:info}
        from setuptools.command import install_scripts {:level=>:info}
      File "/tmp/pip-build-fx3jcul7/distribute/setuptools/command/install_scripts.py", line 3, in <module> {:level=>:info}
        from pkg_resources import Distribution, PathMetadata, ensure_directory {:level=>:info}
      File "/tmp/pip-build-fx3jcul7/distribute/pkg_resources.py", line 1518, in <module> {:level=>:info}
[...]

PR #1556, opened 2018-10-03
Fix virtualenv failure on Python 3.6 (#1406)

jordansissel:


jordansissel:
Patch originated from https://github.com/jordansissel/fpm/issues/1406#issuecomment-348349244 and was tested on Python 3.6 and 2.7.15

PR #1558, opened 2018-10-09
[Fixes #1557] Allow git repo as gem source

### LINKED ISSUE #1557: Allow git repo as gem source
It would be useful to allow the source of a ruby gem to be a git repo and/or branch.  I will work on this.

liger1978:


liger1978:
@jordansissel You have any thoughts on this? Keen to get it merged and in a gem release... :)

howardjones:
Is there a reason I'm missing why this needs to be specifically git->gem and not git->*?
(I got here while looking for a recipe for piping `git archive` into fpm instead of exporting to a temporary directory and using `-s dir`.

liger1978:
@howardjones Only that building gems is different from building python modules (which is different from building CPAN modules, etc).

It would be nice to have a "git" source type (that works like "dir" source type) but that is a slightly different use case to this, as no building is required for "dir" source type.

The aim of this is to allow me to fork a gem git repo and build an rpm from that rather than the upstream gem  (some gems need minor tweaks to make them compatible with, for example, CentOS).

jordansissel:
> Is there a reason I'm missing why this needs to be specifically git->gem and not git->*?

I might be able to help clarify.

This PR adds a way to have `-s gem` fetch the gem source from git. At present, `-s gem` can have as inputs:
* a name, in which case the named gem is downloaded from rubygems.org, such as `fpm -s gem -t rpm rails`
* a directory containing a gemspec file (and any necessary files)
* with this pr, a git repository containing a gemspec file (and any other necessary files)

In this PR's scenario, it still takes a `gem` as an input and allows conversion to any output.

This is different than what you describe which could be `-s git` to "source" a package as input from a git repo. In your scenario, this would convert a git repo as input to any target package as output. In this way, a `-s git` would be more similar to `-s dir` where a file listing is used as input without any knowledge of how to build/preprocess those files.

Does this help?

jordansissel:
@liger1978 this PR adds two new flags and I'd like to pick your brain about them: `--gem-git-repo` and `--gem-git-branch`. I was thinking, could we replace the `--gem-git-repo` flag and use the argument(s) to the input() method instead?

What I Mean is having the `--gem-git-repo` option be a boolean `:flag` instead of taking a value. Alternately, if it's possible, we could maybe check if the input parameter is a url, and if a url, is a remote git repo, then automatically assume git. For example:

`fpm -s gem -t rpm https://github.com/jordansissel/fpm`

Now, we could check if the input (https://...) is a url or known git url like git@server:/path, and if so, test if it's a valid git url (`git ls-remote` seems to do this given some light researching).

If it is a valid git url, we could treat it as one and clone locally. Thoughts?

Additionally, I think there should be no default `branch` and we can let `git `decide which is the correct default branch. If the user specifies a branch, we can checkout after cloning as you do now.

My proposal roughly would be:
* Replace the `--gem-git-repo` flag with some automatic heuristics to detect if any input (`fpm -s gem -t rpm input1 input2 ...`) are valid git urls.
* Let git decide the default branch to checkout - keep `--gem-git-branch` but default it to `nil` and only `g.checkout(...)` if the flag value is non-nil

@liger1978 thoughts? :)

liger1978:
@jordansissel 

1. I had considered the idea of parsing the input (name) argument to determine of it was a git url. I ruled it out because often the repo name does not match the gem name (e,g, often the repo is prefixed 'ruby-').  In rare cases there may even be more than one gemspec file in a repo. Thus the name argument is still required in addition to the git url to determine what to build.  Thoughts?

2. I agree with your idea of not having "master" as a default for the branch. In rare cases "master" is not the default branch. I will change this.

howardjones:
Apologies, I'm not familiar enough with the source to follow the patch properly. I'll be quiet now :-)

liger1978:
@howardjones No need to be quiet or apologise :)  Your idea is a good one as an additional feature.  You should open a dedicated issue for it.

jordansissel:
@liger1978 thanks for your response. I'm +1 this patch after the git branch default is changed.

I did some thinking about the automatic url thing we discussed, and I agree with you that being able to choose the gemspec name would be useful in the (rare-ish?) cases that a repo has multiple gemspecs.

liger1978:
@jordansissel Done.

liger1978:
@jordansissel Are you OK to merge this now I have made the changes?

jordansissel:
I’ll try to make time to review and test in the next few days. I don’t
anticipate any issues and will merge after that :)

liger1978:
@jordansissel Any news on this? :)

jordansissel:
Apologies for the delays. I’m still on parental leave and generally
spending my days away from devices which would allow for further review.
I’ll queue this up so I don’t forget when I return. (In a few days or
weeks, unsure yet)

liger1978:
@jordansissel No worries. Enjoy! :)

liger1978:
@jordansissel  I'd still like this merged if you have more time to review. No worries if you haven't.

liger1978:
@jordansissel Any chance of getting this merged? Thanks!

lib/fpm/package/gem.rb
```diff
@@ -91,21 +101,33 @@ def download(gem_name, gem_version=nil)
...
-
     download_dir = build_path(gem_name)
     FileUtils.mkdir(download_dir) unless File.directory?(download_dir)
 
-    ::Dir.chdir(download_dir) do |dir|
-      logger.debug("Downloading in directory #{dir}")
-      safesystem(*gem_fetch)
+    if attributes[:gem_git_repo]
+      logger.debug("Git cloning in directory #{download_dir}")
+      g = Git.clone(attributes[:gem_git_repo],gem_name,:path => download_dir)
```
CameronNemo, line 109: @jordansissel I am not much of a rubyist. Would I just need to move the `require "git"` line to the top of this block for #1748 ?

PR #1562, opened 2018-10-17
Prevent MITM attacks

Al2Klimov:


Al2Klimov:
*Sigh* Just wanted to change a few lines of doc and all tests fail...

jordansissel:
> Sigh Just wanted to change a few lines of doc and all tests fail...

For what it's worth, it's common for some projects to go through periods of time where some set of tests are either flakey or broken. Nobody's made time to resolve these failures, so they remain failing :)

PR #1576, opened 2018-12-11
added support for RPM ghost files

ruedigerblock:
Moin !
I needed fpm to be able to add ghost files in RPMs. See: [this link](http://ftp.rpm.org/max-rpm/s1-rpm-inside-files-list-directives.html).
Did not had to change a lot. I basically copied the "--rpm-tags" - code.

Regards,
Rüdiger

### DIFF
--- a/lib/fpm/package/rpm.rb
+++ b/lib/fpm/package/rpm.rb
@@ -155,6 +155,14 @@ class FPM::Package::RPM < FPM::Package
     File.expand_path(val) # Get the full path to the script
   end # --posttrans
 
+  rpm_ghost_files = []
+  option "--ghost-file", "FILE",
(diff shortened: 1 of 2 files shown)


jpchavat:
It would be really useful to have the %ghost tag in FPM. Anyone found a workaround for that? or is everybody waiting for the PR?

jpchavat:
> It would be very useful to have the %ghost tag in FPM. Has anyone found a workaround for that, or is everyone waiting for the PR?

It won't be nought just with the %ghost tag, according with the follow information:
https://fedoraproject.org/wiki/PackagingDrafts/Logfiles

Paradox98:
It seems that dhost is still not supported by fpm currently. Is there any other way to avoid the error that the file to delete cannot be found when rpm is uninstalled.

PR #1588, opened 2019-01-17
Add Singularity definition file

jmdf:
[Singularity](https://github.com/sylabs/singularity) is an opensource container platform widely used in EPC and HPC environments. 
I've been using it for some time already and becomes very interesting when someone needs to create packages in multiple different systems, without needing to install all dependencies every time.

This PR adds a definition file that can be used to build the container.

### DIFF
--- a/singularity.def
+++ b/singularity.def
@@ -1,0 +1,26 @@
+# Build command:
+#
+#   singularity build fpm.simg singularity.def
+#
+# Execute instructions:
+#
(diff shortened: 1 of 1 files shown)


jordansissel:
I confess to having no experience with Singularity, so I can't adequately review this. However, because this appears to be a change that cannot introduce any breaking changes to fpm itself, I am in favor of merging.

LGTM :)

Issue #1592, opened 2019-01-30
Unable to install fpm due to childprocess issue


fatmcgav:
Attempting to install FPM today fails with:
```
Fetching: childprocess-1.0.0.gem (100%)
Building native extensions.  This could take a while...
ERROR:  Error installing fpm:
        ERROR: Failed to build gem native extension.

    current directory: /var/lib/gems/2.3.0/gems/childprocess-1.0.0/ext
/usr/bin/ruby2.3 mkrf_conf.rb

current directory: /var/lib/gems/2.3.0/gems/childprocess-1.0.0/ext
/usr/bin/ruby2.3 -rubygems /usr/share/rubygems-integration/all/gems/rake-10.5.0/bin/rake RUBYARCHDIR=/var/lib/gems/2.3.0/extensions/x86_64-linux/2.3.0/childprocess-1.0.0 RUBYLIBDIR=/var/lib/gems/2.3.0/extensions/x86_64-linux/2.3.0/childprocess-1.0.0
/usr/bin/ruby2.3: No such file or directory -- /usr/share/rubygems-integration/all/gems/rake-10.5.0/bin/rake (LoadError)

rake failed, exit code 1

Gem files will remain installed in /var/lib/gems/2.3.0/gems/childprocess-1.0.0 for inspection.
Results logged to /var/lib/gems/2.3.0/extensions/x86_64-linux/2.3.0/childprocess-1.0.0/gem_make.out
```

This appears to be related to a new version of `childprocess` released today. 

Raised upstream as https://github.com/enkessler/childprocess/issues/143

benma:
If possible, the dependency versions should be fixed, or restricted to the major version (dependencies bumping a major version should be reviewed manually).

jordansissel:
I'm working on this now to pin a version.

andrewkroh:
> fpm 1.11.0 released.

This is possibly still in progress, but it looks like you normally git tag upon release. I noticed there isn't a 1.11 tag or changelog entry.

jordansissel:
Oops! I had limited time yesterday and forgot to tag, but I do remember
updating the changelog. I’ll work on that today.

liger1978:
[...]

PR #1595, opened 2019-02-03
Less strict pinning for childprocess #1592

liger1978:


jordansissel:
LGTM as discussed in https://github.com/jordansissel/fpm/issues/1592#issuecomment-459846573

PR #1596, opened 2019-02-03
Attempt to fix failing rspecs.

jordansissel:
```
% bundle exec rspec
........................................................................................................................................................
................................................................................................................................

Finished in 22.03 seconds (files took 0.32223 seconds to load)
280 examples, 0 failures
```

It's been a long time since 100% of the specs have passed locally on my workstation. ;)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ rvm:
   - 2.3.3
   - 2.4.0
   - 2.5.1
+  - 2.6.0
 matrix:
   exclude:
     - os: osx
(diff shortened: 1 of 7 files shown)


jordansissel:
There's only so much energy I have for black-box testing Travis's build environment.

At present,
* Linux Ruby 1.9 and 2.0 fail because of Snap needing Psych::safe_load which didn't exist until Ruby 2.1
* Linux Ruby 2.1+ are succeeding now (and were not prior to this PR)
* Mac build failures are all over the place with weird environmental problems (can't find `virtualenv`, or `bundler` gem not found)

PR #1598, opened 2019-02-05
Updated docs  to reflect Oracle Enteprise Linux instructions

syedaali:
Updated docs/installing.rst to reflect Oracle Enterprise Linux instructions since you have to install an Oracle repo before you can install ruby-devel.


jordansissel:
Cheers :)

Issue #1599, opened 2019-02-05
loosen json dependency to accept > 1.7.7 up to < 3.0


jsvd:
Looking at the [breaking changes in json 2](https://github.com/flori/json/blob/master/CHANGES.md#2015-09-11-200) it seems that we can allow fpm to depend on json `["> 1.7", "< 3.0"]` which would make fpm work on rubies from 1.8 to 2.6.

jordansissel:
#1600 merged. I'll work on a new release sometime today.

xli1996:
@jordansissel Hi, I wonder if it possible to release a new version since JSON has a cve(https://nvd.nist.gov/vuln/detail/CVE-2020-10663). It would be good to release a new version with json dependency up to 3.0. Many thanks.

PR #1600, opened 2019-02-06
relax json dependency

yaauie:
rubies >= 2.5 ship with JSON 2.x, so allowing bundler to resolve to newer
implementations eliminates conflicts with dependency trees that include JSON
2.x.

Breaking changes for JSON 2.x include removed support for older rubies [[1]],
so specifying in this manner ensures that older rubies can find 1.x, while
newer ones can resolve to 2.x

[1]: https://github.com/flori/json/blob/master/CHANGES.md#2015-09-11-200

Resolves: https://github.com/jordansissel/fpm/issues/1599

### DIFF
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
 
   # For parsing JSON (required for some Python support, etc)
   # http://flori.github.com/json/doc/index.html
-  spec.add_dependency("json", ">= 1.7.7", "< 2.0") # license: Ruby License
+  spec.add_dependency("json", ">= 1.7.7", "< 3.0") # license: Ruby License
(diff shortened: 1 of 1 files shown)


jordansissel:
> rubies >= 2.5 ship with JSON 2.x, so allowing bundler to resolve to newer

Interesting. No conflict when I use ruby 2.6.0:

```
% ruby --version
ruby 2.6.0p0 (2018-12-25 revision 66547) [x86_64-linux]
% bundle install | grep json
Using json 1.8.6
```

jordansissel:
Testing:

```
% bundle install
Using json 2.1.0 (was 1.8.6)
...
```

No new test failures from this. I'm good with merging. :+1:

PR #1528, opened 2018-07-10
Perl CPAN Support, Fix All 5 Outstanding Issues, Second Try

wbraswell:
I tried this pull request once before, but forgot to clean up the code in the last commit: https://github.com/jordansissel/fpm/pull/1527

@jordansissel
Please merge this pull request, it solves all 5 outstanding Perl issues:

Issue:  https://github.com/jordansissel/fpm/issues/1517
Fixed:  https://github.com/wbraswell/fpm/commit/a1afdb743b818d98a0453bb0213818a8ddb60484

Issue:  https://github.com/jordansissel/fpm/issues/1519
Fixed:  
https://github.com/wbraswell/fpm/commit/4756a450638f0536fcfbf2fa41bff6affa15900d
https://github.com/wbraswell/fpm/commit/373cb6572182c4cc3753d17d61ec72e566d23a79

Issue:  https://github.com/jordansissel/fpm/issues/1522
Fixed:
https://github.com/wbraswell/fpm/commit/3171ee9f834ff5e1f6af64b869a67227574689d8
https://github.com/wbraswell/fpm/commit/985c39bb23ce4457c3ca2892c39aa05a097f6d2c

Issue:  https://github.com/jordansissel/fpm/issues/1523
Fixed:  https://github.com/wbraswell/fpm/commit/f4f4b563d0b4826ca0bbd31fb8c2afa25e09c74e

Issue:  https://github.com/jordansissel/fpm/issues/1526
Fixed:  
https://github.com/wbraswell/fpm/commit/c01f67a07bacbe7d31aaf94274137514cd60146e
https://github.com/wbraswell/fpm/commit/1d37ec3933dd7802d1bbbc907a5e33e5ff1576cd

Issue #1523, opened 2018-07-04
Perl CPAN Support, Missing 'author' Field


wbraswell:
When attempting to package Class::Data::Inheritable (or presumably any CPAN distribution with blank 'author' field), we get a crash when we should just handle it and move forward.

```
$ fpm --no-cpan-test --cpan-verbose --verbose --debug-workspace --workdir /home/wbraswell/fpm_tmp/ -s cpan -t rpm Class::Data::Inheritable
...
SOME OUTPUT
...
Invalid package configuration: Unexpected CPAN 'author' field type: NilClass. This is a bug.
```

The correct solution is to note that no vendor or author has been provided, and continue with the packaging process.  I have done this in my latest commit here:
https://github.com/wbraswell/fpm/commit/f4f4b563d0b4826ca0bbd31fb8c2afa25e09c74e

wbraswell:
@liger1978 
Found another bug!  LOL

wbraswell:
This issue has been solved by @NicholasBHubbard in the following two pull requests:
https://github.com/jordansissel/fpm/pull/1937
https://github.com/jordansissel/fpm/pull/1942

Issue #1522, opened 2018-07-04
Perl CPAN Support, Frozen cpanm Command, Broken I/O Handling In execmd()


wbraswell:
@liger1978 
Discussion of this issue started here:   https://github.com/jordansissel/fpm/issues/1519

Specifically, these two comments:
https://github.com/jordansissel/fpm/issues/1519#issuecomment-402541298
https://github.com/jordansissel/fpm/issues/1519#issuecomment-402543327

Essentially, this issue causes 2 different negative situations.

First, we can not see any of the output of the `cpanm` commands, including the critically important configure & build processes such as `Makefile.PL` and `Build.PL` and `make` and `gcc` and `as` etc etc.  If I don't know what is happening with `cpanm` then I really have no idea what is happening at all.

Second, we experience an actual freeze/hang during one of the `gcc` and/or `as` commands, as shown in the comments linked above.

I have tracked this problem to one line:
https://github.com/wbraswell/fpm/blob/master/lib/fpm/util.rb#L165
```
if block_given?
```

If I simply replace this line with `if false` then both parts of this issue appear to be solved!  I can now see all the output generated in real time, and the mysterious hanging gcc/as processes no longer hang!  Unfortunately, by doing so I also re-introduce the previous bug 1519, because we need an empty STDIN to be present for any possible interactive processes:
https://github.com/jordansissel/fpm/issues/1519

So, in order to achieve a hybrid solution I actually have to disable all the STDOUT and STDERR parts, while leaving the STDIN parts in tact.  I have done this in my latest commit:
https://github.com/wbraswell/fpm/commit/3171ee9f834ff5e1f6af64b869a67227574689d8

Since I am not familiar with the `execmd()` subroutine and how Ruby handles all the I/O magic, I will not presume that my temporary solution is correct enough to create a pull request.  In fact, I assume this would be a not-entirely-correct solution, and that we need to somehow figure out why the hanging is actually happening in order to create a permanent solution.

liger1978:
I'm not seeing this other issue (CentOS 7).  I've pushed again to https://github.com/liger1978/fpm/tree/1519 which adds the extra required `safesystemin()` calls in `cpan.rb`.

wbraswell:
@liger1978 
This is the command which freezes w/out my hybrid solution:
```
$ fpm --no-cpan-test --cpan-verbose --verbose --debug-workspace --workdir /home/wbraswell/rperl_packager_tmp/ -t rpm -s cpan RPerl
[...]

Issue #1519, opened 2018-07-04
Perl CPAN Support, Frozen cpanm Command, Interactive Makefile.PL


wbraswell:
When I try to run fpm for the Inline::CPP software on CPAN, it gets partway through the process and then hangs.  If you just look at the output, you will think it is waiting for info about Parse::RecDescent, but I put a bunch of checkpoints in the code and it is actually getting well past that.

```
$ fpm --no-cpan-test --cpan-verbose --verbose --debug-workspace --workdir /home/wbraswell/rperl_packager_tmp/ -t rpm -s cpan Inline::CPP
...
LOTS OF OUTPUT HERE
...
Installing /home/wbraswell/rperl_packager_tmp/package-cpan-build-2cb511583273a27fa85f246094e26472cffc43fbd51a5581c969ecf4aa93/cpan/lib/perl5/Test2/Formatter/TAP.pm {:level=>:info}
Appending installation info to /home/wbraswell/rperl_packager_tmp/package-cpan-build-2cb511583273a27fa85f246094e26472cffc43fbd51a5581c969ecf4aa93/cpan/lib/perl5/x86_64-linux-thread-multi/perllocal.pod {:level=>:info}
OK {:level=>:info}
Successfully installed Test-Simple-1.302136 (upgraded from 0.98) {:level=>:info}
Installing /home/wbraswell/rperl_packager_tmp/package-cpan-build-2cb511583273a27fa85f246094e26472cffc43fbd51a5581c969ecf4aa93/cpan/lib/perl5/x86_64-linux-thread-multi/.meta/Test-Simple-1.302136/install.json {:level=>:info}
Installing /home/wbraswell/rperl_packager_tmp/package-cpan-build-2cb511583273a27fa85f246094e26472cffc43fbd51a5581c969ecf4aa93/cpan/lib/perl5/x86_64-linux-thread-multi/.meta/Test-Simple-1.302136/MYMETA.json {:level=>:info}
<== Installed dependencies for /home/wbraswell/rperl_packager_tmp/package-cpan-build-2cb511583273a27fa85f246094e26472cffc43fbd51a5581c969ecf4aa93/module. Finishing. {:level=>:info}
9 distributions installed {:level=>:info}
Asking metacpan about a module {:module=>"Carp", :level=>:info}
Asking metacpan about a module {:module=>"Inline", :level=>:info}
Asking metacpan about a module {:module=>"Inline::C", :level=>:info}
[...]

PR #1610, opened 2019-02-23
Fix a few CPAN issues

jordansissel:
This builds off of the excellent work done by @wbraswell in #1528.

This PR should fix:
* #1519, #1522 (`fpm -s cpan -t rpm Inline::CPP` hangs on stdin)
* #1517 (Alien::astyle fails due to a bug in FPM's `tar` invocation)
* #1523 (Allow blank author field in CPAN metadata)

Because there are known test failures on master, testing on master vs this PR locally shows:
* master: 13 failures
* this pr: 13 failures

For this, I believe nothing broke in this PR, but the above issues should be fixed. When I test the known cases, they work for me; specifically, the following all fail on master but succeed on this PR.

* `bin/fpm --debug -s cpan -t rpm --no-cpan-test Inline::CPP`
* `bin/fpm --verbose --no-cpan-test -s cpan -t rpm Alien::astyle`
* `bin/fpm --verbose --no-cpan-test -s cpan -t rpm Class::Data::Inheritable`

### DIFF
--- a/lib/fpm/package/cpan.rb
+++ b/lib/fpm/package/cpan.rb
@@ -112,6 +112,8 @@ class FPM::Package::CPAN < FPM::Package
     self.vendor = case metadata["author"]
       when String; metadata["author"]
       when Array; metadata["author"].join(", ")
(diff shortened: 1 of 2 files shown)


jordansissel:
Closing based on the work done recently by @NicholasBHubbard and @wbraswell (#1944, #1955, etc)

Issue #1472, opened 2018-03-09
--deb-after-purge Does not Write to postrm File


brianraymes:
It appears that the **--deb-after-purge** flag does not work.

The following always works as expected by including the contents of **remove-file** within **postrm**'s **after_remove()** method when creating a debian package:

```bash
fpm \
  ... \
  --after-remove remove-file \
  ...
```

The following does not work as expected:

```bash
fpm \
  ... \
  --deb-after-purge purge-file \
  ...
```

I would expect that the contents of **purge-file** would end up within **postrm**'s **after_purge()** method when used.

Also, if one only provides the **--deb-after-purge** flag, the **postrm** file is not created and does not exist in the completed package.

evymaphia:
+1
fpm version 1.10.2 has still the same issue

Jamesits:
can reproduce this bug on fpm 1.11.0

Z9n2JktHlZDmlhSvqc9X2MmL3BwQG7tk:
**--deb-after-purge** does not work in fpm-1.15.1 (released Feb 1, 2023) too.
Bug is ~5 years old (!!!). What is the problem to fix it ?

Qwarctick:
Hello @jordansissel
https://github.com/jordansissel/fpm/pull/1620 seems a good fix for this issue? No?

What can we do more to close this issue?

sikmir:
fpm version 1.17.0 has still the same issue

PR #1620, opened 2019-03-20
Fixing deb_purge_script

diorcety:
- deb_after_purge script was never read from disk
- since after_remove and after_purge write to the same location,
  we only want to write this out once.
- Replace #1579


polyedre:
Hi @jordansissel, I'm interested in this getting merged. What is necessary for this to be merged ? It seems that the original PR #1579 was first submitted in 2018. The issue #1472 is still open too. I would be happy to help testing or proposing a MR that passes the tests.

PR #1623, opened 2019-04-01
fix bug on deb compression type for control

asouini:
even if deb compression flag is set to another supported compression type, it is always compressed as **control.tar.gz**

this is causing problems on some artifact indexing systems like **artifactory** where it is expecting bionic packages to have both control and data be compressed with **xz**

~~PS : i do no know ruby so that return value i added in **write_control_tarball** could be wrong but i needed the variable outside the function~~ tested the patch localy and it works, cant wait for the next release :)


jordansissel:

PR #1646, opened 2019-06-18
updating to alpine3.9 - ruby 2.5.5

double-a-io:


BartKoppers (commented):
Too dated - stable Alpine is 3.12

gliptak:
is this still target? [Dockerfile](https://github.com/jordansissel/fpm/blob/main/Dockerfile#L20) now `Ubuntu` based

PR #1652, opened 2019-07-16
Fix building on OSX where File.link is not working as expected. Closes #1651.

### LINKED ISSUE #1651: On OSX File.link is creating an invalid hardlink
Replace using the `File.link` with copying rather if the os is darwin based.

Skarlso:


blkperl:
@jordansissel do you have any insight into why the tests failed?

Skarlso:
Mostly, I think, they are just outdated at some point and need a bit of love.

PR #1655, opened 2019-07-23
Allow FPM to be loaded from a environment variable. 

cwood:
I have a use case where I want to do a monorepo and have separate subdirectories. 

An example:
a/.fpm
b/.fpm
.fpm

I want to be able to have a global .fpm for my monorepo and have each service specify there own .fpms to be added.  

I expect the run to look like `FPMRC=a fpm ...`

Hopefully, this will resolve that. Let me know what else needs to be done.

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -524,7 +524,7 @@ class FPM::Command < Clamp::Command
   def run(run_args)
     logger.subscribe(STDOUT)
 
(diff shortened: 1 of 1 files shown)


BartKoppers:
Nice idea! However, what's the addition here, besides looping through the directories? 
Maybe you can provide a somewhat full(er) example FPMRC-file?

PR #1658, opened 2019-08-03
Fixed parameter to gem install

henne90gen:
The two options that were there before do no longer work.
The new option is called '--no-document'


jordansissel:
#1667 also fixed this. Thank you for improving fpm! :)

Issue #1693, opened 2020-03-31
deb-postinst: implement triggered


root360-AndreasUlm:
Hi,

as a debian package can install triggers (1) to allow running commands when other packages requires them FPM should implement a mechanism to allow adding those.
IMHO it can be the same as after-install because triggers must be defined in postinst too.

I'm working on this one and will create a pull request soon.

Regards,
Andreas

(1) https://wiki.debian.org/DpkgTriggers

matikij:
I have also run into that. So in other words: bump

What I am currently doing is a nasty workaround, in the `--after-install` script I am injecting an if clause:
```
echo "This is after_install()"
after_install_continued
} # this closes the after_install() definition

triggered() {
  echo "Triggered by trigger ${1}"
}

if [ "${1}" = "triggered" ]; then
  triggered ${2}
fi

after_install_continued(){
  # empty command, just in case
  : 
```

This results the following to be rendered in the `postinst` Debian maintainer script:
```
# [...]
after_install() {
    :
echo "This is after_install()"
after_install_continued
} # this closes the after_install() definition

triggered() {
  echo "Triggered by trigger ${1}"
}

if [ "${1}" = "triggered" ]; then
  triggered ${2}
fi

# The contents of after_install from the template continue here
after_install_continued(){
 :


}

if [ "${1}" = "configure" -a -z "${2}" ] || \
# [...]
```

This results in the following output when triggering on `/bin/producer_script`:
```
# dpkg -i producer_1.deb 
(Reading database ... 37307 files and directories currently installed.)
Preparing to unpack producer_1.deb ...
Unpacking producer (1) over (1) ...
Setting up producer (1) ...
Processing triggers for consumer (1) ...
Triggered by trigger /bin/producer_script
```

I know this is not pretty, this is precisely the patch proposed by @root360-AndreasUlm should be accepted :-)

PR #1694, opened 2020-03-31
implemented triggered & fixed deb_systemd_enable #1693

root360-AndreasUlm:
* implemented option triggered to define dpkg trigger script
* fixed postinst script in case deb_systemd is true and
deb_systemd_enable is false as it failed shellcheck

Signed-off-by: Andreas Ulm <andreas.ulm@root360.de>

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -212,6 +212,12 @@ class FPM::Command < Clamp::Command
         "Currently only supports deb, rpm and pacman packages." do |val|
     File.expand_path(val) # Get the full path to the script
   end # --before-upgrade
+  option "--triggered", "FILE",
+    "A script snippets to define and run package triggers, see\n" \
+    "https://wiki.debian.org/DpkgTriggers and\n" \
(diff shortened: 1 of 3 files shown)


lib/fpm/command.rb
```diff
@@ -212,6 +212,12 @@ def help(*args)
         "Currently only supports deb, rpm and pacman packages." do |val|
     File.expand_path(val) # Get the full path to the script
   end # --before-upgrade
+  option "--triggered", "FILE",
```
jordansissel, line 215: The trigger support here seems to be targeted specifically at debian packages, so it feels best to move this option to the deb target instead of the global command line flags. Thoughts?

jordansissel (changes requested):

lib/fpm/command.rb
root360-AndreasUlm: rpm implements triggers too (http://ftp.rpm.org/api/4.4.2.2/triggers.html) but they are different.
Thus I'll change it to `--deb-triggered`.

Issue #1696, opened 2020-03-31
fix maintainer-script-ignores-errors


root360-AndreasUlm:
Hi,

although 'maintainer-script-ignores-errors' is only a warning in lintian the effect can be vital.
If an important command fails within a bad written script the whole maintainer script can do a lot of harm.

See this sample scenario:

Assume the --after-install script looks like this:
```
mkdir /tmp/test
cd /tmp/test
[generate temporary files and directories]
mv ./* /usr/share/<package>/
rm -rf ./*
```

When the creation of `/tmp/test` works everything is working as intended.
But when the creation fails e.g. because `/tmp/test` is a file the change of directory fails and every following command is run in `/` thus creating random stuff there and at last removing all files and directories.
This can be prevented by setting the error exit flag of the shell.

That is why I'd like the flag to be added to the maintainer scripts.

I'll create a pull request for that soon.

Regards,
Andreas

BartKoppers:
A maintainer could also decide to include set -e in the script itself.

It's a simple check to test if /tmp/test does exist and is a directory. 
If this is important, the shellscript should solve this. And IMHO not fpm.

But, although I think this doesn't really solve these initial problems, it doesn't 'hurt' either.
Well,  *at least where /bin/sh is linked to Bash* (on Mac Catalina it isn't anymore). So you could (should?) check this.

root360-AndreasUlm:
> A maintainer could also decide to include set -e in the script itself.

Yes a maintainer could but he/*/she cannot set it globally as the scripts provided via the cli-parameters are merged into a complete script based on skeletons defined by fpm.
Thus the maintainer can never solve the lintian error [maintainer-script-ignores-errors](https://lintian.debian.org/tags/maintainer-script-ignores-errors.html) on his own.

> It's a simple check to test if /tmp/test does exist and is a directory.
> If this is important, the shellscript should solve this. And IMHO not fpm.

The sample is just a simple one to show the point I'm talking about.

> Well, _at least where /bin/sh is linked to Bash_ (on Mac Catalina it isn't anymore). So you could (should?) check this.

The shell option 'errexit' which is activated when `set -e` is executed is supported by all modern POSIX-compliant shells [1] and is not a Bash-only feature.
Debian uses since a long time Dash as default shell.

[1]:
* zsh https://linux.die.net/man/1/zshoptions (option ERR_EXIT)
* dash https://linux.die.net/man/1/dash (option errexit)
[...]

PR #1697, opened 2020-03-31
added set -e to debian maintainer scripts #1696

root360-AndreasUlm:
Signed-off-by: Andreas Ulm <andreas.ulm@root360.de>

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -193,6 +193,11 @@ class FPM::Package::Deb < FPM::Package
     File.expand_path(val) # Get the full path to the script
   end # --after-purge
 
+  option "--maintainerscripts-force-errorchecks", :flag ,
+    "Activate errexit shell option according to lintian. " \
+    "https://lintian.debian.org/tags/maintainer-script-ignores-errors.html",
+    :default => false
+
   def initialize(*args)
     super(*args)
     attributes[:deb_priority] = "extra"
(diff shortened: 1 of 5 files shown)


BartKoppers (commented):
See [Why doesn't set -e (or set -o errexit, or trap ERR) do what I expected?](http://mywiki.wooledge.org/BashFAQ/105). Also set -e is bash-specific. 
If accepted then at least a check is needed if  /bin/sh is linked to Bash. 
Or is /bin/sh on Debian *guaranteed* to be bash?

root360-AndreasUlm:
> See [Why doesn't set -e (or set -o errexit, or trap ERR) do what I expected?](http://mywiki.wooledge.org/BashFAQ/105). Also set -e is bash-specific.
> If accepted then at least a check is needed if /bin/sh is linked to Bash.
> Or is /bin/sh on Debian _guaranteed_ to be bash?

@BartKoppers 
The shell option 'errexit' which is activated when `set -e`is executed is supported by all modern POSIX-compliant shells [1] and is not a Bash-only feature.
Debian uses sind a long time Dash as default shell.

As this PR changes only scripts used to create Debian Packages and Debian will always use POSIX-compliant shells there is no check necessary.

[1]:
* zsh https://linux.die.net/man/1/zshoptions (option ERR_EXIT)
* dash https://linux.die.net/man/1/dash (option errexit)
* csh https://linux.die.net/man/1/csh (option -e)
* tcsh https://linux.die.net/man/1/tcsh (option -e)
* ksh https://linux.die.net/man/1/ksh (option -e)
* etc.

BartKoppers:
Discussion in issue #1696

BartKoppers (approved):

jordansissel:
Thanks for the thoughtful discussion and improvement here. <3

PR #1699, opened 2020-04-08
deb: Fixes gem pessimistic version constrains for single component versions

vholer:
If gem had a version constraint with just major version (e.g., ~>1),
the Debian packages were generated with wrong zero upper limit
(e.g., >= 1 and << 0). This results in unresolvable dependencies.


jordansissel:
Thanks for this

PR #1706, opened 2020-05-15
Add option to disable remove_python_compiled_files from virtualenv

combsjy:
Prior to this change if an egg built with `--exclude-source-files` is used, it will result in a broken environment.

### DIFF
--- a/lib/fpm/package/virtualenv.rb
+++ b/lib/fpm/package/virtualenv.rb
@@ -46,6 +46,8 @@ class FPM::Package::Virtualenv < FPM::Package
     :multivalued => true, :attribute_name => :virtualenv_find_links_urls,
     :default => nil
 
+  option "--binary-removal", :flag, "Prune virtualenv of pyc/pyo", :default => true
+
   private
 
   # Input a package.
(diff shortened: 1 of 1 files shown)


BartKoppers:
> Prior to this change if an egg built with `--exclude-source-files` is used, it will result in a broken environment.

Could you clarify what the bug was that caused the environment to brake?

combsjy:
When a python egg is built with `--exclude-source-files` the only files left in the egg are python byte codes files or `.pyc` files.  

Without this option, when pulled into a virtualenv by fpm, it will have its `.pyc` files striped.  This leaves the package empty in the virtualenv.

BartKoppers:
> When a python egg is built with `--exclude-source-files` the only files left in the egg are python byte codes files or `.pyc` files.
> Without this option, when pulled into a virtualenv by fpm, it will have its `.pyc` files striped.

Ok, AFAIU python eggs are not meant *at all* to be importable, and deprecated. Wheel is intended to import binary. 
See [Wheel vs Egg](https://www.python.org/dev/peps/pep-0427/#comparison-to-egg)

Sounds to me as a rather specific use-case?

Issue #1708, opened 2020-06-03
Latest ffi gem requires ruby >= 2.3, so gem install fpm no longer works on centos 7


fidothe:
Assuming 2.0 is the minimum required version, this just needs a simple tweak to the gemspec. I will open a PR

jameskaf:
Any workaround?

rconde01:
here's a workaround:
```
yum install git
yum install rh-ruby23 rh-ruby23-ruby-devel
scl enable rh-ruby23 'gem install fpm'
scl enable rh-ruby23 'fpm --version'
```

PR #1709, opened 2020-06-03
Pin ffi to 1.12.x, for compatibility with Ruby 2.0

fidothe:
ffi 1.13 requires Ruby 2.3, which is not supplied by many still-supported distros (e.g Centos/RHEL 7), 1.12.X works there, so restrict to that.

Addresses #1708

### DIFF
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -47,7 +47,9 @@ Gem::Specification.new do |spec|
   spec.add_dependency("childprocess", "< 1.0.0") # license: ???
 
   # For calling functions in dynamic libraries
-  spec.add_dependency("ffi") # license: GPL3/LGPL3
+  # This is pinned because 1.13.0 requires Ruby 2.3 or later
+  # Ref: https://github.com/jordansissel/fpm/issues/1708
+  spec.add_dependency("ffi", "~> 1.12.0") # license: GPL3/LGPL3
 
   spec.add_development_dependency("rake", "~> 10") # license: MIT
 


luto:
This would be great to have to fix fpm on CentOS 7 :)

jordansissel:
I agree this fix would be helpful for folks stuck on older/abandoned versions of Ruby (v2.0 was EOL 4 years ago). I am in favor of these kinds of fixes!

I need help testing this change on other platforms to understand any possible negative impacts of this change.

joespitch:
FYI I also just encountered this error on CentOS 7, but was able to work around it by just installing RVM and setting the default version to 2.3.3:

- https://www.techoism.com/install-ruby-2-3-3-centosrhel-rvm/

after that, the FPM installation worked as normal.

lleroyconsult:
Any news on this PR ?

sajjad037:
I had faced a similar error on `Oracle Linux 7.x` and the workaround that I followed to fix it was, after installing the [prerequisites](https://fpm.readthedocs.io/en/latest/installing.html#installing-things-fpm-needs)  packages. Install the `ffi 1.9.14` package First and then install the `fpm`.

```
gem install ffi -v '1.9.14'
# then installed fpm
gem install --no-ri --no-rdoc fpm
fpm --version
Doing `require 'backports'` is deprecated and will not load any backport in the next major release.
Require just the needed backports instead, or 'backports/latest'.
1.11.0
```

jordansissel:
I am OK with this change, though I don't know what negative impact this will have on newer rubies, if any. I briefly reviewed the changelog for ruby ffi and 1.12.2 is a year old at this point which seems recent enough. Seems safe to make this change.

felixbuenemann:
@jordansissel This change breaks compatibility with arm64 based Macs, since they require ffi 1.14.0 or newer.

Can't systems with outdated Rubies just install an older version of the ffi gem manually?

luto:
@felixbuenemann Installing an older ffi version prior to fpm works fine as documented in https://github.com/jordansissel/fpm/pull/1709#issuecomment-747533986, yes. We've been doing that for months.

fosskers:
This seems to prevent the [fpm package on Arch](https://aur.archlinux.org/packages/fpm) from being updated, since only `ruby-ffi-1.15.0` is available. I will do some experiementation.

fosskers:
Got through it by patching the `ffi` bounds in the `PKGBUILD`:
```
  sed -i 's/"ffi", "~> 1.12.0"/"ffi", ">= 1.15.0"/' fpm.gemspec
```

felixbuenemann:
I'm voting to revert this change.

Breaking newer Rubies to support an outdated Ruby version that is EOL a long time is not a good idea…

Users of such old Rubies can install a compatible version of the ffi gem by pinning to an old version using bundler, as long as the gemspec does not exclude those versions.

jordansissel:
I’m not sure yet what good solution might exist for this problem.

On one hand, there are many systems which cannot reasonably upgrade their
version of Ruby, and “gem install fpm” will fail if we don’t pin ffi. It
doesn’t matter much to me that they’re EOL — they’re still in production.

The other hand, newer rubies are broken when we pin.

It’s not great to choose which large population of users to hurt.

I’m open to ideas. Maybe we can fix this somehow in the documentation to
express how older Ruby users can successfully use fpm when ffi is unpinned?

fosskers:
For sure this is a problem that many open source projects have to come to a fair conclusion on. Namely:

> When do we leave (real or hypothetical) users behind?

or more technically:

> What should `fpm`'s support window be?

The latter is clearer, as `fpm` already has an implicit support window. Surely Ruby 1.0 users aren't being considered here, although _perhaps_ there are a few still lurking around, who could potentially want `fpm`? We typically can't support an infinitely wide backcompat window, as after a while projects grind to a halt under the weight of the workarounds that are often necessary.

So we're caught between two poles: the necessity of keeping up with the evolving ecosystem (or we stagnate) and the necessity of considering users who for reasons possibility out of their control cannot keep up with us. So now we're just negotiating: what's a reasonable support window for `fpm`?

jordansissel:
To give you an idea of where I typically stand on support windows, fpm was
likely one of the last projects which still supported Ruby 1.8.

The idea of support Windows often focuses on the developer/product team, in
my experience. However, as fpm’s target audience includes
sysadmin/operators who often have no choice about their operating system
platform upgrade schedules, I have tended to support old rubies. There were
many years where Ruby 1.8 had been long abandoned by the Ruby development
team but was still the only Ruby version available on CentOS 6/7.

It’s mostly dependencies that abandon old rubies which force fpm to follow.
Usually the `backports` gem helps me defend against this pressure, but C
extensions like ffi aren’t covered by backports.

I still support 1.8, but it’s very hard to get some gem dependencies to run
on it.

jordansissel:
> This change breaks compatibility with arm64 based Macs

I'm still not sure how best to solve this. It feels like the only way to solve this is by having some separate installation steps based on the ruby version?

That is, revert this change and require `ffi` with no version constraint, and then have the documentation provide a table of safe ffi+ruby versions? This feels like a bad solution that'll also break things (requiring some amount of surprise work) for folks who have automation using fpm installation?

I wonder if the folks at ffi have any suggestions.

jordansissel:
I filed an issue upstream for advice since I'm not coming up with any solid ideas right now -- https://github.com/ffi/ffi/issues/906 -- perhaps the kind ffi folks might have ideas. I'm hoping they've had similar concerns in the past. Maybe we'll get lucky ;)

PR #1727, opened 2020-09-01
backports and pacman errors fixed

jjgalvez:
Please consider this pull request. It fixes the backports error warning, as well as changes the compression standard for pacman to zsdt which is the current standard. The build still fails two test. One has to do with copy_entry preserve flag not working in Linux (at least on Arch). and the other has to do with python #! which I didn't mess with. 
Note I'm not all that familiar with ruby code so please take my edits with grain of salt.

### DIFF
--- a/lib/fpm/package.rb
+++ b/lib/fpm/package.rb
@@ -4,7 +4,7 @@ require "pathname" # stdlib
 require "find"
 require "tmpdir" # stdlib
 require "ostruct"
-require "backports/2.0.0/stdlib/ostruct"
+require "backports/latest"
 require "socket" # stdlib, for Socket.gethostname
 require "shellwords" # stdlib, for Shellwords.escape
 require "erb" # stdlib, for template processing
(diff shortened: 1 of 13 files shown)


BartKoppers (changes requested):
I'm no pacman expert. However, using latest is - in general - considered bad practice for production, as not specific enough to maintain a coherent build. 
I propose that you substitute "latest" by a (semver) constant pointing at specific and tested gem module(s).

jordansissel:
@BartKoppers Thanks for reviewing! The "backports/latest" is, to me, roughly equivalent to the prior change "backports" and has to do with how the backports ruby library provides functionality to older rubies. Ideally, we replace all backports invocations with exactly what we need, instead of backports/latest, or review the codebase for any need for backports at all and remove where not needed.

This change looks good to me. I don't use Arch but do trust that this change ( default pacman compression ) is correct.

jordansissel:
Thank you for these fixes. :)

PR #1729, opened 2020-09-30
Use head/tail over sed for metadata fetching in sh.erb template

ehealy:
This was nice performance improvement @jlogsdon made some time ago to a private fork.


jordansissel:
Cool. Thanks for this patch :)

PR #1737, opened 2020-11-18
Add trusted-host option for pip

vv-p:
Sometimes private devpi registry doesn't have a valid ssl-certificate, so pip is unable to install something from it. This patch adds option "--trusted-host" which allows work with unsecure devpi registies.

### DIFF
--- a/lib/fpm/package/python.rb
+++ b/lib/fpm/package/python.rb
@@ -29,6 +29,9 @@ class FPM::Package::Python < FPM::Package
   option "--pypi", "PYPI_URL",
     "PyPi Server uri for retrieving packages.",
     :default => "https://pypi.python.org/simple"
+  option "--trusted-host", "PYPI_TRUSTED",
+    "Mark this host or host:port pair as trusted for pip",
+    :default => nil
   option "--package-prefix", "NAMEPREFIX",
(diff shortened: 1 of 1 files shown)


BartKoppers:
> Sometimes private devpi registry doesn't have a valid ssl-certificate, so pip is unable to install something from it. This patch adds option "--trusted-host" which allows work with unsecure devpi registies.

AFAIKS this parameter can hold (just) 1 host? 
Might be more logical to use/allow a list of hosts here. Something along  --pip-unsecure-hosts

jordansissel:
@BartKoppers Good insight. I think this code could be easily updated to allow multiple hosts without harming usability -- but I think we can wait until that is requested. Patch looks OK to me :)

lib/fpm/package/python.rb
```diff
@@ -136,7 +139,31 @@ def download_if_necessary(package, version=nil)
...
+        ]
+      end
+
+      setup_cmd += [
+        "--build",
+        target,
+        want_pkg,
+      ]
+      
+      safesystem(setup_cmd*" ")
```
jordansissel, line 166: safesystem() can take a list of arguments, so I'll make a small change after merging to use it that way.

jordansissel (approved):

PR #1738, opened 2020-11-24
virtualenv.rst: fix heading level

kenyon:
This makes the "Example uses" heading a subsection of the virtualenv section.


BartKoppers (approved):
Textual change only

jordansissel:
Thank you both for working on this :)

PR #1747, opened 2021-01-19
Call safesystem() with a list

jordansissel:
instead of a space-separated argument string

Tested with this command:

    bundle exec bin/fpm --debug -s python -t deb --python-pip =pip django

And verified that safesystem() to invoke pip is given a list.

Mentioned this here:
https://github.com/jordansissel/fpm/pull/1737#discussion_r560474705


jordansissel:
https://github.com/jordansissel/fpm/pull/1737#discussion_r560474705

Issue #1749, opened 2021-01-20
Regression in 1.12.0 when building deb package


rmanus:
Since the upgrade to 1.12.0, my deb package cannot install anymore with the following error:

```
/var/lib/dpkg/info/jagos.postinst: 71: /var/lib/dpkg/info/jagos.postinst: Syntax error: \"else\" unexpected
dpkg: error processing package jagos (--install):
installed jagos package post-installation script subprocess returned error exit status 2
```

The culprit code around line 71 is

```
systemctl --system daemon-reload >/dev/null || true
debsystemctl=$(command -v deb-systemd-invoke || echo systemctl)
if ! systemctl is-enabled setup-hugepages >/dev/null
then
else
    $debsystemctl restart setup-hugepages >/dev/null || true
fi
```

Notice that there is no code in 'if' section and this is causing the error.

Fyi, my package is built with 
```
--deb-systemd ./resources/build/jagos/setup-hugepages.service
```

dev-gto:
Hi!
I can also confirm this issue. It seems to be related to the lines beginning at

https://github.com/jordansissel/fpm/blob/d50f756d0e3f9e52bba18bee8e01c740982d18ad/templates/deb/postinst_upgrade.sh.erb#L23

I'm using fpm within Ubuntu 18.04

jordansissel:
I can probably provide a workaround in a patch, but I don't have the bandwidth to determine if it is the correct fix.

jordansissel:
#1752 is an attempt to resolve this

PR #1752, opened 2021-01-20
Ensure an if-then-else doesn't have an empty if block

jordansissel:
Issue #1749

This should help prevent errors like this during deb package installation:

```
/var/lib/dpkg/info/jagos.postinst: 71: /var/lib/dpkg/info/jagos.postinst: Syntax error: \"else\" unexpected
dpkg: error processing package jagos (--install):
installed jagos package post-installation script subprocess returned error exit status 2
```


adammohammed (approved):
Ran into this issue as well, new deb built with this fix has a valid postinst script.

statik (approved):
I can confirm that this fixes #1749, would it be possible to get a release with this included? I have tested on debian stretch.

jordansissel:
Thank you for testing! I've merged now and will include in the next release.

Issue #1759, opened 2021-01-26
deb package: still problem of file name with compression other that gzip


ppoilbarbe:
When creating a debian package with compression other than gzip, there is still a problem.
The file control.tar.xz is created, but the command "ar" does not find "control.tar.gz".

```
Creating {:path=>"/tmp/package-deb-build-a5ffe2e751bf520f38f15b2cf90d674d0666f0ac54e2ff27712592841f75/control.tar.xz", :from=>"/tmp/package-deb-build-a5ffe2e751bf520f38f15b2cf90d674d0666f0ac54e2ff27712592841f75/control", :level=>:info}
ar: control.tar.gz: No such file or directory {:level=>:info}
Process failed: ar failed (exit code 1). Full command was:["ar", "-qc", "/path/of/my/file.deb", "debian-binary", "control.tar.gz", "/tmp/package-deb-build-a5ffe2e751bf520f38f15b2cf90d674d0666f0ac54e2ff27712592841f75/data.tar.xz"] {:level=>:error}
```
I think that this is due to the hardcoded file name "control.tar.gz" in "lib/fpm/package/deb.rb" line 619:
```
    # pack up the .deb, which is just an 'ar' archive with 3 files
    # the 'debian-binary' file has to be first
    File.expand_path(output_path).tap do |output_path|
      ::Dir.chdir(build_path) do
        safesystem(*ar_cmd, output_path, "debian-binary", "control.tar.gz", datatar)
      end
    end
```
I am not sure of my ruby skills to make a path or pull request. I don't know where the correct file name is stored since there is many places where compression and file names are checked.

PR #1760, opened 2021-01-26
Replace hardcoded control.tar.gz by name depending on compression (#1…

ppoilbarbe:
A way to resolve the problem of control.tar.gz hardcoded and not depending on compression (while the correct file is created) see #1759

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -178,7 +178,7 @@ class FPM::Package::Deb < FPM::Package
   end
 
   option "--systemd", "FILEPATH", "Add FILEPATH as a systemd script",
-	:multivalued => true do |file|
+    :multivalued => true do |file|
     next File.expand_path(file)
   end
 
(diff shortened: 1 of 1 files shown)


ppoilbarbe:
Check failed with other error already in master branch concerning pacman.

jordansissel:
Confirmed that this looks like the correct approach. If I check an upstream Ubuntu/Debian package, it uses the same compression filename for both control and data:

```
% ar t /var/cache/apt/archives/containerd.io_1.4.6-1_amd64.deb
debian-binary
control.tar.xz
data.tar.xz
_gpgbuilder
```

jordansissel:
Testing manually:

```
% bundle exec bin/fpm -s empty -t deb --deb-compression xz -n example
Created package {:path=>"example_1.0_amd64.deb"}

% ar t example_1.0_amd64.deb
debian-binary
control.tar.xz
data.tar.xz

% sudo dpkg -i example_1.0_amd64.deb
Selecting previously unselected package example.
(Reading database ... 219975 files and directories currently installed.)
Preparing to unpack example_1.0_amd64.deb ...
Unpacking example (1.0) ...
Setting up example (1.0) ...
```

jordansissel:
Looks good. I'll write some tests to cover this after merging.

PR #1764, opened 2021-02-14
Fix broken postinst script

kimw:
`:` is required to make sure the final postinst-script is legal. Otherwise, it will be looks as below, sometimes,

```sh
if ! systemctl is-enabled <%= service %> >/dev/null 
then   # <--- broken here
else
    $debsystemctl restart <%= service %> >/dev/null || true
fi
```


Forage:
Duplicate PR of #1752

kimw:
Thanks for the report. PR closed.

kimw:
Any news about this issue, please move your step to #1752.

PR #1772, opened 2021-03-17
Use a newer version of ffi for Apple M1

al45tair:
Apple Silicon Macs need a newer version of ffi (I think 1.14 or so).

### DIFF
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -47,9 +47,15 @@ Gem::Specification.new do |spec|
   spec.add_dependency("childprocess", "< 1.0.0") # license: ???
 
   # For calling functions in dynamic libraries
-  # This is pinned because 1.13.0 requires Ruby 2.3 or later
-  # Ref: https://github.com/jordansissel/fpm/issues/1708
-  spec.add_dependency("ffi", "~> 1.12.0") # license: GPL3/LGPL3
+  if (Gem::Platform.local.os == "darwin" &&
+      Gem::Platform.local.cpu == "arm64")
+    # Apple Silicon Macs need ffi 1.14 at least (I tested 1.15)
+    spec.add_dependency("ffi", "~> 1.15")
+  else
+    # This is pinned because 1.13.0 requires Ruby 2.3 or later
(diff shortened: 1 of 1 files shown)


fpm.gemspec
```diff
@@ -47,9 +47,15 @@ Gem::Specification.new do |spec|
   spec.add_dependency("childprocess", "< 1.0.0") # license: ???
 
   # For calling functions in dynamic libraries
-  # This is pinned because 1.13.0 requires Ruby 2.3 or later
-  # Ref: https://github.com/jordansissel/fpm/issues/1708
-  spec.add_dependency("ffi", "~> 1.12.0") # license: GPL3/LGPL3
+  if (Gem::Platform.local.os == "darwin" &&
+      Gem::Platform.local.cpu == "arm64")
```
onnimonni, line 51: At least on my M1 machine `Gem::Platform.local.cpu` returns `universal` instead of `arm64` 🤷

jordansissel:
Unfortunately, `gem build` turns the `.gemspec` file into YAML which doesn't support executable conditions as your change proposes.

You can observe this by looking at the output, and here's what it looks like when I run it from master:

```
% gem build fpm.gemspec
% tar -xf fpm-*.gem -O metadata.gz | gzip -dc
...
dependencies:
- !ruby/object:Gem::Dependency
  name: ffi
  requirement: !ruby/object:Gem::Requirement
    requirements:
    - - "~>"
      - !ruby/object:Gem::Version
        version: 1.12.0
```

I'm still searching for a way to solve this ffi version problem.

jordansissel:
https://github.com/jordansissel/fpm/issues/1795 -- I'm to try removing `ffi` from fpm's dependency list.

jordansissel:
#1785 is resolved by #1786 which removes the `ffi` dependency entirely from ffi! Woo!

Thank you for this proposed change. I'm closing this now since I believe this issue to be resolved by #1786. 

One note, though: I don't have an M1 mac, so I don't have easy access to test that this issue is resolved. If you can, please test the next release and let me know :)

PR #1785, opened 2021-05-11
dir.rst: First example also requires trailing slash

zoeimogen:


jordansissel (approved):

jordansissel:
Thanks for the improvement! :)

Issue #1795, opened 2021-06-17
Discussion: fpm, ruby, ffi version conflicts


jordansissel:
The spread of ruby versions, platforms, and dependencies is making it harder and harder to provide a good user experience when installing and using fpm.

Problem situation:

In #1709, we pinned a dependency ffi to the 1.12.x branch to enable support for Ruby 2.2 and older. However, this causes problems for Apple M1 OSX users who need ffi 1.14 or newer. Further, some platforms distributing Ruby 3.0 will only make newer versions of ruby ffi available. If we allow any version _newer_ than 1.12, these older rubies will likely still fetch the _newer-and-incompatible_ ffi versions.

This is largely an impossible situation where a "simple" solution needlessly punishes some population of users. Do we make it harder for Mac users? or do we make it harder for CentOS 7 users? It's a tough situation just leaving me feeling sour.

This crossroads is giving me some pause. I'm not sure exactly what solution will come.

Focusing just on _ffi_, I wonder if we can remove `ffi` as a dependency?

If I look at the Gemfile.lock for the dependency tree, I see that fpm has dependencies on `ruby-xz` and `childprocess` both of which require ffi.

* lib/fpm/package/freebsd.rb does `require "xz"`
* lib/fpm/util.rb does `require "childprocess"`
* lib/fpm/util.rb uses FFI to access `mknod`

If we can replace these, fpm's need for `ffi` is eliminated.

jordansissel:
Related issues: #1772, #1649, #1709, #1783, #1770, #1742, #1713, #1708, #1638

jordansissel:
In hindsight, removing `ffi` seems critical now given the previous comment listing 9 issues caused by `ffi` incompatibilities with Ruby versions.

jordansissel:
I asked upstream for guidance, if any, and got a somewhat confusing response -- maybe I didn't provide enough context... https://github.com/ffi/ffi/issues/906

jordansissel:
Removing `xz` from the `freebsd.rb` could be done by using external program `xz`. We have a way to check for a program's presence using `program_exists?` and `program_in_path?`. The test suite uses this as well as the `pear` package support:

https://github.com/jordansissel/fpm/blob/72a925b048053f226eacd88b7ce2558e41b722db/lib/fpm/package/pear.rb#L39-L41

In particular for freebsd.rb, replacing this section with non-ffi code:

[...]

PR #1796, opened 2021-06-18
Work in progress: Remove FFI as a dependency

jordansissel:
In this PR, the following changes shall occur:

* [X] Replace ChildProcess library with Ruby standard `Process.spawn`
* [X] Replace FFI providing `mknod` syscall with Ruby standard methods, where possible.
* [x] Replace FreeBSD package support `xz` library with external `xz` program.

Acceptance criteria: fpm can be installed without any `ffi` gem dependency.

Expected outcomes, all of the following:
* Ruby 2.0-2.2 shall be supported again? (Needs more research)
* New Apple M1 Macs shall be able to use fpm successfully.
* Newer Arch systems which ship only Ruby FFI 1.15 shall be able to use fpm successfully

This PR aims to address issues listed in #1795

### DIFF
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -46,11 +46,6 @@ Gem::Specification.new do |spec|
   # Ref: https://github.com/jordansissel/fpm/issues/1592
   spec.add_dependency("childprocess", "< 1.0.0") # license: ???
 
-  # For calling functions in dynamic libraries
-  # This is pinned because 1.13.0 requires Ruby 2.3 or later
-  # Ref: https://github.com/jordansissel/fpm/issues/1708
-  spec.add_dependency("ffi", "~> 1.12.0") # license: GPL3/LGPL3
-
   spec.add_development_dependency("rake", "~> 10") # license: MIT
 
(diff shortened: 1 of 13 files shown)


jordansissel:
`rspec` is working as well as can be expected. Notable failures are things that have failed for a while, unrelated to this change (like cpan issues due to fpm test suite invoking cpan... wrongly?)

Test 1: FFI is not installed, confirmed below:
```
% grep -c ffi fpm.gemspec Gemfile Gemfile.lock
fpm.gemspec:0
Gemfile:0
Gemfile.lock:0
```

I spot-checked a few manual tests (`ffi -s gem -t rpm`, etc) as well as watching successful runs of the test suite. 👍👍

* I believe the `childprocess` replacement was successful.
* `mknod` is replaced by appropriate alternate implementations where possible. 
* However, character device (like /dev/tty) and block devices are no longer supported in copies. It is unclear what negative impact this will have.
* freebsd's package generation using ruby-xz is replaced with a `tar -Jc ...` equivalent that I hope is correct. There's no test suite for this package support, so I'll rely on users to report any issues.

PR #1797, opened 2021-06-24
add 'brew install rpm' as a prerequisite in the OSX/macOS section

ajoslin103:
add 'brew install rpm' as a prerequisite in the OSX/macOS section

### DIFF
--- a/docs/installing.rst
+++ b/docs/installing.rst
@@ -18,6 +18,8 @@ like OSX, come with Ruby already, but some do not. Depending on your operating s
 On OSX/macOS::
 
     brew install gnu-tar
+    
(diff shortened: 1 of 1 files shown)


jordansissel:
Nice! Thank you for improving fpm :)

ajoslin103:
I have been absolutely a fan of fpm since I found it back around 2015 ?

So long ago I can't remember !!

I've used it to make RH & Mac packages and tout it every time I'm asked (and a lot of times when I'm not!!)

Thank you for this program !!

jordansissel:
6 years! I’m glad fpm keeps bringing good results :)

<3

Issue #280, opened 2012-10-22
.deb control file broken by Provides param in FPM 0.4.20


adamgibbins:
0.4.20 generates a control file like so:

```
Package: rubygem-builder
Version: 3.1.4
License: MIT
Vendor: Jim Weirich
Architecture: all
Maintainer: <agibbins@ldn-dev-agibbins>
Installed-Size: 104
#Pre-Depends: 
Provides: rubygem-builder = 3.1.4
Section: Languages/Development/Ruby
Priority: extra
Homepage: http://onestepback.org
Description: Builder provides a number of builder objects that make creating structured data
 simple to do.  Currently the following builder objects are supported:
 .
 * XML Markup
 * XML Events
```

This throws an error such as "E: Problem parsing Provides line" when trying to use the package.

The package generated by 0.4.19 (without the version number in the provides line) works fine.  I believe you need to wrap most version numbers in brackets? e.g. "Provides: rubygem-builder (= 3.1.4)" although I haven't confirmed this.

PR #1803, opened 2021-07-06
Allow versions in `Provides` field for Debian packages

### LINKED ISSUE #1788: fpm --provides="foo (= 1.0)" doesn't preserve the specified version in the deb package
Example:
$ fpm -t deb -s empty -n foo_pkg -v 1.0rc --provides="foo (= 1.0)" --provides="bar" --deb-no-default-config-files
Created package {:path=>"foo-pkg_1.0rc_amd64.deb"}

$ dpkg --info foo-pkg_1.0rc_amd64.deb | grep Provides
 Provides: foo, bar

The expected output is:
Provides: foo (= 1.0), bar

jordansissel:
(more history/research for this PR available in the parent issue #1788)

---

This fixes #1788. This also reverts #280. For #280, at the time, this
change to ignore versions was correct. Two years after #280, Debian
began allowing `Provides` field to have versions.

This change also fixes bug in gem-to-deb conversion where previously an
incorrect Provides syntax would be generated (but thanks to #280,
removed), so this bug was only noticed after #280 was undone!

Computers are hard sometimes.

Added tests for gem-to-deb conversion specifically for the Provides
field.

Tested manually with Docker on Ubuntu 14.04 and 18.04 and results meet
expectations.

The history here is follows:
* In 2012, fpm was patched to remove version specifiers in Provides
  field because Debian didn't support it.
* In 2014, Debian dpkg[1] added support for versions in Provides field
* Somewhere between 2015-2018, Debian and Ubuntu included this new
  version of dpkg.
* Debian packaging policy docs (v4.4.0) was updated to allow versions
  in the Provides field.

Expected impacts:
* Older versions of dpkg/etc should _ignore_ the presence of a version
  specifier. Testing on Ubuntu 14.04 confirmed this.
* Newer versions of dpkg/etc should respect the presence of a version
  specifier. Testing on Ubuntu 18.04 confirmed this.

[1] https://launchpad.net/debian/+source/dpkg/1.17.11

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -687,6 +687,19 @@ class FPM::Package::Deb < FPM::Package
         File.unlink(changelog_path)
       end
     end
+
+    if origin == FPM::Package::Gem
+      # fpm's gem input will have provides as "rubygem-name = version"
+      # and we need to convert this to Debian-style "rubygem-name (= version)"
+      self.provides = self.provides.collect do |provides|
+        m = /^(#{attributes[:gem_package_name_prefix]})-([^\s]+)\s*=\s*(.*)$/.match(provides)
+        if m
+          "#{m[1]}-#{m[2]} (= #{m[3]})"
+        else
+          provides
+        end
+      end
+    end
   end # def converted_from
 
   def debianize_op(op)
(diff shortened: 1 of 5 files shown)


jordansissel:
Test suite looks good, and manual testing seems to work as expected.

Manual tests are described in #1788

PR #1806, opened 2021-07-27
Fix finding dependencies of old python modules

bugfood:
The fpm python helper relies on the `setup.py` file provided by the python module being built. Unfortunately, many python modules get their `setup()` command via:

```
from distutils.core import setup
```

`distutils` provides a `Command` class that does not support the `install_requires` attribute from which the helper extracts the dependencies. Thus, such modules will build, but the resulting packages will not specify any python requirements.

`distutils` is soon to be deprecated, and `setuptools` is an easy drop-in replacement.
https://www.python.org/dev/peps/pep-0632/

We cannot expect every module in the wild to make that change, though; for example, at the time of this writing, there are about 48,000 results on a github search:
https://github.com/search?q=from+distutils.core+import+setup+extension%3A.py+path%3A%2F+filename%3Asetup.py&type=Code&ref=advsearch&l=&l=
...and, from a rough check of the first few pages of results, it looks like about half of those rely entirely on `distutils`.

The `Command` class does not provide any way to get the original arguments of `setup()`; by the time the helper can run, unrecognized arguments have already been warned and dropped. The most simple way I could find to handle this is to create a wrapper to monkey-patch `distutils.core.setup` to be `setuptools.setup` instead.

Here is a simple example `setup.py`:
```
from distutils.core import setup

setup(name='example',
      version='0.0.1',
      packages=['example'],
      install_requires=['requests'])
```

Note the empty `dependencies` when calling it directly:
```
$ PYTHONPATH=/home/chickey/git/external/fpm/lib/fpm/package:. python3.9 setup.py --command-packages=pyfpm get_metadata --output test.json && cat test.json ; echo
/usr/lib/python3.9/distutils/dist.py:274: UserWarning: Unknown distribution option: 'install_requires'
  warnings.warn(msg)
running get_metadata
{
  "name": "example",
  "version": "0.0.1",
  "author": "UNKNOWN <UNKNOWN>",
  "description": "UNKNOWN",
  "license": "UNKNOWN",
  "url": "UNKNOWN",
  "architecture": "all",
  "dependencies": []
}
```

Note the populated dependencies when calling it via the wrapper:
```
$ PYTHONPATH=/home/chickey/git/external/fpm/lib/fpm/package:. python3.9 /home/chickey/git/external/fpm/lib/fpm/package/python_patch.py --command-packages=pyfpm get_metadata --output test.json && cat test.json ; echo
running get_metadata
{
  "name": "example",
  "version": "0.0.1",
  "author": "UNKNOWN <UNKNOWN>",
  "description": "UNKNOWN",
  "license": "UNKNOWN",
  "url": "UNKNOWN",
  "architecture": "all",
  "dependencies": [
    "requests"
  ]
}
```

As an addendum, use `setuptools` in the helper as well. Note that this on its own does not fix the problem described above, since the `Command` object attributes are set by `setup()` in `setup.py`. The aim for this is mostly to make fpm future-proof against an eventual lack of `distutils.core` availability.

I have tested this locally on python 2.6, 2.7, 3.5, and 3.9.


bugfood:
Updated PR contents to:
* fix use with modules which run `setup()` under a check for `__main__`
* fix use with modules which specify `distclass` as well

Updated PR description to fix a minor technical error.

bugfood:
Updated PR contents to fix inconsistency in package names vs. dependency names. This fixes a pre-existing bug (not a bug I introduced).

bugfood:
Fixed some typos in my commit message.

bugfood:
Updated PR contents to:
* compile `setup.py` contents before` exec()` in order to be able to have better line numbers
* set `__file__` for the `exec()` environment in order to fix use of `__file__` in setup.py

Sorry this keeps evolving; I find more edge cases as I build more packages.

bugfood:
Updated PR contents to:
* change default behavior to allow packages with underscores
* support name remapping

jordansissel:
@bugfood Can I test this? Or do you want me to wait on testing and reviewing? No pressure :)

bugfood:
Yes please. Sorry for the churn earlier. I was in the midst of a project and kept finding edge cases, and I'm done with that now. All the edge cases I encountered are now handled.

jordansissel:
No apology needed. :)

jordansissel:
Can you tell me about the new flag? I think I understand the scenario you want it to solve, and I'd like to go deeper to think about edge cases or other kinds of things -- https://github.com/jordansissel/fpm/pull/1806/commits/b14cd60752519cc102d23ab31003a6de07acc3d3

For example, remapping python-pyyaml to PyYAML makes sense if you want to depend directly on your distribution's version of the package. This one makes sense, though I can imagine some errors if your python module depends on a _different_ version of PyYAML than the distribution provides. Allowing users to make this choice and mapping seems like a good idea! I like it so far :)

PyYAML seems like a special oddity because even Debian calls it "python3-yaml" or "python-yaml" and not "python-pyyaml" 🤦, so fpm's guess of "python-pyyaml" is wrong in most cases. Tricky!

For the new flag, I like the idea itself -- allowing package name translation/mapping -- but I'm unsure about the flag name and the syntax. I don't think JSON is the right approach mainly due to the syntax complexity. Including JSON objects in shell requires some difficult quoting scenarios `--name-map '{ "foo": "bar" }'` where single quotes must be used to allow double-quotes. Thinking more, it seems like this name mapping is always going to be string-to-string, so we could use "foo=bar" to map dependency names. fpm already uses "foo=bar" for the `dir` package type's path mapping, so it wouldn't be too strange to have this same syntax used for package name mappings.

"--name-map" feels too short and not descriptive, though right now, I don't have any specific suggestions.

As for where the name mapping is applied, it feels like the right place to apply the name mapping is right after `output = input.convert(...)`: https://github.com/jordansissel/fpm/blob/b14cd60752519cc102d23ab31003a6de07acc3d3/lib/fpm/command.rb#L496

I'd like some test cases to cover this new functionality if only to help us make sure the behavior is what we expect. The current PR will do the name mapping before a bunch of default settings are added.

Could you move the `--name-map` feature to a separate PR where we can figure out how best to move forward? I think it's important enough to discuss in a new place where we wouldn't be mixing other discussion with the python metadata handling.

jordansissel:
For the python changes, I'd like to get some tests written (I can help). In order to do that, it might be useful to know about one or more python modules that don't work today with `fpm -s python`. Can you guide me on that? Are there public python modules that fail today but work with this PR? (I might have missed it in the PR description?)

Additionally, we could add another python test fixture to spec/fixtures/ which is known to fail _without_ this PR but succeeds with it. Thoughts?

bugfood:
Thanks for the feedback. I can work on addressing it in between some other work. If I don't address some parts for a while, I probably have not forgotten--I just haven't been able to allocate the time.

jordansissel:
No rush — there’s only so many hours and energy per day 👍🏻👍🏻

bugfood:
For an example module that doesn't have dependencies detected at all:
https://pypi.org/project/statelessd/

Prep work:
1. Start with the #1861 branch.
2.  Run:
```
make docker-release-everything
```

To build a package:
```
docker run -v "$(pwd):/src" --entrypoint /src/bin/fpm fpm --verbose -s python -t rpm -f --architecture noarch --rpm-auto-add-directories statelessd
```

Results on unpatched fpm:
```
$ rpm -qpR python-statelessd-0.0.6-1.noarch.rpm 
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
```

Results with my patch "fix detecting requirements of many older python modules":
```
$ rpm -qpR python-statelessd-0.0.6-1.noarch.rpm 
python-pika
python-tinman
rpmlib(CompressedFileNames) <= 3.0.4-1
rpmlib(PayloadFilesHavePrefix) <= 4.0-1
```

***

For an example module that has an issue with underscores, consider:
```
docker run -v "$(pwd):/src" --entrypoint /src/bin/fpm fpm --verbose -s python -t rpm -f --architecture noarch --rpm-auto-add-directories --python-bin python2 tornado
docker run -v "$(pwd):/src" --entrypoint /src/bin/fpm fpm --verbose -s python -t rpm -f --architecture noarch --rpm-auto-add-directories --python-bin python2 backports_abc
```

Results on unpatched fpm:
```
$ rpm -qpR python-tornado-5.1.1-1.noarch.rpm | grep abc
python-backports-abc >= 0.4
$ rpm -qpi python-backports_abc-0.5-1.noarch.rpm | grep Name
Name        : python-backports_abc
```
Note the mismatch between `backports-abc` and `backports_abc` ; unpatched fpm _cannot_ satisfy that dependency.

Results with my patch "fix inconsistent naming between packages and dependencies" (and its parents):
````
$ rpm -qpR python-tornado-5.1.1-1.noarch.rpm | grep abc
python-backports_abc >= 0.4
$ rpm -qpi python-backports_abc-0.5-1.noarch.rpm | grep Name
Name        : python-backports_abc
```

Thanks,
Corey

bugfood:
I made an independent PR for the name mapping here:
https://github.com/jordansissel/fpm/pull/1857
We can defer that for now.

I rebased this PR against current `master` and omitted the name mapping.

I think I have addressed all your questions for this PR. Let me know if I missed any or need to clarify.

Thanks,
Corey

bugfood:
@jordansissel can you please look at this old PR? It still applies cleanly and still seems to work for me. I had earlier moved the controversial name-mapping commit into a separate PR, to be considered later if we can.

Thanks,
Corey

PR #1808, opened 2021-07-27
fixes #1807; pip>20.3 removed --build flag

### LINKED ISSUE #1807: Python build fails when downloading with pip (pip>20.3)
Pip removed the --build option in pip 20.3, so Python auto-downloading builds fail when using new versions of pip. The suggested workaround is to use environment variables to set TMPDIR. Using TMPDIR does require modifying FPM's behavior slightly and will produce a warning, but this warning can safely be ignored per pypa/pip#9617. Using TMPDIR also does not appear to break versions of pip previous to when --build was removed. I'll attempt to submit a PR soon with the needed changes.

ipbeegle:
I'm too much of a FPM/Ruby noob to get all tests to work (seems to be something with my environment), but the changes do work for pip>20.3. If anyone is able to properly test these changes, I'd appreciate it. Submitting this as a draft until then.

### DIFF
--- a/lib/fpm/package/python.rb
+++ b/lib/fpm/package/python.rb
@@ -158,17 +158,19 @@ class FPM::Package::Python < FPM::Package
       end
 
       setup_cmd += [
-        "--build",
-        target,
         want_pkg,
       ]
+
+      # Use TMPDIR environment variable due to pip>20.3 removing the --build argument.
+      env = Hash["TMPDIR" => target]
+      setup_cmd.unshift(env)
       
       safesystem(*setup_cmd)
     end
 
-    # easy_install will put stuff in @tmpdir/packagename/, so find that:
-    #  @tmpdir/somepackage/setup.py
-    dirs = ::Dir.glob(File.join(target, "*"))
+    # pip will download stuff to @tmpdir/pip-download-*/, so find that:
+    #  @tmpdir/pip-download-*/somepackage-*/setup.py
+    dirs = ::Dir.glob(File.join(target, "pip-download-*/*"))
     if dirs.length != 1
       raise "Unexpected directory layout after easy_install. Maybe file a bug? The directory is #{build_path}"
     end


frebib:
I tested this on one package and it seems to do the trick 👍🏻

frebib:
We discovered that this breaks on Debian Stretch with this older version of setuptools:
```
python3-setuptools:
  Installed: 33.1.1-1
  Candidate: 33.1.1-1
  Version table:
 *** 33.1.1-1 500
        500 https://deb.debian.org/debian stretch/main amd64 Packages
        100 /var/lib/dpkg/status
        ```

**Comment by frebib** [comment]
With older versions of pip/setuptools (on Debian Stretch) we needed to look in `pip-build-*` instead. This patch on top seems to fix it:
```diff
 lib/fpm/package/python.rb | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/lib/fpm/package/python.rb b/lib/fpm/package/python.rb
index 89c27c6..7f45329 100644
--- a/lib/fpm/package/python.rb
+++ b/lib/fpm/package/python.rb
@@ -168,9 +168,9 @@ class FPM::Package::Python < FPM::Package
       safesystem(*setup_cmd)
     end
 
-    # pip will download stuff to @tmpdir/pip-download-*/, or @tmpdir/pip-build-*/ so find that:
+    # pip will download stuff to @tmpdir/pip-download-*/, so find that:
     #  @tmpdir/pip-download-*/somepackage-*/setup.py
-    dirs = ::Dir.glob(File.join(target, "pip-{build,download}-*/*"))
+    dirs = ::Dir.glob(File.join(target, "pip-download-*/*"))
     if dirs.length != 1
       raise "Unexpected directory layout after easy_install. Maybe file a bug? The directory is #{build_path}"
     end
```

rooterkyberian:
seems important, especially due to #1820  , any plan on continuation of this work?

jordansissel:
@rooterkyberian I haven't tried changing how fpm invokes pip. I'll take a look to see if we can avoid the `--build` flag as proposed by this PR in my work on #1820.

jordansissel:
I can confirm that the propsed patch works. I need to do more research to test newer vs older versions of pip to see how well this works for other versions. I tested on pip 20.0.2 (which is older than the `--build` flag removal, nice!)

PR #1812, opened 2021-08-03
FreeBSD package: Fix missing leading `/` in archive

### LINKED ISSUE #1811: FreeBSD package not installable
The change introduced by commit ab4eb18b5fe41045a6956313010375e8a74e6547 (which targeted v1.13.0) changed the non-metadata files inside the FreeBSD package archive so they don't have leading `/` anymore, as noted in the change as well:
https://github.com/jordansissel/fpm/blob/55f9613c0da3e95a0e93ff5c7efd4dad5eb48e48/lib/fpm/package/freebsd.rb#L89-L91

Unfortunately, such FreeBSD package can't be installed. It needs a leading `/` in non-metadata file names in archive:

```
$ find /tmp/src
/tmp/src
/tmp/src/usr
/tmp/src/usr/bin
/tmp/src/usr/bin/script
/tmp/src/etc
/tmp/src/etc/config

$ fpm --force --name 'test-package' --version 0.0.1 --output-type freebsd --input-type dir --chdir /tmp/src
Force flag given. Overwriting package at test-package-0.0.1.txz {:level=>:warn}
Created package {:path=>"test-package-0.0.1.txz"}

$ tar --list -f test-package-0.0.1.txz
+COMPACT_MANIFEST
+MANIFEST
etc/config
usr/bin/script

# pkg install -y test-package-0.0.1.txz || echo FAIL
Updating FreeBSD repository catalogue...
FreeBSD repository is up to date.
All repositories are up to date.
Checking integrity... done (0 conflicting)
The following 1 package(s) will be affected (of 0 checked):

New packages to be INSTALLED:
	test-package: 0.0.1

Number of packages to be installed: 1
[1/1] Installing test-package-0.0.1...
Extracting test-package-0.0.1:   0%
pkg: File //etc/config not specified in the manifest         # <-------------
Extracting test-package-0.0.1: 100%
FAIL
```

vholer:
This is a simple workaround, which transforms the filenames inside the
tar archive so that they start with `/`. This happens only in case
the filename doesn't begin with `+`, which is expected (or at least is
very likely) to be a metadata file.

```
$ tar --list -f test-package-0.0.1.txz
+COMPACT_MANIFEST
+MANIFEST
/etc/config
/usr/bin/script

$ pkg install -y test-package-0.0.1.txz && echo OK
Updating FreeBSD repository catalogue...
FreeBSD repository is up to date.
All repositories are up to date.
Checking integrity... done (0 conflicting)
The following 1 package(s) will be affected (of 0 checked):

New packages to be INSTALLED:
	test-package: 0.0.1

Number of packages to be installed: 1
[1/1] Installing test-package-0.0.1...
Extracting test-package-0.0.1: 100%
OK
```

Fixes #1811

Signed-off-by: Vlastimil Holer <vholer@opennebula.io>

### DIFF
--- a/lib/fpm/package/freebsd.rb
+++ b/lib/fpm/package/freebsd.rb
@@ -86,9 +86,7 @@ class FPM::Package::FreeBSD < FPM::Package
     # We use --files-from here to keep the tar entries from having `./` as the prefix.
     # This is done as a best effor to mimic what FreeBSD packages do, having everything at the top-level as
     # file names, like "+MANIFEST" instead of "./+MANIFEST"
-    # Note: This will include top-level files like "/usr/bin/foo" listed in the tar as "usr/bin/fo" without
-    # a leading slash. I don't know if this has any negative impact on freebsd packages.
-    safesystem("tar", "-Jcf", output_path, "-C", staging_path, "--files-from", build_path("file_list"))
+    safesystem("tar", "-Jcf", output_path, "-C", staging_path, "--files-from", build_path("file_list"), "--transform", 's|^\([^+]\)|/\1|')
   end # def output
 
   # Handle architecture naming conversion:


jordansissel:
Nice! Thank you for finding this solution.

jordansissel:
Testing manually:

```
% bundle exec bin/fpm -s dir -t freebsd -n example /usr/bin/sh
Created package {:path=>"example-1.0.txz"}

% tar -tf example-1.0.txz
+COMPACT_MANIFEST
+MANIFEST
tar: Removing leading `/' from member names
/usr/bin/sh
```

Looks good!

davidnewhall:
This is a brilliant solution and I probably wouldn't have found that tar option myself.

It requires GNU tar and it makes it difficult to build packages for FreeBSD on FreeBSD.

PR #1816, opened 2021-08-13
fixes #1307 - Also use setuptools' extras_require which contains complex dependencies

### LINKED ISSUE #1307: Python packages: FPM ignores dependencies markers (PEP0508)

jvdmr:
ex. 'charset_normalizer~=2.0.0; python_version >= "3"' is now included if the marker matches the build environment.


jordansissel:
I don't have the python experience to evaluate exactly what this fixes. However, the code looks OK, so I'm good with merging it.

PR #1819, opened 2021-08-19
Parse ~= operator according to PEP440 rules (compatible versions allowed)

jvdmr:
Currently python dependency versions using `~=` are converted to `=`
This changes that to a version range using `>=` and `<` according to the rules specified in PEP440 (https://www.python.org/dev/peps/pep-0440/#compatible-release)

Note that PEP440 actually specifies using `>= <major>.<minor>` and `== <major>.*`.
However at least .deb packages do not allow wildcards, so ie. instead of `== 2.*` I opted for `< 3.0`, which together with the `>= 2.2` would have the same effect and which would work as expected in any package type.

### DIFF
--- a/lib/fpm/package/python.rb
+++ b/lib/fpm/package/python.rb
@@ -162,7 +162,7 @@ class FPM::Package::Python < FPM::Package
         target,
         want_pkg,
       ]
-      
+
       safesystem(*setup_cmd)
     end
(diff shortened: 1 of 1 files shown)


jvdmr:
Hold on this one, I seem to have run into a bug in this code...

PR #1821, opened 2021-08-21
By default, use pip for fetching and installing python packages.

### LINKED ISSUE #1820: Use pip by default, if possible.
fpm has supported `pip` for a while as an optional alternative to `easy_install`. I'm not sure if easy_install is even available anymore on newer Python installations, and pip seems to be the recommended/default way to install python packages.

Objective: fpm should now try to use `pip` if it is available and fall back to `easy_install` if needed.

I don't know what negative impact this might have.

easy_install has been [deprecated since 2019 ](https://setuptools.readthedocs.io/en/latest/history.html#v42-0-0), per the python setuptools release notes:

> Breaking Changes
> ... Mark the easy_install script and setuptools command as deprecated

On my workstation, a recent-ish install of Ubuntu 18.04, easy_install isn't installed for me when installing Python. I don't even see it available in `apt search easy_install`. Pip, however, is available as an Ubuntu `python-pip` package.

Supporting Links:
* https://packaging.python.org/discussions/pip-vs-easy-install/
* https://setuptools.readthedocs.io/en/latest/history.html#v42-0-0

Related: https://github.com/jordansissel/fpm/issues/1780

jordansissel:
This adds a new flag, --python-internal-pip, which is enabled by default.

"internal pip" means using 'python -m pip' to invoke pip. Ideally this will make fpm more correctly use pip.

Tested on python 2.7.17 and 3.6.9 on Ubuntu 18.04

All python tests passing 👍👍

Fixes #1820

### DIFF
--- a/lib/fpm/package/python.rb
+++ b/lib/fpm/package/python.rb
@@ -79,7 +79,11 @@ class FPM::Package::Python < FPM::Package
   option "--setup-py-arguments", "setup_py_argument",
     "Arbitrary argument(s) to be passed to setup.py",
     :multivalued => true, :attribute_name => :python_setup_py_arguments,
-    :default => []
+    :default => [] 
+  option "--internal-pip", :flag,
+    "Use the pip module within python to install modules",
+    :attribute_name => :python_internal_pip,
+    :default => true
 
   private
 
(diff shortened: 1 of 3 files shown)


jordansissel:
A cool thing is that some python tests were failing before and this helped fix things. One of the tests was using python3 and specifying "pip" which, on my workstation, was python2's pip. This made for amusing clashes as pip downloaded an ancient version of Django and python3 failed trying to install it.

jordansissel:
pip now complains about being invoked.

> WARNING: pip is being invoked by an old script wrapper. This will fail in a future version of pip. 
> Please see https://github.com/pypa/pip/issues/5599 for advice on fixing the underlying issue. 
> To avoid this problem you can invoke Python with '-m pip' instead of running pip directly.

jordansissel:
```
% bundle exec rspec spec/fpm/package/python_spec.rb
.............

Finished in 12.2 seconds (files took 0.22577 seconds to load)
13 examples, 0 failures
```

With the last commit, the test suite now passes. 👍

jordansissel:
```
% bundle exec bin/fpm -s python -t deb --python-bin python3 django
Created package {:path=>"python-django_3.2.9_all.deb"}
```

Django seems to be packaging ok.

jordansissel:
I'm nervous about merging this, but I gain confidence with passing rspec tests as well as manual tests.

Two manual tests:

Running using the new default, internal pip aka "python -m pip":

```
% bundle exec bin/fpm -s python -t deb --python-bin python3 requests
Created package {:path=>"python-requests_2.26.0_all.deb"}
```

Running with internal pip disabled. This should fall back to easy_install by default, which isn't present on my system:

```
% bundle exec bin/fpm -s python -t deb --python-bin python3 --no-python-internal-pip requests
Need executable 'easy_install' to convert python to deb {:level=>:error}
```

Issue #1833, opened 2021-09-22
rpm: An iteration with a dash causes rpmbuild to fail.


jordansissel:
Related issue: https://github.com/electron-userland/electron-builder/issues/5976

At present, passing `--iteration 1-2_3` to fpm when building an rpm will cause rpmbuild to fail:

```
% fpm -s empty -t rpm -n example --iteration 1-2_3 --verbose
...
error: line 41: Illegal char '-' (0x2d) in: Release: 1-2_3 {:level=>:info}
Process failed: rpmbuild failed (exit code 1). ...
```

fpm has other scenarios where it will automatically "correct" any package attributes that are not valid in the target package format:
* `--name Example` will be downcased to `example` if the target package is `deb`. 
* `--version` with dashes will convert to underscores if the target is `deb`

As suggested by @JonathanLydall in [the related electron-builder issue](https://github.com/electron-userland/electron-builder/issues/5976#issuecomment-880417173), we can automatically correct for this by changing the dash to underscore.

jordansissel:
Creeping scope a bit, I wondered (also mentioned in the related issue) what other characters are not valid.

@, #, and & are rejected. % is accepted (when used as %%), parentheses `()` are not valid either. I'm not thinking of any particular easy replacements for these invalid characters.

I think fpm could detect them, know they are invalid, and provide an actionable error message. However, for this issue, I'll stay focused on replacing `-` with `_`

PR #1834, opened 2021-09-22
rpm: Replace dash with underscore.

jordansissel:
(note: fpm calls 'iteration' what rpm calls 'release')

rpmbuild will reject the `Release` tag containing a dash with the
following error (via fpm --verbose):

```
error: line 41: Illegal char '-' (0x2d) in: Release: 1-1 {:level=>:info}
Process failed: rpmbuild failed (exit code 1).
```

This patch copies the dash-to-underscore operation that is already
applied to the version field.

Adds tests for both iteration and version field.

Fixes: #1833

### DIFF
--- a/lib/fpm/package/rpm.rb
+++ b/lib/fpm/package/rpm.rb
@@ -254,6 +254,12 @@ class FPM::Package::RPM < FPM::Package
 
   # This method ensures a default value for iteration if none is provided.
   def iteration
+    if @iteration.kind_of?(String) and @iteration.include?("-")
+      logger.warn("Package iteration '#{@iteration}' includes dashes, converting" \
+                   " to underscores. rpmbuild does not allow the dashes in the package iteration (called 'Release' in rpm)")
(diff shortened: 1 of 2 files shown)


jordansissel:
Running manually, fpm will now log a warning about the text replacement that it is doing (as it does for other minor corrections)

```
% fpm -s empty -t rpm -n example --iteration '1-1'          
Package iteration '1-1' includes dashes, converting to underscores. rpmbuild does not allow the dashes in the package iteration (called 'Release' in rpm) {:level=>:warn}
...
Created package {:path=>"example-1.0-1_1.x86_64.rpm"}
```

jordansissel:
Test suite says 🆗 

```
% bundle exec rspec spec/fpm/package/rpm_spec.rb  ./spec/fpm/package/rpm_spec.rb
....................................................................................................................................

Finished in 5.81 seconds (files took 0.18619 seconds to load)
132 examples, 0 failures
```

Issue #1844, opened 2021-10-29
freeBSD broke between v1.11.0 and v1.13.1


claytonjwong:
Hi, did this code change make it into the `v1.13.1` release? 
 If so, then it breaks freeBSD packaging: https://github.com/jordansissel/fpm/commit/63fdb9434704563dfbe20769e22531be58fa0c18

---

**Test Performed:** I performed the following 2 tests back to back within freeBSD 12 on a vagrant instance on my localhost mac:

1. ❌ Use fpm `v1.13.1` to show incorrect package manifest, ie. each file is missing a `/` prefix 😵
```
[vagrant@freebsd12 ~/package]$ tar tf speedtest-1.2.3.txz
+COMPACT_MANIFEST
+MANIFEST
speedtest
speedtest.5
speedtest.md
```

2. ✅ Use fpm `v1.11.0` on the same exact files to show correct package manifest, ie. each file has a `/` prefix 😊

```
+COMPACT_MANIFEST
+MANIFEST
/speedtest
/speedtest.5
/speedtest.md
```

---

**Full Test Input/Output:**

```
[vagrant@freebsd12 ~/package]$ fpm --verbose -s dir -t freebsd -n speedtest -v 1.2.3 -a amd64 -m "support@ookla" --package . .=/
Setting workdir {:workdir=>"/tmp", :level=>:info}
Setting from flags: architecture=amd64 {:level=>:info}
Setting from flags: epoch= {:level=>:info}
Setting from flags: iteration= {:level=>:info}
Setting from flags: maintainer=support@ookla {:level=>:info}
Setting from flags: name=speedtest {:level=>:info}
Setting from flags: url=http://example.com/no-uri-given {:level=>:info}
Setting from flags: version=1.2.3 {:level=>:info}
Converting dir to freebsd {:level=>:info}
Created package {:path=>"./speedtest-1.2.3.txz"}
[vagrant@freebsd12 ~/package]$ tar tf speedtest-1.2.3.txz
+COMPACT_MANIFEST
+MANIFEST
speedtest
speedtest.5
speedtest.md
[vagrant@freebsd12 ~/package]$ fpm --version
1.13.1
[vagrant@freebsd12 ~/package]$ sudo gem uninstall fpm

Select gem to uninstall:
 1. fpm-1.11.0
 2. fpm-1.13.1
 3. All versions
> 2
Successfully uninstalled fpm-1.13.1
[vagrant@freebsd12 ~/package]$ fpm --version
[...]

PR #1845, opened 2021-10-30
Add tests for FreeBSD packages.

jordansissel:
- Files in the tarball should begin with / (#1811, #1844)
- Assert certain top-level manifest fields
- Assert manifest files are present

Idea from #1844


jordansissel:
```
% bundle exec rspec spec/fpm/package/freebsd_spec.rb
................

Finished in 0.71171 seconds (files took 0.14319 seconds to load)
16 examples, 0 failures
```

PR #1852, opened 2021-11-10
Only show /etc "config files" warning when files are present in /etc.

### LINKED ISSUE #1851: Only warn about Debian's /etc config-files thing when there are files in the package's /etc
Today, any debian package, even empty ones, will show this warning:

```
% bundle exec bin/fpm -s empty -t deb -n foo
Created package {:path=>"foo_1.0_all.deb"}
```

It would be better if this warning was only given if there were files in the `/etc` path of the package.

jordansissel:
This should silence a warning in cases when packages do not contain any
/etc files. Prior to this commit, most debian package creation would
have this warning emitted:

```
```

Fixes #1851

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -1007,7 +1007,7 @@ class FPM::Package::Deb < FPM::Package
     etcfiles = []
     # Add everything in /etc
     begin
-      if !attributes[:deb_no_default_config_files?]
+      if !attributes[:deb_no_default_config_files?] && File.exists?(staging_path("/etc"))
         logger.warn("Debian packaging tools generally labels all files in /etc as config files, " \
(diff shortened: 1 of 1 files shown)


jordansissel:
Testing:

```
# No files in /etc, should have no warning ✅
% bundle exec bin/fpm -s empty -t deb -n foo
Created package {:path=>"foo_1.0_amd64.deb"}
```

```
# Files in /etc, should have warning by default. ✅
% bundle exec bin/fpm -s dir -t deb -n foo /etc/bash.bashrc
Created package {:path=>"foo_1.0_amd64.deb"}
```

```
# Files in /etc, override flag given. No warning ✅
% bundle exec bin/fpm -s dir -t deb --deb-no-default-config-files -n foo  /etc/bash.bashrc
Created package {:path=>"foo_1.0_amd64.deb"}
```

Issue #1854, opened 2021-11-10
packaging deb using 1.14.0 leads to a NameError


kares:
namely `NameError: uninitialized constant FPM::Package::CPAN` (compared to version 1.13.x)

due the use of the [`FPM::Package::CPAN`](https://github.com/jordansissel/fpm/blob/v1.14.0/lib/fpm/package/deb.rb#L680) from https://github.com/jordansissel/fpm/commit/13621c97229ee66b17e99ec0ecb765e1f0511fc7 (in the `FPM::Package::Deb` class)

trying to use `require "fpm/package/deb"` one needs to these extra "seemingly unrelated" requires:
```ruby
require "fpm/package/gem"
require "fpm/package/cpan"

require "fpm/package/deb"
...
dir.convert(FPM::Package::Deb)
```
(likely a similar limitation exists for `FPM::Package::Rpm` -> `require "fpm/package/gem"`)

jordansissel:
Thanks for trying the new version!

That’s odd.. is this for Logstash (which I think invokes fpm using the Ruby
api)? I don’t get this error using the cli.

Should be an easy fix and I’ll get this fixed shortly.

jordansissel:
```
% bundle exec ruby -r./lib/fpm/package/deb.rb -e 'FPM::Package::Deb.new.convert(FPM::Package::Deb)'
/home/jls/projects/fpm/lib/fpm/package/deb.rb:680:in `converted_from': uninitialized constant FPM::Package::CPAN (NameError)
```

Reproduced via api

jordansissel:
I'm on the fence about the "best" solution here. Added `require` is simple, but maybe checking the type as a string instead of a const...

jordansissel:
fpm 1.14.1 pushed which has (hopefully) fixed this. Can you try?

jsvd:
Confirmed it works now, thanks for the super fast fix <3

PR #1855, opened 2021-11-10
Add missing requires for package type constants

jordansissel:
In #1854, it is reported that packaging fails. I think this happens because that specific use case invokes fpm via api and not via cli. When the cli is run, it loads all package type files. However, with the api, you `require()` your own files. 

I think files should require things they need, so this PR adds require() calls for places where  one package refers to another (like when converting things to deb). 

I used `grep FPM::Package:: lib/fpm/package/*.rb` to find cases where requires might be needed.

Test cases that failed without this PR:

```
% bundle exec ruby -r./lib/fpm/package/rpm.rb -e 'FPM::Package::RPM.new.tap { |x| x.name = "fancy" }.convert(FPM::Package::RPM)' 

% bundle exec ruby -r./lib/fpm/package/deb.rb -e 'FPM::Package::Deb.new.tap { |x| x.name = "fancy" }.convert(FPM::Package::Deb)' 
```


jordansissel:
I don't know why github things there's 10 commits, there's only two:

```
% git log --oneline master..issue/1854
6357ba3 (HEAD -> issue/1854, origin/issue/1854) Require fpm/package/dir to make FPM::Package::Dir available (for #1854)
0e8dc2e Add requires so that other package types are available
```

jordansissel:
I'm closing this because github still thinks there's 10 commits to merge. I'll try a new PR.

PR #1857, opened 2021-11-11
support a user-specified name mapping

bugfood:
This PR is split off from the original:
https://github.com/jordansissel/fpm/pull/1806
...per request from @jordansissel .

The original PR is a soft dependency of this PR: in order to remap package dependencies, the dependencies need to be generated properly in the first place.

I have not tested this PR in isolation. Let's work out the remainder of the original PR first.

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -5,6 +5,7 @@ require "fpm/util"
 require "clamp"
 require "ostruct"
 require "fpm"
+require "json"
 require "tmpdir" # for Dir.tmpdir
 
 if $DEBUG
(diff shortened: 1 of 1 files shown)


bugfood:
To address:
https://github.com/jordansissel/fpm/pull/1806#issuecomment-964778985

As to why I used JSON: it's been a while since I wrote that patch, but I think my thought process was basically:

1. Something like `--map src1=dst1 --map src2=dst2` would make sense.
2. Can I guarantee that `fpm` does not and will not ever support a packaging format that allows `=` in package names?
3. I'll have to provide some escaping mechanism just in case.
4. I don't want to write my own serialization format.
5. Everybody knows JSON (well, ok, not everybody, but the target user base ought to know it).
6. Typing the JSON boilerplate is a bit annoying, so just use one big argument instead of many little ones, to minimize the overhead.

I do concede that this isn't the only way to do it, but it seemed like the most correct approach at the time. If you prefer a different way, I can change it.

bugfood:
One other way that JSON works well is when needing to support a mapping that is not static across different OS/python version combinations. Once this gets complex enough, then using a scripting language like perl/python/ruby instead of bash is pretty much a necessity, in order to incrementally build up the mapping as a hash/dictionary data structure. At that point, generating the JSON from the hash/dictionary is trivial.

PR #1865, opened 2021-12-03
Fix regression in Provides parser for Debian packages

Lekensteyn:
Allow periods and plus in package name to match the Debian docs:

    "Package names (both source and binary, see Package) must consist
    only of lower case letters (a-z), digits (0-9), plus (+) and minus
    (-) signs, and periods (.). They must be at least two characters
    long and must start with an alphanumeric character."

Also note that the previous filter `A-z` would accidentally match
characters such as `^` by accident. Underscores are also invalid
(presumably to avoid breaking parsers of things like
`name_version_arch.deb`. Both issues have been fixed.

Fixes #1829


Lekensteyn:
Odd, `make test` fails. Investigating:
```
  1) FPM::Package::Deb#output with Provides values should reject 'NotValid'
     Failure/Error: insist { original.output(target) }.raises FPM::InvalidPackageConfiguration
     Insist::Failure:
       Expected exception 'FPM::InvalidPackageConfiguration' but none was raised
     # ./spec/fpm/package/deb_spec.rb:442:in `block (4 levels) in <top (required)>'

  2) FPM::Package::Deb#output with Provides values should reject 'not_valid'
     Failure/Error: insist { original.output(target) }.raises FPM::InvalidPackageConfiguration
     Insist::Failure:
       Expected exception 'FPM::InvalidPackageConfiguration' but none was raised
     # ./spec/fpm/package/deb_spec.rb:442:in `block (4 levels) in <top (required)>'
```

Lekensteyn:
Updated the patch to account for conflicts due to commit 88c1e6ab0abf634993b8fbfe6202d42e04010ca7 (which fixed the period issue and already adds a link to the docs), and commit f898ef82326055afee53c2bb7baf33efbf16dfa1 (which split the regex onto multiple lines).

The original bug (disallowing `+` and accepting things such as `^` due to `A-z`) will be fixed by this PR.

PR #1878, opened 2022-02-24
Restore development_dependency on `rake`

### LINKED ISSUE #1877: Removing `rake` as a development_dependency causes the test suite to fail.
Commit db9db670c39a84dccc92123eed4413bf09d89bc5 removed `rake` as a development dependency. I was pretty sure rake wasn't used in `fpm`'s development environment, so this seemed safe.

However, `rspec` (the test suite) now fails:

```
% bundle exec rspec
bundler: failed to load command: rspec (/home/jls/.rbenv/versions/3.0.2/bin/rspec)
/home/jls/projects/fpm/lib/fpm/rake_task.rb:3:in `require': cannot load such file -- rake (LoadError)
	from /home/jls/projects/fpm/lib/fpm/rake_task.rb:3:in `<top (required)>'
	from /home/jls/projects/fpm/spec/fpm/rake_task_spec.rb:3:in `require'
...
```

`lib/fpm/rake_task.rb` was added in 2014 (#756). 

Thoughts:
- I don't have any reason to _remove_ this functionality.
- adding `rake` as a regular dependency might be ok
- but [new versions of Rake require Ruby 2.2 or newer](https://github.com/ruby/rake/blob/5c60da8644a9e4f655e819252e3b6ca77f42b7af/rake.gemspec#L98).
- older versions (12.3.3 and below) of rake cause Dependabot and other automated bots to harass me as a maintainer.
- I don't know if anyone still uses this fpm code (FPM::RakeTask)

-----

In order to make sure this code continues to work, I think that reverting the db9db670c39a84dccc92123eed4413bf09d89bc5 might be the correct choice -- with one caveat, that there is no specific Rake version required, just any should be fine, but I'll let the test suite confirm that.

jordansissel:
A prior commit removed `rake` as a development dependency on the
assumption that nothing in fpm's development process actually required
the use of `rake`.

However, I had forgotten about #756 (year 2014) which adds FPM::RakeTask
and some test coverage. FPM::RakeTask was added to allow folks to more
easily invoke FPM from within a Rake task.

At this time, I see no reason to remove FPM::RakeTask. Further, because
`rake` is typically (in my experience) used only in development
environments. Therefore, I believe the right solution is to restore
`rake` as a development_dependency in order to allow the test suite to
pass. For users of FPM::RakeTask, I would assume (hopefully correctly!)
that they already have `rake` installed as a dependency in their own
project, so the `fpm` gem has no need to specify `rake` as a general-use
dependency.

When studying newer versions of Rake, I found:
* Rake v13 requires Ruby 2.2 or newer
* Rake v12.xx and older are flagged as having security vulnerabilities.

In order to minimize chaos, this commit adds an unversioned dependency
on rake. This is to help fpm service more versions of Ruby and resist
efforts by any dependency to dictate which version of Ruby is used.

This reverts db9db670c39a84dccc92123eed4413bf09d89bc5.

Fixes #1877

### DIFF
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -57,6 +57,8 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency("insist", "~> 1.0.0") # license: Apache 2
   spec.add_development_dependency("pry")
 
+  spec.add_development_dependency("rake") # For FPM::RakeTask, #1877, #756
+
   spec.files = files
   spec.require_paths << "lib"
   spec.bindir = "bin"


jordansissel:
```
% bundle exec rspec spec/fpm/rake_task_spec.rb
.....

Finished in 0.00419 seconds (files took 0.17951 seconds to load)
5 examples, 0 failures

```

PR #1879, opened 2022-03-08
Fix `--deb-compression none`

howardjohn:
Previously, we would pass the literal `""` as an argument to `tar`.
`tar` would interpret this as a file name, which does not exist, and
fail.

This fixes the command to just pass no compression flag at all to tar
when `--deb-compression none` is set.

I did not add tests since I couldn't figure out how to execute them -
this is my first time working in ruby.

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -613,25 +613,24 @@ class FPM::Package::Deb < FPM::Package
       when "gz", nil
         datatar = build_path("data.tar.gz")
         controltar = build_path("control.tar.gz")
-        compression = "-z"
+        compression_flags = ["-z"]
       when "bzip2"
         datatar = build_path("data.tar.bz2")
         controltar = build_path("control.tar.gz")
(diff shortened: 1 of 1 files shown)


howardjohn:
Hi @jordansissel is there anything we can do to get this merged? This cuts our CI time dramatically but we would prefer not to use a fork

jordansissel:
> is there anything we can do to get this merged

I sat down to test this PR, and before even `rspec` finishes, my kids call for me in the other room.

I think we're all powerless in this regard, though I appreciate the question! The pandemic completely rearranged where my time and energy is spent, and it shows no real signs of improving in the next 6 months.

The deb spec suite fails in both master and this pr, so I don't think it introduces any problems.

I'll merge it now, but I don't know when I'll be able to do a new release (this process takes about 5-15 minutes, but I don't know when I"ll have that kind of time)

howardjohn:
Totally understand - thanks for taking the time. We really appreciate `fpm`

PR #1888, opened 2022-03-31
Release prep for v1.14.2

jordansissel:


jordansissel:
Test skips: OSX and virtualenv were skipped because I don't' have a mac laptop and I also don't have virtualenv installed at this time.

Test failures:

```
  1) FPM::Package::Deb compression when --deb-compression is bzip2 should not cause lintian to crash
     Failure/Error: insist { system("lintian", "-C", "symlinks", target) } == true
     Insist::Failure:
       Expected true, but got false
     # ./spec/fpm/package/deb_spec.rb:546:in `block (5 levels) in <top (required)>'

  2) FPM::Package::Deb compression when --deb-compression is xz should not cause lintian to crash
     Failure/Error: insist { system("lintian", "-C", "symlinks", target) } == true
     Insist::Failure:
       Expected true, but got false
     # ./spec/fpm/package/deb_spec.rb:546:in `block (5 levels) in <top (required)>'

  3) FPM::Package::Deb compression when --deb-compression is gz should not cause lintian to crash
     Failure/Error: insist { system("lintian", "-C", "symlinks", target) } == true
     Insist::Failure:
       Expected true, but got false
     # ./spec/fpm/package/deb_spec.rb:546:in `block (5 levels) in <top (required)>'

  4) FPM::Package::Python python_scripts_executable is set should have scripts with a custom hashbang line
     Failure/Error: fd = File.new(path, "r")
     Errno::ENOENT:
       No such file or directory @ rb_sysopen - /tmp/package-python-staging-02ea77df67d2a0afc15182fdb679e626601407e8564b7c423df1509066d4/usr/local/bin/django-admin.py
     # ./spec/fpm/package/python_spec.rb:210:in `initialize'
     # ./spec/fpm/package/python_spec.rb:210:in `new'
     # ./spec/fpm/package/python_spec.rb:210:in `block (3 levels) in <top (required)>'
```

jordansissel:
The "lintian -C symlinks ..." tests (3) are failing for some reason. This test used to pass, but I wonder if Debian/lintian removed the "symlink" check? 

The goal of this check was to detect cases where fpm would generate a package file that would cause lintian to crash.

```
% lintian -C symlinks example_1.0_amd64.deb
Unrecognized check script (via -C): symlinks
```

🤷

The 4th failure, python_scripts_executable, is likely due to the Django project changing things and possibly not providing any `/usr/local/bin/django-admin.py` file? 

We can investigate later. These failures seem like a problem with the test itself and not with fpm.

PR #1897, opened 2022-05-02
Call ERB.new correctly depending on the RUBY_VERSION

### LINKED ISSUE #1894: Ruby 3.1.0 and above give deprecation warning about ERB.new() usage
Running `fpm` on Ruby 3.1.0 or newer can result in a warning that is caused by Ruby 3.1.0 deprecating a specific way that `ERB.new` is invoked by fpm.

## Background

For context, ERB is a templating library that comes with Ruby. Fpm uses ERB for templating rpm spec files, package scripts like `--after-install`, etc. 

At this time, fpm creates a new ERB instance with `ERB.new(the_template_code, nil, "-")` where the 2nd and 3rd arguments are describing ERB's "safe_level" and "trim_mode".

It appears as of Ruby 3.1.0, specifying `safe_mode` is not a feature anymore and will be removed. Further, specifying `trim_mode` now should be done as a keyword argument, such as `ERB.new(the_template_code, trim_mode: "-")`

# Research

## When was this introduced?

Here's some testing I did to find out when this warning was introduced:

```
% for i in 2.6.0 3.0.0 3.1.0; do podman run -it docker.io/ruby:$i ruby -rerb -e 'puts RUBY_VERSION; ERB.new("<%= 123 %>",
 nil, "-")'; done
2.6.0
3.0.0
3.1.0
-e:1: warning: Passing safe_level with the 2nd argument of ERB.new is deprecated. Do not use it, and specify other arguments as keyword arguments.
-e:1: warning: Passing trim_mode with the 3rd argument of ERB.new is deprecated. Use keyword argument like ERB.new(str, trim_mode: ...) instead.
```

Shown above, Ruby 3.1.0 reports these warnings while older versions do not.

Idea: Depending on the version of Ruby being used, invoke `ERB.new` in whatever way that works _and_ doesn't print these warnings.

## When was the new method syntax allowed?

Testing Ruby versions shown below, we can see that Ruby 2.3.0 and 2.4.0 do _not_ seem to support the keyword argument for `trim_mode` and throw an exception when given a keyword argument for `trim_mode`
```
% for i in 2.3.0 2.4.0 2.6.0 3.0.0 3.1.0; do podman run -it docker.io/ruby:$i ruby -rerb -e 'puts RUBY_VERSION; erb = ERB.new("<%= 123 %>", trim_mode: "-"); p erb.result(binding)'; done
2.3.0
/usr/local/lib/ruby/2.3.0/erb.rb:860:in `block in result': no implicit conversion of Hash into Integer (TypeError)
        from /usr/local/lib/ruby/2.3.0/erb.rb:862:in `result'
        from -e:1:in `<main>'
2.4.0
/usr/local/lib/ruby/2.4.0/erb.rb:892:in `block in result': no implicit conversion of Hash into Integer (TypeError)
        from /usr/local/lib/ruby/2.4.0/erb.rb:894:in `result'
        from -e:1:in `<main>'
2.6.0
"123"
3.0.0
"123"
3.1.0
"123"
```

# Conclusion

* Ruby versions 2.4.0 and older do not support the newer `ERB.new(template, trim_mode: "-")` argument
* Ruby versions 3.1.0 and newer print a warning when using the older syntax, `ERB.new(template, safe_level, trim_mode)`

In order to support older versions of ruby, I think `ERB.new` calls in FPM should be the old syntax for ruby versions 3.0.x and older, and use the new syntax for ruby versions 3.1.x and newer.

The goal is to reduce useless(*) warnings printed to fpm users. By "useless" I mean that there is no obvious action a user can take to resolve these warnings in fpm because the problem is a change in Ruby 3.1.0 and how fpm invokes ERB.

This can be done by checking RUBY_VERSION before calling ERB.new.

jordansissel:
* On Ruby 3.0.x and older, call ERB.new(template_code, nil, "-")
* On Ruby 3.1.0 and newer, call ERB.new(template_code, trim_mode: "-")

Fixes #1894

### DIFF
--- a/lib/fpm/package.rb
+++ b/lib/fpm/package.rb
@@ -331,10 +331,11 @@ class FPM::Package
     template_path = File.join(template_dir, path)
     template_code = File.read(template_path)
     logger.info("Reading template", :path => template_path)
-    erb = ERB.new(template_code, nil, "-")
+    erb = erbnew(template_code)
     erb.filename = template_path
     return erb
   end # def template
+
 
   #######################################
   # The following methods are provided to
@@ -518,7 +519,7 @@ class FPM::Package
   # flag), then apply it as an ERB template.
   def script(script_name)
     if attributes[:template_scripts?]
-      erb = ERB.new(scripts[script_name], nil, "-")
+      erb = erbnew(scripts[script_name])
       # TODO(sissel): find the original file name for the file.
       erb.filename = "script(#{script_name})"
       return erb.result(binding)
--- a/lib/fpm/util.rb
+++ b/lib/fpm/util.rb
@@ -406,6 +406,29 @@ module FPM::Util
   def logger
     @logger ||= Cabin::Channel.get
   end # def logger
+
+  def erbnew(template_code)
+    # In Ruby 2.6(?), Ruby changed how ERB::new is invoked.
+    # First, it added keyword args like `ERB.new(..., trim_mode: "-")`
+    # Later, it deprecated then removed the safe_level feature.
+    # As of Ruby 3.1, warnings are printed at runtime when ERB.new is called with the old syntax.
+    # Ruby 2.5 and older does not support the ERB.new keyword args.
+    #
+    # My tests showed:
+    # * Ruby 2.3.0 through 3.0 work correctly with the old syntax.
+    # * Ruby 3.1.0 and newer (at time of writing, Ruby 3.2) require the new syntax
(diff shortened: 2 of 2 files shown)


jordansissel:
Test results:

```
% bundle exec rspec spec/fpm/package/rpm_spec.rb spec/fpm/package_spec.rb        
...................................................................................................................................

Finished in 3.64 seconds (files took 0.1879 seconds to load)
131 examples, 0 failures
```

jordansissel:

PR #1898, opened 2022-05-02
Use YAML.unsafe_load on Ruby 3.1.0 and newer

### LINKED ISSUE #1895: psych/class_loader.rb:99:in `find': Tried to load unspecified class: Gem::Specification (Psych::DisallowedClass)
* When doing `fpm -s gem -t deb rails`
* On some versions of ruby

fpm tries to load a downloaded gem's metadata, it can fail with an error 

```
psych/class_loader.rb:99:in `find': Tried to load unspecified class: Gem::Specification (Psych::DisallowedClass)
```

This seems to be a new failure caused by an upstream ruby or psych change. This ruby psych issue seems related: https://github.com/ruby/psych/pull/487

The above ruby psych change mentions that `safe_load` is now the default (`YAML.load` is "safe_load") and the previous `load` method has been renamed `unsafe_load`.

Need to dig a bit more to figure out the best solution that does safe loading, but for now, calling `unsafe_load` when it's available is probably a short fix.

jordansissel:
On older rubies, YAML.load _is_ the unsafe load method. At some point,
Ruby 3.1.0 / Psych 4(?) made two renames:

* YAML.load -> YAML.unsafe_load
* YAML.safe_load -> YAML.load

A quick test is to try converting a gem. This would fail if `YAML.load`
was the "safe" method because it would fail with this message:

    Tried to load unspecified class: Gem::Specification (Psych::DisallowedClass

`fpm -s gem -t empty rails` will crash on Ruby 3.1.0 prior to this
commit.

Fixes #1895

### DIFF
--- a/lib/fpm/package/gem.rb
+++ b/lib/fpm/package/gem.rb
@@ -135,9 +135,19 @@ class FPM::Package::Gem < FPM::Package
     return gem_files.first
   end # def download
 
+  GEMSPEC_YAML_CLASSES = [ ::Gem::Specification, ::Gem::Version, Time, ::Gem::Dependency, ::Gem::Requirement, Symbol ]
   def load_package_info(gem_path)
-
-    spec = YAML.load(%x{#{attributes[:gem_gem]} specification #{gem_path} --yaml})
+    # TODO(sissel): Maybe we should check if `safe_load` method exists instead of this version check?
+    if ::Gem::Version.new(RUBY_VERSION) >= ::Gem::Version.new("3.1.0")
+      # Ruby 3.1.0 switched to a Psych/YAML version that defaults to "safe" loading
+      # and unfortunately `gem specification --yaml` emits YAML that requires
+      # class loaders to process correctly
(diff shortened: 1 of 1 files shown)


jordansissel:
I haven't tested this one yet. I tried tonight but my spare install of ruby 3.0.2 was not built correctly and now it is sleep time.

jordansissel:
Testing both ruby 3.1.0 and 3.0.2, tests pass!

Prior to this PR, Ruby 3.0.2 would pass tests, but Ruby 3.1.0 would not due to the upstream Psych change.

```
% ruby --version
ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux-gnu]

% bundle exec rspec spec/fpm/package/gem_spec.rb
........

Finished in 15.26 seconds (files took 0.21311 seconds to load)
8 examples, 0 failures
```

```
% ruby --version
ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [x86_64-linux]
% bundle exec rspec spec/fpm/package/gem_spec.rb
........

Finished in 7.55 seconds (files took 0.18037 seconds to load)
8 examples, 0 failures
```

PR #1904, opened 2022-05-18
Use bash in deb templates

mkenigs:
Currently, fpm has a bug where if I use #!bash for an --after-install
script that uses bash specific features, it works, but if I add an
--after-upgrade script, the previously working --after-install script
now fails, because the script is now run with sh.

If this change is too opinionated, maybe an alternative fix would be to respect the shebang in the `--after-install` script?

### DIFF
--- a/templates/deb/ldconfig.sh.erb
+++ b/templates/deb/ldconfig.sh.erb
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env bash
 # This script is automatically added by fpm when you specify
 # conditions that usually require running ldconfig upon
 # package installation and removal.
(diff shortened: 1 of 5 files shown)


jordansissel:
> maybe an alternative fix would be to respect the shebang in the --after-install script?

I like this alternate proposal, though I understand the implementation might be trickier.

automatically choosing the right shebang line seems like a good approach. Default to /bin/sh unless a script uses something else. We might need to detect conflicting shebang lines, like if after upgrade and after install have different shebang lines?

mkenigs:
Yeah that'd probably be better! Don't think I have time to do it that way right now because I have a workaround for what I'm using it for

jordansissel:
> Don't think I have time to do it that way right now 

No worries! We can make forward progress as time permits for anyone available to do the work (including myself). For now, I'm glad we could bounce ideas and come up with a solution that should work well, while you at least have a decent work around for yourself. <3

PR #1909, opened 2022-06-04
Use pkgbuild's --install-location flag when fpm's --prefix is given.

### LINKED ISSUE #1908: osxpkg: Unable to use generated package in newer versions of MacOS
Since fpm is using pkgbuild under the hood, could we add a parameter that passes the value to the '--install-location' parameter of pkgbuild? 

Without the ability to specify that parameter, the installation location defaults to the root. Which in Mac OS Catalina and above is defaulted to read-only, so the package created ends up unusable.

jordansissel:
A user reported that on macOS Catalina, the default pkgbuild install
location may result in a .pkg file which cannot be installed.

This change makes fpm's `--prefix` option pass through to `pkgbuild`'s
`--install-location` flag

> With macOS Catalina, you can no longer store files or data in the read-only system volume, nor can you write to the "root" directory ( / ) from the command line, such as with Terminal.

References:
* macOS Catalina's new read-only root filesystem: https://support.apple.com/en-us/HT210650

Fixes #1908

### DIFF
--- a/lib/fpm/package/osxpkg.rb
+++ b/lib/fpm/package/osxpkg.rb
@@ -148,6 +148,11 @@ class FPM::Package::OSXpkg < FPM::Package
       write_scripts
       args += ["--scripts", scripts_path]
     end
+
+    if attributes[:prefix]
+      args += ["--install-location", attributes[:prefix]]
+    end
+
     args << output_path
 
     safesystem("pkgbuild", *args)


jordansissel:
I don't have a macOS Catalina machine to test with, so I'll have to test and hope I'm doing it right from Linux.

On my linux machine, I faked it by making a `pkgbuild` executable that just prints the arguments:

```
#!/bin/sh

echo ">>" "$0" "$@"
```

Testing fpm defaults (no given `--prefix` flag), and trying to print just the command-line arguments. In this example, no `--install-location` flag should be present because fpm's `--prefix` flag is not set:

```
% bundle exec bin/fpm --debug -s empty -t osxpkg -n example | grep pkgbuild | fex '0,/:message/"2'
Running command
>> /home/jls/bin/pkgbuild --identifier example --info /tmp/fpm-PackageInfo20220603-19372-k5evbs --version 1.0 --ownership recommended --root /tmp/package-osxpkg-staging-d37efd5a5a3e5e9ee4ad96c580fea275f8e865fa250570073465f86771d0 example-1.0.pkg
```

Now, if I specify `--prefix /Applications/example`, fpm should include `--install-location` when invoking pkgbuild:

```
% bundle exec bin/fpm --debug -s empty -t osxpkg -n example --prefix /Applications/example | grep pkgbuild | fex '0,/:message/"2'
Running command
>> /home/jls/bin/pkgbuild --identifier example --info /tmp/fpm-PackageInfo20220603-19385-ahzws5 --version 1.0 --ownership recommended --root /tmp/package-osxpkg-staging-004778ce9aea5b2f40a4e5b12d08db76656aeb5b6c71da49ae1d714a5f1e --install-location /Applications/example example-1.0.pkg
```

(Github makes pre-formatted things scrollable horizontally, so you might have to scroll to the right on the above examples.

Based on this testing, I think this change *should* work, roughly summarizing:

* Without `--prefix`: `pkgbuild ... example-1.0.pkg` (no use of `--install-location`)
* With `--prefix /Applications/example` given:  `pkgbuild ... --install-location /Applications/example example-1.0.pkg`

PR #1913, opened 2022-06-24
Change default Debian package priority to optional

### LINKED ISSUE #1398: The debian package priority extra is deprecated

chrisnovakovic:
Since Debian Policy version 4.0.1, the "extra" priority has been deprecated and replaced with "optional"; this triggers Lintian warnings for Debian packages built by FPM. Make this the new default for Debian package outputs when no value of `--deb-priority` is given.

Fixes #1398.

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -93,7 +93,7 @@ class FPM::Package::Deb < FPM::Package
   end
 
   option "--priority", "PRIORITY",
-    "The debian package 'priority' value.", :default => "extra"
+    "The debian package 'priority' value.", :default => "optional"
 
   option "--use-file-permissions", :flag,
     "Use existing file permissions when defining ownership and modes"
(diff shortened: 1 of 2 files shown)


jordansissel:
Nice! Thank you for improving fpm :)

jordansissel:

PR #1921, opened 2022-08-09
Allow aborted upgrades to fail gracefully

matthauck:
Without this, `abort-upgrade` (due to a non-zero exit code from a before-upgrade script)
also fails the postrm script, which leaves the debian package in an error state, rather than
simply aborting the upgrade.

### DIFF
--- a/templates/deb/postrm_upgrade.sh.erb
+++ b/templates/deb/postrm_upgrade.sh.erb
@@ -42,7 +42,7 @@ then
     # 'apt-get purge' runs 'on remove' section, then this section.
     # There is no equivalent in RPM or ARCH.
     after_purge
-elif [ "${1}" = "upgrade" ]
+elif [ "${1}" = "upgrade" -o "${1}" = "abort-upgrade" ]
 then
     # This represents the case where the old package's postrm is called after
     # the 'preinst' script is called.


jordansissel:
Can you tell me a bit more about this? I'm having trouble finding a failing test case that this PR fixes.

I created two empty scripts for both `--after-upgrade` and `--after-remove` and I don't get any errors. I think I'm not understanding the issue so I'm at a loss for how to test it. Can you help me understand more?

Here's what I thought would trigger the bug, but it seems to work fine:
```
% cat /tmp/after-upgrade
#!/bin/sh

echo "after-upgrade: $*"

% cat /tmp/after-remove
#!/bin/sh

echo "after-remove $*"

% for i in 1.0 1.1; do bundle exec bin/fpm -f -s empty -t deb -v $i --after-upgrade /tmp/after-upgrade --after-remove /tmp/after-remove -n example; done
Adding action files
Created package {:path=>"example_1.0_all.deb"}
Adding action files
Created package {:path=>"example_1.1_all.deb"}

## Now testing the packages in a container
% podman run -it -v $PWD/example_1.0_all.deb:/example_1.0_all.deb  -v $PWD/example_1.1_all.deb:/example_1.1_all.deb ubuntu sh -c 'dpkg -i example_1.0_all.deb; dpkg -i example_1.1_all.deb; dpkg -r example'
Selecting previously unselected package example.
(Reading database ... 4406 files and directories currently installed.)
Preparing to unpack example_1.0_all.deb ...
Unpacking example (1.0) ...
Setting up example (1.0) ...
(Reading database ... 4408 files and directories currently installed.)
Preparing to unpack example_1.1_all.deb ...
Unpacking example (1.1) over (1.0) ...
Setting up example (1.1) ...
after-upgrade: 1.0
(Reading database ... 4408 files and directories currently installed.)
Removing example (1.1) ...
after-remove
```

matthauck:
Sure, I had an installer that had a `before-upgrade` script that had an `exit 1` in it to abort the upgrade, and it got stuck in a "very bad inconsistent state".

on trying to `dpkg -r`, it would say:
```
dpkg: error processing package foo (--remove):
 package is in a very bad inconsistent state; you should
 reinstall it before attempting a removal
Errors were encountered while processing:
 foo
```

Issue #1929, opened 2022-08-27
Distribution should be set to `--deb-dist` instead of 'whatever'


WilliamDEdwards:
In [`deb.changes`](https://github.com/jordansissel/fpm/blob/121989e0a00cd8cf7997e49abcbf0ef2041bf47c/templates/deb/deb.changes.erb#L17) and [`changelog.erb`](https://github.com/jordansissel/fpm/blob/121989e0a00cd8cf7997e49abcbf0ef2041bf47c/templates/deb/changelog.erb#L1), the distribution is hardcoded to `whatever`. They should instead be set to the value of `--deb-dist`. What's more, the default `--deb-dist` value seems to be `unstable`, which causes the following Lintian warning by default:

    changes: distribution-and-changes-mismatch unstable whatever

jordansissel:
Can you tell me about the negative impact of this issue? Knowing more about it will help in writing test cases for it.

WilliamDEdwards:
> Can you tell me about the negative impact of this issue?

When I, as a user, am able to set the distribution (using `--deb-dist`), I expect it to be set everywhere. Instead, it's only set in one (control) out of three (control, changelog, changes) places. This could cause confusion.

lchabert:
And without a perfect alignment with deb-dist, some tools refuses to upload the package into debian repository. (for instance, dput-ng)

jordansissel:
fpm 1.15.0 is released and, I believe, has resolve this issue. Please let me know if you find otherwise in a new issue (if a new problem) or this issue (if same problem)

PR #1934, opened 2022-09-29
Replace whatever in changelog.erb and deb.changes.erb

lchabert:
In order to support packaging upload, the distribution must be set/aligned in changelog and changes files.
Required for example inside dput-ng repo upload app.

### DIFF
--- a/templates/deb/deb.changes.erb
+++ b/templates/deb/deb.changes.erb
@@ -14,7 +14,7 @@ Description: <%= firstline %>
 <%= remainder.collect { |l| l =~ /^ *$/ ? " ." : " #{l}" }.join("\n") %>
 <% end -%>
 Changes:
-<%= name %> (<%= "#{epoch}:" if epoch %><%= version %><%= "-" + iteration.to_s if iteration %>) <%= distribution %>; urgency=medium
+ <%= name %> (<%= "#{epoch}:" if epoch %><%= version %><%= "-" + iteration.to_s if iteration %>) <%= distribution %>; urgency=medium
   * Package created with FPM.
 Checksums-Sha1:
 <% changes_files.each do |file| -%>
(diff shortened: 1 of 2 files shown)


lchabert:
It will help to closes #1929

templates/deb/deb.changes.erb
```diff
@@ -14,7 +14,7 @@ Description: <%= firstline %>
 <%= remainder.collect { |l| l =~ /^ *$/ ? " ." : " #{l}" }.join("\n") %>
 <% end -%>
 Changes:
- <%= name %> (<%= "#{epoch}:" if epoch %><%= version %><%= "-" + iteration.to_s if iteration %>) whatever; urgency=medium
+<%= name %> (<%= "#{epoch}:" if epoch %><%= version %><%= "-" + iteration.to_s if iteration %>) <%= distribution %>; urgency=medium
```
jordansissel, line 17: I think the leading whitespace removal will cause a breakage. Debian's changes file docs say this, from https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-changes - (Bold emphasis added by me)

> The first line of the field value (the part on the same line as Changes:) is always empty. The content of the field is expressed as **continuation lines, with each line indented by at least one space**. Blank lines must be represented by a line consisting only of a space and a full stop (.).

jordansissel (changes requested):
Thank you for working on this!
I ran the test suite on this PR and found no new failures, nice!

That said, there's no test coverage for the `.changes` file nor for the changelog file either. I added one inline comment which I believe to cause a failure due, but I haven't tested this with any external Debian tools to confirm.

templates/deb/deb.changes.erb
lchabert: Sorry, i didnt see the missing space. It sould be ok now.

lchabert:
As i wrote on you code comment, it's should be ok with the space now.

jordansissel:
Thanks for the quick response! Change looks good :)

jordansissel:

PR #1935, opened 2022-10-05
chore: refactor Dockerfile

gnought:
- change base image to ubuntu:20.04
- support ruby 2.7
- docker lint


jordansissel:
At present I don''t have a machine with Docker, so I'm not able to test.

Looking at the diff, the changes look pretty good, though.

Issue #1517, opened 2018-07-04
Perl CPAN Support, Broken tar --strip-components Command


wbraswell:
Another bug, this time found while packaging the Alien::astyle software from CPAN.

```
$ fpm --no-cpan-test --cpan-verbose --verbose --debug-workspace --workdir /home/wbraswell/rperl_packager_tmp/ -t rpm -s cpan Alien::astyle
...
Setting workdir {:workdir=>"/home/wbraswell/rperl_packager_tmp/", :level=>:info}
Asking metacpan about a module {:module=>"Alien::astyle", :level=>:info}
Downloading perl module {:distribution=>"Alien-astyle", :version=>nil, :level=>:info}
Setting package name from 'distribution' {:distribution=>"Alien-astyle", :level=>:info}
Installing any build or configure dependencies {:level=>:info}
cpanm (App::cpanminus) 1.7044 on perl 5.016003 built for x86_64-linux-thread-multi {:level=>:info}
Work directory is /home/wbraswell/.cpanm/work/1530682819.10424 {:level=>:info}
You have make /usr/bin/make {:level=>:info}
You have LWP 6.05 {:level=>:info}
You have /usr/bin/tar: tar (GNU tar) 1.26 {:level=>:info}
Copyright (C) 2011 Free Software Foundation, Inc. {:level=>:info}
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>. {:level=>:info}
This is free software: you are free to change and redistribute it. {:level=>:info}
There is NO WARRANTY, to the extent permitted by law. {:level=>:info}
 {:level=>:info}
Written by John Gilmore and Jay Fenlason. {:level=>:info}
You have /usr/bin/unzip {:level=>:info}
Entering /home/wbraswell/rperl_packager_tmp/package-cpan-build-8a40f27acfd35733cd6c9f21ebcd00a98f5d8e5b87eaf730b3551d185a59/module {:level=>:info}
--> Working on /home/wbraswell/rperl_packager_tmp/package-cpan-build-8a40f27acfd35733cd6c9f21ebcd00a98f5d8e5b87eaf730b3551d185a59/module {:level=>:info}
Configuring /home/wbraswell/rperl_packager_tmp/package-cpan-build-8a40f27acfd35733cd6c9f21ebcd00a98f5d8e5b87eaf730b3551d185a59/module ... N/A {:level=>:info}
[...]

PR #1940, opened 2022-10-17
CPAN support - Fix broken tar command

NicholasBHubbard:
This PR fixes issue #1517 using the fix provided by #1518.

For handling CPAN tarballs with leading "./Foo/" directory structures, we must employ a tar `--transform` regex instead of a tar `--strip-components` command.

This bug can be reproduced by running the following command on the main branch:
```
fpm --no-cpan-test --cpan-verbose --verbose --debug-workspace --workdir $HOME/tmp -t rpm -s cpan -v 0.010000 Alien::astyle
```
I also added a test to `cpan_spec.rb` that ensures Alien::astyle version 0.010000 can be built, but it should be noted that (on my machine) this tests causes the test suite to take about 4 minutes to run versus only about 30 seconds without this test.

### DIFF
--- a/lib/fpm/package/cpan.rb
+++ b/lib/fpm/package/cpan.rb
@@ -305,7 +305,7 @@ class FPM::Package::CPAN < FPM::Package
     directory = build_path("module")
     ::Dir.mkdir(directory)
     args = [ "-C", directory, "-zxf", tarball,
-      "--strip-components", "1" ]
+             %q{--transform=s,[./]*[^/]*/,,} ]
     safesystem("tar", *args)
     return directory
   end
--- a/spec/fpm/package/cpan_spec.rb
+++ b/spec/fpm/package/cpan_spec.rb
@@ -1,4 +1,5 @@
 require "spec_setup"
+require "tmpdir" # for Dir.mktmpdir
 require "fpm" # local
 require "fpm/package/cpan" # local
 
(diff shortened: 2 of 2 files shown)


wbraswell:
@jordansissel 
As with the other Perl-related code, I have personally reviewed this pull request and I approve it.  Please approve and merge, thanks!

wbraswell:
Ping @jordansissel

jordansissel:
I haven’t been at a desk much this week, so I haven’t been able to review this PR.

jordansissel (changes requested):
Thanks for working on this!

This new test takes a pretty long time on my workstation. I tried timing it, but my attempt errored after 4 minutes because my test container was missing `g++` -- it surely would have taken more time if I had that dependency.

Are there any other ways we can try test this that might be faster? It also adds additional external dependencies (g++, etc) which adds additional maintenance burdens to the test suite.

One thought is that we could test the `unpack` method directly. This method has `private` access, but we can still invoke it in Ruby using `instance_eval`, like this:

```
directory = subject.instance_eval { unpack("my.tar.gz") }

# The 'directory' variable is a string with the path to where the module was unpacked.
# Now we can list things in that directory and verify the paths are as expected.
```

If we include a .tar.gz in this test as a fixture with a known directory structure, specifically the relative `./foo/bar` style paths, it could be tested without reaching out to the internet, without adding a dependency, and much faster.

Here's an example I came up with to test -- 

```
# Create a tarball with a file './foo/bar'
$ mkdir -p /tmp/z/foo
$ touch $_/bar
$ tar -C /tmp/z -cvzf hello.tar.gz .
./
./foo/
./foo/bar

$ bundle exec irb
...
irb(main):003:0> directory = subject.instance_eval { unpack("hello.tar.gz") }
=> "/tmp/package-cpan-build-c133f629e51c353b598c8924fe88767b6407896f2fa875e41b3eb2a58a7c/module"
```

The above shows how we can invoke `unpack` directly, and we can use other tools like ruby's `Find.find` method or similar to make assertions about the structure and contents of the unpacked tarball.

What do you think?

NicholasBHubbard:
Thanks for the idea @jordansissel! I used the example code you gave me to create a test that shows that cpan.rb's unpack method can handle tarballs containing `./` leading paths. Let me know what you think.

jordansissel:
Nice! Much faster.

```
% bundle exec rspec spec/fpm/package/cpan_spec.rb
....

Finished in 11.32 seconds (files took 0.20792 seconds to load)
4 examples, 0 failures
```

wbraswell:
Thanks @jordansissel !  :-)

PR #1944, opened 2022-10-25
Fix frozen cpanm caused by interactive Makefile.PL

NicholasBHubbard:
This PR fixes issue #1519 and #1522. I also added a test to `cpan_spec.rb` that makes sure Inline::CPP version 0.80 can be built.

I based my fix off [this commit](https://github.com/jordansissel/fpm/pull/1610/commits/ae9594eb098f74c55166729d317d86af83e3b3f9) from #1610 that was never merged.

Because Inline::CPP version 0.80 has an interactive Makefile.PL, the bug can be reproduced by running the following command on the main branch (this command will hang):

```
fpm --no-cpan-test --cpan-verbose --verbose --debug-workspace --workdir $HOME/tmp -t rpm -s cpan -v 0.80 Inline::CPP
```

### DIFF
--- a/lib/fpm/util.rb
+++ b/lib/fpm/util.rb
@@ -158,6 +158,9 @@ module FPM::Util
       stdout_r.close                unless stdout_r.closed?
       stderr_r.close                unless stderr_r.closed?
     else
+      # If no block given (not interactive) we should close stdin_w because we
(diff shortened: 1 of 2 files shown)


NicholasBHubbard:
Deprecated for #1955

PR #1946, opened 2022-10-26
Remove `git` gem dependency.

### LINKED ISSUE #1923: Dependency Error public_suffix requires Ruby version >= 2.6 and breaks installing on Debian 10 / Buster
```
Error installing fpm:
	The last version of public_suffix (< 6.0, >= 2.0.2) to support your Ruby & RubyGems was 4.0.7. Try installing it with `gem install public_suffix -v 4.0.7` and then running the current command again
	public_suffix requires Ruby version >= 2.6. The current ruby version is 2.5.0.
```
Maybe a similar cause like this one: https://github.com/jordansissel/fpm/issues/1918

I assume this will happen more and more as direct and transitive ruby dependencies will require higher ruby versions that are only available in current Linux distributions. All the LTS distributions like Debian LTS, Ubuntu LTS, RedHat, etc. are "sensitive" to this problem. Debian 11 is not affected and i assume the latest Ubuntu and Fedora versions are not affected too.

P.S.
I think fpm is such "sensitive" because many developers must build software that dynamically links to libraries from these LTS distributions and thus the packaging build step runs on these LTS distributions as well.

jordansissel:
Folks are reporting that fpm cannot be installed easily (or at all) on older systems because a transitive dependency(1) rejects ruby versions older than 2.6.

(1) rubygem git depends on addressable which depends on public_suffix

Since the `git` dependency is only used in the `gem` source when using a git repo as a installation source, and that usage seems pretty simple -- clone a repo, checkout a branch, etc -- it feels safe to remove this dependency while still keeping the same functionality.

Fixes #1923

### DIFF
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -44,9 +44,6 @@ Gem::Specification.new do |spec|
   # For sourcing from pleaserun
   spec.add_dependency("pleaserun", "~> 0.0.29") # license: Apache 2
 
-  # For sourcing from git repos
-  spec.add_dependency("git", ">= 1.3.0", "< 2.0") # license: MIT
-
   spec.add_dependency("stud")
 
   # In Ruby 3.0, rexml was moved to a bundled gem instead of a default one,
--- a/lib/fpm/package/gem.rb
+++ b/lib/fpm/package/gem.rb
@@ -104,19 +104,18 @@ class FPM::Package::Gem < FPM::Package
     FileUtils.mkdir(download_dir) unless File.directory?(download_dir)
 
     if attributes[:gem_git_repo]
-      require "git"
       logger.debug("Git cloning in directory #{download_dir}")
-      g = Git.clone(attributes[:gem_git_repo],gem_name,:path => download_dir)
+      safesystem("git", "-C", download_dir, "clone", attributes[:gem_git_repo], ".")
       if attributes[:gem_git_branch]
-        g.branch(attributes[:gem_git_branch]).checkout
-        g.pull('origin',attributes[:gem_git_branch])
+        safesystem("git", "-C", download_dir, "checkout", attributes[:gem_git_branch])
       end
-      gem_build = [ "#{attributes[:gem_gem]}", "build", "#{g.dir.to_s}/#{gem_name}.gemspec"]
-      ::Dir.chdir(g.dir.to_s) do |dir|
+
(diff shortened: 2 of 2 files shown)


jordansissel:
To test this, I converted the `insist` gem from a git repo into a .deb package and checksum'd the file listing and package info:

```
% for i in main issues/1923; do echo Branch: $i; git checkout $i; bundle exec bin/fpm -f -s gem -t deb --gem-git-repo https://github.com/jordansissel/ruby-insist insist; echo -n "File listing: "; dpkg -c rubygem-insist_1.0.0_all.deb  |md5sum; echo -n "Package info: "; dpkg --info rubygem-insist_1.0.0_all.deb| md5sum; done
Branch: main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
Replacing 'provides' entry rubygem-insist = 1.0.0 with syntax 'name (= version)' {:level=>:warn}
Force flag given. Overwriting package at rubygem-insist_1.0.0_all.deb {:level=>:warn}
Created package {:path=>"rubygem-insist_1.0.0_all.deb"}
File listing: eaba7630b7634593ba180577d3515c9b  -
Package info: 0e52f3e53585cbc06abf859d7304fd86  -
Branch: issues/1923
Switched to branch 'issues/1923'
Replacing 'provides' entry rubygem-insist = 1.0.0 with syntax 'name (= version)' {:level=>:warn}
Force flag given. Overwriting package at rubygem-insist_1.0.0_all.deb {:level=>:warn}
Created package {:path=>"rubygem-insist_1.0.0_all.deb"}
File listing: 46eb53175cf80fce867a9be6f5c27dbf  -
Package info: 1648fee10413723d4c0683be9a4eb17d  -
```

File listing and package info is the same output on both branches.  Using `--debug` flag shows fpm invoking `git clone` command  on this PR but not on `main` branch.

jordansissel:
@liger1978 This change may impact your work on #1558/#1557. I did some tests to make sure the `--gem-git-repo` and `--gem-git-branch` still work correctly, but I might have missed something. Let me know if you run into any issues next time you try this :)

khagesh:
Are we planning to release 1.14.3 on rubygems? How should I install the latest version which has this fix?

jordansissel:
@khagesh I'm working on a release. Takes a bit to write the changelog <3

jordansissel:

khagesh:
Thank you so much @jordansissel !

PR #1948, opened 2022-10-30
Fixed bug in execmd: stdin_w_close -> stdin_w.close

NicholasBHubbard:
I believe you meant `stdin_w.close` but accidentally typed `stdin_w_close` in the `execmd` method.

### DIFF
--- a/lib/fpm/util.rb
+++ b/lib/fpm/util.rb
@@ -154,7 +154,7 @@ module FPM::Util
 
       yield(*args3)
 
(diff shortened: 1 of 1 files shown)


jordansissel:
Nice catch. I think this bug went unnoticed because this statement is never executed. The only place I could find a `execmd` used with a block is in the python.rb which closes stdin itself.

wbraswell:
Great work, everybody!  :-)

PR #1949, opened 2022-11-02
Bump minimum required version to meet required dependencies

edolnx:
While fpm may still stick to the conventions of Ruby 1.9, several of it's dependencies no longer support that release. I recently discovered this trying to build on an older ruby in a resource constrained system. The `json` gem requires ruby 2.3 or later and the `public_suffix` gem requires ruby 2.6 or later. So I'm proposing this PR to bump the minimum version requirement to 2.6.2 which seems to work without issue. Hopefully this will help someone in the future who is in the same boat I was in.

@jordansissel if you feel this is inappropriate, just let me know what you think would be a better approach. Happy to just make a PR to the README if that makes more sense.

### DIFF
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
     "management without wasting pointless hours debugging bad rpm specs!"
   spec.license = "MIT-like"
 
-  spec.required_ruby_version = '>= 1.9.3'
+  spec.required_ruby_version = '>= 2.6.2'
 
   # For parsing JSON (required for some Python support, etc)
   # http://flori.github.com/json/doc/index.html


jordansissel:
Hi Carl! I’m in favor of making the minimum version more accurate as you propose :)

I have some good news — I recently removed the dependency on public_suffix in #1946, and that should bring the minimum down to what the json gem requires.

as for json? Maybe we can figure out what minimum version of Ruby ships with json and go with that? I forget some of the history, but Ruby ships with json in the standard library, so it’s possible we don’t even need as an external gem dependency?

edolnx:
A quick search tells me that a version of `json` is bundled in 1.9.2, but for whatever reason when doing a `gem install fpm` it wants to pull a newer version. So I'm not sure what to make of that...

edolnx:
In fairness, I don't think it's a big deal that the `json` gem is being updated. The only frustration I had was the system I was trying to install on had no ruby and all I wanted to do was run `fpm`. So I started with 1.9.5 as per the `fpm.gemspec` and then got caught by dependency errors. So maybe just sticking with 2.3 is good enough? I can certainly re-run my tests since I have the older rubys installed via asdf-vm just for this reason once the new gem with #1946 is pushed.

jordansissel:
Looking at the history, the `json` gem was added in in The Ancient Times (63494e9096a1775ef82ce160b591e4ecd90d8c75, January 2011!). 

As a test, I tried removing the `json` dependency from the latest main branch (56a97c43be6e9cbdf6c2f5c64920312e623a644f) and things seem to work!

Test suite without `json` gem on Ruby 3.1.0

```
356 examples, 0 failures, 14 pending
```

Maybe we could try removing the `json` dependency also? Might be worth a try.

I have trouble testing very old Ruby versions because they dont' compile against newer versions of OpenSSL, etc, and I haven't tried using VMs.

jordansissel:
Here's the diff I used:

```diff
diff --git a/fpm.gemspec b/fpm.gemspec
index 1848da4..f4f8f04 100644
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|

   # For parsing JSON (required for some Python support, etc)
   # http://flori.github.com/json/doc/index.html
-  spec.add_dependency("json", ">= 1.7.7", "< 3.0") # license: Ruby License
+  #spec.add_dependency("json", ">= 1.7.7", "< 3.0") # license: Ruby License

   # For logging
   # https://github.com/jordansissel/ruby-cabin
```

If you can test, that'd be awesome. You can build your own fpm gem file by running `gem build fpm.gemspec` and it'll spit out a .gem file you can `gem install whatever.gem` as a local file; it'll still ask rubygems.org for help satisfying dependencies.

jordansissel:
Regrettably, I have taken this too far, and fortunately, I now have some fun news to share! I was able to get Ruby 1.9.3 to run fpm with only minor modifications.

Would that help?

edolnx:
Sorry it took so long to get back to you. Works fine in my environment. Thanks Jordan!

PR #1955, opened 2022-11-06
Execmd: prevent prompting for input

NicholasBHubbard:
This PR fixes issue https://github.com/jordansissel/fpm/issues/1519 and https://github.com/jordansissel/fpm/issues/1522.

I based my fix off [this commit](https://github.com/jordansissel/fpm/pull/1610/commits/ae9594eb098f74c55166729d317d86af83e3b3f9) from https://github.com/jordansissel/fpm/pull/1610 that was never merged.

Because Inline::CPP version 0.80 has an interactive Makefile.PL, the bug can be reproduced by running the following command on the main branch (this command will hang):
```
fpm --no-cpan-test --cpan-verbose --verbose --debug-workspace --workdir $HOME/tmp -t rpm -s cpan -v 0.80 Inline::CPP
```
Part of the issue in #1522 is that @wbraswell could not see the output from cpanminus, however this problem was fixed at some point. I added a test to `util_spec.rb` to ensure that the loggers `pipe` method is invoked during execution of `safesystem`. I would have liked to say `expect(logger).to receive(:pipe).with(instance_of(IO) => :info, instance_of(IO) => :info)`, but rspec does not allow you to use the `instance_of` matcher on the keys of a hash argument. Due to this limitation I am not able to ensure that `pipe` is called with the correct arguments, but this test should still be good enough.

### DIFF
--- a/lib/fpm/util.rb
+++ b/lib/fpm/util.rb
@@ -158,6 +158,9 @@ module FPM::Util
       stdout_r.close                unless stdout_r.closed?
       stderr_r.close                unless stderr_r.closed?
     else
+      # If no block given (not interactive) we should close stdin_w because we
+      # won't be able to give input which may cause a hang.
+      stdin_w.close
       # Log both stdout and stderr as 'info' because nobody uses stderr for
       # actually reporting errors and as a result 'stderr' is a misnomer.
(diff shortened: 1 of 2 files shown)


jordansissel (approved):

jordansissel:
Looks good. Thank you :)

Issue #1954, opened 2022-11-04
Buildable SRPM Support


wbraswell:
Soon fpm will support unbuildable SRPMs:
https://github.com/jordansissel/fpm/issues/237
https://github.com/jordansissel/fpm/pull/1657
https://github.com/jordansissel/fpm/pull/1952

What we really need is support for buildable SRPMs:
https://github.com/jordansissel/fpm/issues/237#issuecomment-415941828
https://github.com/bernd/fpm-cookery/issues/8#issuecomment-1302654409

This is a new issue created for adding buildable SRPM support, per request of @jordansissel:
https://github.com/jordansissel/fpm/pull/1657#issuecomment-1303911534

NicholasBHubbard:
If we were able to merge in the fpm-cookery "recipe" functionality, then we could use FPM to build any kind of source package, not just SRPM's.

jordansissel:
From the parent issue:
> An SRPM with an empty %build can not be built, and thus defeats the primary purpose of an SRPM existence.

these can still be built with rpmbuild, though, and empty %build sections is how I’ve used rpm specs for many years even prior to fpm. I do understand this may seem an unusual use, but the tooling didn’t complain (rpmbuild, mrepo, etc).

Can you describe your workflow and what tools are used? I’ll share one case from myself:

My first job where I used CentOS (~2007ish), we built rpm’s almost entirely with empty %build sections. This deployed dozens of applications to production systems and the workflow was mostly, as I recall:
* rpmbuild -bb on a .spec file to build a binary rpm. We configured it such that the spec only had a %files section and flags were set on rpmbuild to teach it where to get the files, similar to how fpm does it today.
* Copy the resulting rpm to the yum repository and have mrepo do its work to
* Install the rpms on servers with Puppet’s package provider fetching the package with the usual tooling from the central company repo.

For my scenario, I think the build steps (what rpm might call %prep, %build, etc) were run before the rpmbuild process. IIRC, this allowed us more freedom to use the developer’s build tooling and was a little easier to integrate than trying to replicate the build tooling/process in the rpm spec.

wbraswell:
We have our own internal Perl building tools, which have not yet been publicly released on CPAN etc.
Our goal is to generate automatically-buildable source packages via properly-populated ```%build``` data.
We do not want source packages with empty ```%build``` data.

[...]

PR #1963, opened 2022-11-23
Support for source based packages

NicholasBHubbard:
This PR is not meant to be merged!

This a prototype showing that FPM can support source based packages (#1954) with the purpose of furthering the discussion. This PR almost certainly contains bugs and is not feature complete in terms of producing SRPM's.

Given the following recipe:

```
[download]
https://cpan.metacpan.org/authors/id/E/ET/ETHER/File-Temp-0.2311.tar.gz

[build]
perl Makefile.pl
make
make test

[install]
make install
```
If we run the command:
```
bundle exec fpm --workdir $WORKDIR -t srpm -s cpan -r $RECIPE_DIR/recipe File::Temp
```
We get a SRPM with the following spec file:

```
Name: perl-File-Temp
Version: 0.2311
License: perl_5
Release: 1
Url: https://cpan.metacpan.org/authors/id/E/ET/ETHER/File-Temp-0.2311.tar.gz

Source0: https://cpan.metacpan.org/authors/id/E/ET/ETHER/File-Temp-0.2311.tar.gz

Summary: return name and handle of a temporary file safely

%description
return name and handle of a temporary file safely

%build
perl Makefile.pl
make
make test

%install
make install
```

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -51,6 +51,9 @@ class FPM::Command < Clamp::Command
   option ["-s", "--input-type"], "INPUT_TYPE",
     "the package type to use as input (gem, rpm, python, etc)",
     :attribute_name => :input_type
+  option ["-r", "--recipe"], "RECIPE_FILE",
+    "path to recipe file used for building a source based package",
+    :attribute_name => :recipe_file
   option ["-C", "--chdir"], "CHDIR",
     "Change directory to here before searching for files",
     :attribute_name => :chdir
@@ -311,7 +314,7 @@ class FPM::Command < Clamp::Command
       return 1
     end
     input_class = FPM::Package.types[input_type]
-    output_class = FPM::Package.types[output_type]
+    output_class = FPM::Package.types[output_type] || FPM::SourcePackage.types[output_type]
 
     input = input_class.new
 
@@ -470,7 +473,7 @@ class FPM::Command < Clamp::Command
     end
 
     # Convert to the output type
-    output = input.convert(output_class)
+    output = input.convert(output_class, recipe_file)
 
     # Provide any template values as methods on the package.
     if template_scripts?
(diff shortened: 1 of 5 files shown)


jordansissel:
I'd like to dig deeper on the 'recipe' concept. I'm feeling like this concept shouldn't be necessary for most input types and, instead, fpm could be taught how to do this.

Comparing an RPM SPEC to fpm, the difference is pretty large. With fpm, almost everything is automated -- you don't need to know how where to download the source, what steps are required to build/install, nor where things should install. In many cases, you don't even need to know how to find what the dependencies are. Contrast to SPEC files, for every single SPEC, you have to write the build steps, source file locations, dependencies, etc, and none of this is automated.

If we follow this recipe approach, the automation parts of fpm get lost and replaced by write-everything-by-hand which in most cases is a lot of copy-pasting. By way of example, we can compare the steps required to use this example and create a second example to package Regexp::Common.

For these recipes, we can copy the existing File::Temp recipe and make the following modifications:
* We must change the `Name` field
* We must look up the correct URL for the source files as hosted on CPAN
* We must look up the correct version we want to use
* We must write our own %description and Summary fields

Likely unchanged:
* The %build and %install steps are likely unchanged

In practice, it feels like these recipes would require the same cognitive and source control burdens as rpm spec files, and I'd like to have fpm reduce cognitive loads where possible.

If we compare this experience to fpm's general use, a user wanting to package File::Temp can use a nearly identical command-line for any other Perl module, such as:

* `fpm -s cpan -t deb File::Temp`
* `fpm -s cpan -t deb Regexp::Common`

In both of the above cases, fpm defaults to the "latest" version and inspects the cpan content to determine the package name, version number, dependencies, description, etc. Note how few changes are required in the above example as compared to the recipe example.

If we think about extending fpm to buildable (with 'standard tooling'[1]) source package formats like SRPM, then I would expect that something like the following should produce an SRPM, given some hypothetical code implementation that makes this work:

* `fpm -s cpan -t srpm File::Temp`
* `fpm -s cpan -t srpm Regexp::Common`

I don't think a recipe should be required in these kinds of cases where fpm could be taught this knowledge.

[1] 'standard tooling' refers to whatever the common platform tools for building packages are, such as `rpmbuild --rebuild` or `debuild`

jordansissel:
I forgot to say -- thanks for working on this prototype! It helps me understand what kind of user interface and experience (recipes, new flags, etc) you're thinking about.

NicholasBHubbard:
Deprecated in favor of the #1967 prototype

PR #1967, opened 2022-11-30
Experiment to create an srpm output.

jordansissel:
[ Not intended for merging at this time ]

Currently only supports cpan-to-srpm

Ideas for #1954

### DIFF
--- a/lib/fpm/package/srpm.rb
+++ b/lib/fpm/package/srpm.rb
@@ -1,0 +1,243 @@
+require "fpm/namespace"
+require "fpm/package"
+require "fpm/errors"
+require "fpm/util"
+
+# For handling conversion
+require "fpm/package/cpan"
+
+class FPM::Package::SRPM < FPM::Package
+  def output(output_path)
+    source_archive = ::Dir.glob(build_path("*")).select(&File.method(:file?)).first
+    source_archive_dirname = `tar -ztf #{Shellwords.escape(source_archive)}` \
+      .split("\n").map { |path| path.split("/").first }.uniq.first
+
+    # Generate an rpm spec with Source0: <source_archive>
+    rpmspec = template("srpm.erb").result(binding)
+
+    # Copied from rpm.rb ---
+    %w(BUILD RPMS SRPMS SOURCES SPECS).each { |d| FileUtils.mkdir_p(build_path(d)) }
+    args = ["rpmbuild", "-bs"]
+
+    if %x{uname -m}.chomp != self.architecture
+      rpm_target = self.architecture
+    end
+
+    # issue #309
+    if !attributes[:rpm_os].nil?
+      rpm_target = "#{architecture}-unknown-#{attributes[:rpm_os]}"
+    end
+
+    # issue #707
+    if !rpm_target.nil?
+      args += ["--target", rpm_target]
+    end
+
+    # set the rpm dist tag
+    args += ["--define", "dist .#{attributes[:rpm_dist]}"] if attributes[:rpm_dist]
+
+    args += [
+      "--define", "buildroot #{build_path}/BUILD",
+      "--define", "_topdir #{build_path}",
+      "--define", "_sourcedir #{build_path}",
+      "--define", "_rpmdir #{build_path}/RPMS",
+      "--define", "_tmppath #{attributes[:workdir]}"
+    ]
+
+    args += ["--sign"] if attributes[:rpm_sign?]
+
+    if attributes[:rpm_auto_add_directories?]
+      fs_dirs_list = File.join(template_dir, "rpm", "filesystem_list")
+      fs_dirs = File.readlines(fs_dirs_list).reject { |x| x =~ /^\s*#/}.map { |x| x.chomp }
+      fs_dirs.concat((attributes[:auto_add_exclude_directories] or []))
+
+      Find.find(staging_path) do |path|
+        next if path == staging_path
+        if File.directory? path and !File.symlink? path
+          add_path = path.gsub(/^#{staging_path}/,'')
+          self.directories << add_path if not fs_dirs.include? add_path
+        end
+      end
+    else
+      self.directories = self.directories.map { |x| self.prefixed_path(x) }
+      alldirs = []
+      self.directories.each do |path|
+        Find.find(File.join(staging_path, path)) do |subpath|
+          if File.directory? subpath and !File.symlink? subpath
+            alldirs << subpath.gsub(/^#{staging_path}/, '')
+          end
+        end
+      end
+      self.directories = alldirs
+    end
+
+    # include external config files
+    (attributes[:config_files] or []).each do |conf|
+      dest_conf = File.join(staging_path, conf)
+
+      if File.exist?(dest_conf)
+        logger.debug("Using --config-file from staging area", :path => conf)
+      elsif File.exist?(conf)
+        logger.info("Copying --config-file from local path", :path => conf)
+        FileUtils.mkdir_p(File.dirname(dest_conf))
+        FileUtils.cp_r conf, dest_conf
+      else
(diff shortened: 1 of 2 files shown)


lib/fpm/package/srpm.rb
```diff
@@ -0,0 +1,243 @@
+require "fpm/namespace"
+require "fpm/package"
+require "fpm/errors"
+require "fpm/util"
+
+# For handling conversion
+require "fpm/package/cpan"
+
+class FPM::Package::SRPM < FPM::Package
```
jordansissel, line 9: Lots of this file is copied directly from `rpm.rb` (the FPM::Package::RPM class).  I'd like to reduce code duplication if we can, such as the duplicate methods for architecture, to_s, etc, and the nearly identical `output` method.

lib/fpm/package/srpm.rb
```diff
@@ -0,0 +1,243 @@
...
+        # Fun hack to find the instance of the origin class
+        # So we can find the build_path
+        input = nil
+        ObjectSpace.each_object { |x| input = x if x.is_a?(origin) }
+        if input.nil?
+            raise "Something bad happened. Couldn't find origin package in memory? This is a bug."
+        end
+
+        # Pick the first file found, should be a tarball.
+        source_archive = ::Dir.glob(File.join(input.build_path, "*")).select(&File.method(:file?)).first
```
jordansissel, line 151: This trick is all so we can find the original source archive that came from cpan because the 'output' is not told anything about it. If we move forward, we'll want a better way to express these things from input to output conversions, such as the cpan input being able to say "My source file is /some/path/here"

lib/fpm/package/srpm.rb
```diff
@@ -0,0 +1,243 @@
...
+class FPM::Package::SRPM < FPM::Package
+  def output(output_path)
+    source_archive = ::Dir.glob(build_path("*")).select(&File.method(:file?)).first
+    source_archive_dirname = `tar -ztf #{Shellwords.escape(source_archive)}` \
+      .split("\n").map { |path| path.split("/").first }.uniq.first
+
+    # Generate an rpm spec with Source0: <source_archive>
+    rpmspec = template("srpm.erb").result(binding)
+
+    # Copied from rpm.rb ---
```
jordansissel, line 18: Most of the below is copied directly from `rpm.rb` with small modifications, for example invoking `rpmbuild -bs` to request only a source rpm.

lib/fpm/package/srpm.rb
```diff
@@ -0,0 +1,243 @@
...
+      return attributes[:prefix].chomp('/')
+    else
+      return "/"
+    end
+  end # def prefix
+
+  def to_s(format=nil)
+    if format.nil?
+      format = if attributes[:rpm_dist]
+        "NAME-VERSION-ITERATION.DIST.src.rpm"
```
jordansissel, line 177: This method is copied and slightly modified from `rpm.rb` where the extension is now hard-coded to '.src.rpm'

templates/srpm.erb
```diff
@@ -0,0 +1,233 @@
+# Hello packaging friend!
```
jordansissel, line 1: Much of this file is copied and slightly modified from the original rpm.erb

templates/srpm.erb
```diff
@@ -0,0 +1,233 @@
...
+<%# The closes equivalent in RPM to "replaces" is "Obsoletes" -%>
+Obsoletes: <%= repl %>
+<% end -%>
+<%# rpm rejects descriptions with blank lines (even between content), so hack
+    around it by replacing blank lines with ' .' -%>
+%description
+<%= description.gsub(/^\s*$/, " .") %>
+
+%prep
+%autosetup -n <%= source_archive_dirname %>
```
jordansissel, line 84: rpm assumes the source archive contains a directory that matches the same name as the package itself, such as `perl-File-Temp-0.2311/`, but things from cpan are more typically containing a path named after the module itself and not following redhat's naming scheme.

The source_archive_dirname is a variable that contains the first toplevel directory within the source tarball. It seems to work well enough for %autosetup and %setup to do their jobs.

templates/srpm.erb
```diff
@@ -0,0 +1,233 @@
...
+
+%build
```
jordansissel, line 86: The %build section was hand-written loosely based on what the cpan input already does.

I probably still need to set `-Mlocal::lib=./cpan` when invoking perl, though.

templates/srpm.erb
```diff
@@ -0,0 +1,233 @@
...
+mkdir -p ./cpan
+cpanm -L ./cpan . --installdeps
+
+perl Makefile.PL
+make
+
+%install
+%make_install
```
jordansissel, line 94: This uses rpm's %make_install macro. This will work for some projects, but not most, especially ones which don't use `make`

templates/srpm.erb
```diff
@@ -0,0 +1,233 @@
...
+<%# Output config files and then regular files. -%>
+<% config_files.each do |path| -%>
+%config(noreplace) <%= rpm_file_entry(path) %>
+<% end -%>
+<%# list directories %>
+<% directories.each do |path| -%>
+%dir <%= rpm_file_entry(path) %>
+<% end -%>
+<%# list only files, not directories? -%>
+/
```
jordansissel, line 230: Since we're building an rpm from source with rpmbuild, we might be able to get away with listing `/` as the %files entry and including all files that end up in the %buildroot.

This is mostly a hack at this time to avoid having to do additional file list generation in this prototype.

templates/srpm.erb
jordansissel: Notably, this specfile only works with cpan input and would fail everything else.

For this to go forward, we'd to codify the prep/build/install steps for other inputs (gem, npm, pear, etc).

jordansissel (commented):
I've annotated a few areas of the prototype

lib/fpm/package/srpm.rb
```diff
@@ -0,0 +1,243 @@
...
+    safesystem(*args)
+
+    ::Dir["#{build_path}/SRPMS/**/*.rpm"].each do |rpmpath|
+      # This should only output one rpm, should we verify this?
+      FileUtils.cp(rpmpath, output_path)
+    end
+  end
+
+  def converted_from(origin)
+    if origin ==  FPM::Package::CPAN
```
jordansissel, line 141: We'll probably need a mapping of { input type => instructions }, where the instructions are the appropriate %prep, %build, %install, etc steps.

I'm not sure this can be generalized to other output targets, such as Debian source packages, Arch BUILD files, FreeBSD Ports, Nix files, and other kinds of package-from-source systems. We don't have to target everything at once, but it's good to think about ways the work can be reused when targeting things besides Red Hat packages.

templates/srpm.erb
jordansissel: I have some ideas, but we'll need to study the existing inputs (rpm, deb, gem, npm, et al) to see what common elements exist between them.

NicholasBHubbard:
Thanks for creating this prototype!

I understand now that it is possible to automatically generate build steps for certain package input types. I would like to try to extend this prototype to support another input type (Ruby gems) and another output type (deb source packages). I want to see if I can generalize the prototype so in the future FPM can support any source-based package.

Would you please give me write access to this branch?

wbraswell:
@jordansissel 
I am in favor of @NicholasBHubbard working on this branch with you.

jordansissel:
> Would you please give me write access to this branch?

GitHub doesn't have access controls per branch. I'm open to working on a fork separately or we can find some other way to work on this?

NicholasBHubbard:
@jordansissel: I created #1972 which is an exact copy of this prototype, and added you to my fork. We should now both be able to write to my forks "source-based-package-support" branch (which is the branch for #1972).

Issue #1847, opened 2021-11-06
Reject invalid version strings for debian packages.


jordansissel:
While writing docs tonight, I found that fpm can create a debian package with an invalid version string.

```
% fpm -s empty -t deb -v "v1.2.3" -n example
Created package {:path=>"example_v1.2.3_amd64.deb"}

% dpkg --field example_v1.2.3_amd64.deb Version
dpkg-deb: error: parsing file '/tmp/dpkg-deb.AlPRzj/control' near line 2 package 'example':
 'Version' field value 'v1.2.3': version number does not start with digit
```

mthbrown:
Just ran into this as well

jordansissel:
Based on #1968 and this issue, I'd like two behaviors:

1. If the version string is something like 'v1.2.3' then fpm should drop the 'v' and use '1.2.3' for the version field
2. Excluding the item above, if it's an invalid version string, fpm should log an actionable error message and exit with failure.

Item 1 (vX.Y.Z) should help with automatic versioning that some folks do with Git tags where a release is tagged vX.Y.Z for a version number.

PR #1969, opened 2022-12-02
Drop leading v from version for Debian

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

#1847

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -292,6 +292,15 @@ class FPM::Package::Deb < FPM::Package
     return (attributes[:prefix] or "/")
   end # def prefix
 
+  def version
+    if @version.kind_of?(String) and @version.start_with?("v")
+      logger.warn("Drop leading v from package version '#{@version}'")
+      @version = @version.gsub(/^v/, "")
+    end
+
+    return @version
+  end
+
   def input(input_path)
     extract_info(input_path)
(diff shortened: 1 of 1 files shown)


jordansissel:
Thanks for working on this!

I think there's a few other places that use the 'version' field. I recommend, instead of modifying the deb.changes template, overriding the 'version' method as fpm does in the rpm.rb, shown here:

https://github.com/jordansissel/fpm/blob/db06af3c0357e212df48009ecd73c945519a1e76/lib/fpm/package/rpm.rb#L598-L606

Such a method `def version ... end` would go into `lib/packages/fpm/deb.rb`. This would allow any corrections to the version field to be used everywhere (the control file, deb.changes, etc)

gliptak:
@jordansissel thank you for the pointer. updated

PS I'm not setup to test the project locally

jordansissel:
> PS I'm not setup to test the project locally

No worries! I'll take care of testing :)

jordansissel:
Looks good! Thank you :)

jordansissel:
```
% fpm -s empty -t deb -v v1.2.3  -n example
Drop leading v from package version 'v1.2.3' {:level=>:warn}
Created package {:path=>"example_1.2.3_all.deb"}

# Verify the version
% dpkg -f example_1.2.3_all.deb Version
1.2.3
```

PR #1970, opened 2022-12-04
Update GHA build

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

https://github.com/gliptak/fpm/actions/runs/3611226074

After getting `virtual_spec.rb` green, `virtualenv` to be installed

### DIFF
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -1,8 +1,0 @@
-workflow "New workflow" {
-  on = "push"
-  resolves = ["GitHub Action for Docker"]
-}
-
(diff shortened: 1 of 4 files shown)


jordansissel:
I'm not quite sure I'm ready to use GHA yet, but it's worth a shot, I suppose! 

Using Travis for some years made more trouble than it was worth when trying to debug flakey tests that /only/ failed on Travis. I don't expect GHA to be any easier. I'll disable GHA if this becomes a problem.

Thank you for your improvements here :)

PR #1973, opened 2022-12-11
Upgrades to cpan_spec.rb

NicholasBHubbard:
I upgraded `cpan_spec.rb` so every method defined in `cpan.rb` has its own test.

I upgraded the existing tests to explicitly set package versions to avoid the scenario that a new version of a CPAN distribution is released that breaks the test.

I also removed the stuff related to Travis CI as FPM now uses GitHub Actions.

On my machine these changes make the test suite take 5 seconds longer to run.

Let me know what you think!


jordansissel:
Thanks again for improving things! The test suite passes for me now :)

PR #1974, opened 2022-12-12
Minor comment cleanup

NicholasBHubbard:
This PR fixes up some of the `end # def method_name` comments in cpan.rb, package.rb, command.rb, gem.rb, and util.rb.

Now each of these files uses this comment convention for all of their methods. There were also a few of these comments with an incorrect function name that I fixed.

No non-comment code is changed in this PR.

### DIFF
--- a/lib/fpm/package.rb
+++ b/lib/fpm/package.rb
@@ -227,7 +227,7 @@ class FPM::Package
   def converted_from(origin)
     # nothing to do by default. Subclasses may implement this.
     # See the RPM package class for an example.
-  end # def convert_from
+  end # def converted_from
 
   # Add a new source to this package.
   # The exact behavior depends on the kind of package being managed.


wbraswell:
Howdy @jordansissel , long time no see!  :-)
My old friend @NicholasBHubbard and I are back at it again - we have been carefully reviewing all of our old FPM pull requests including this one, and we have updated it to be in a state which we believe is ready for your approval.
This particular PR only fixes comments, but some of them are actually incorrect so personally I do suggest that we fix it.
Thanks for your time and we look forward to working with you on the other pull requests as well!

jordansissel:
Looks good! Thanks for your effort tidying things up :)

wbraswell:
@jordansissel 
Wonderful, thank you sir!  :-)

NicholasBHubbard:
@jordansissel Thanks!

PR #1976, opened 2022-12-18
lintian fixes, add triggers

morph027:
backport of

* https://github.com/root360/fpm/tree/1693_implement_triggered
* https://github.com/root360/fpm/tree/1575_fix_changelog_name
 
Signed-off-by: morph027 <stefan.heitmueller@gmx.com>

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -213,6 +213,13 @@ class FPM::Command < Clamp::Command
     File.expand_path(val) # Get the full path to the script
   end # --before-upgrade
 
+  option "--triggered", "FILE",
+    "A script snippets to define and run package triggers, see\n" \
+    "https://wiki.debian.org/DpkgTriggers and\n" \
+    "https://stackoverflow.com/questions/15276535/dpkg-how-to-use-trigger\n" do |val|
+    File.expand_path(val) # Get the full path to the script
+  end # --triggered
+
   option "--template-scripts", :flag,
     "Allow scripts to be templated. This lets you use ERB to template your " \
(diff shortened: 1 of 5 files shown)


lib/fpm/command.rb
```diff
@@ -213,6 +213,13 @@ def help(*args)
     File.expand_path(val) # Get the full path to the script
   end # --before-upgrade
 
+  option "--triggered", "FILE",
```
jordansissel, line 216: Trying to figure out if this flag is best put in `package/deb.rb` instead of this file. Some questions to think about?
Are triggers common to multiple package systems? Or just Debian?

If they're common, is the implementation common-enough across multiple package systems? That is, would the same flag work for both Debian and rpm packages?

morph027: Hm, looks like rpm has something too: https://jfearn.fedorapeople.org/en-US/RPM/4/html/RPM_Guide/ch05s03.html

Not sure if it works the same way.

morph027:
@jordansissel i've moved the flag to a deb-specific flag.

morph027:
Hm, moving to `--deb-trigger` does not not work somehow. I'm not getting `script?(:triggered)` (evals to false all the time).

morph027:
is it missing in `SCRIPT_MAP`?

morph027:
Ah, got it working. Also added the fixes from https://github.com/jordansissel/fpm/pull/1620

plauche:
@morph027 @jordansissel I'm very interested in seeing the `deb-after-purge` fix from this PR land. Is there any remaining work that I can help with to get this over the line?

PR #1981, opened 2022-12-29
Replace deprecated {File,Dir}.exists? with {File,Dir}.exist?

NicholasBHubbard:
Ruby 3.2.0 was just released, and the deprecated methods `File.exists?` and `Dir.exists?` were completely removed. These methods are used in FPM, so FPM is currently broken under Ruby 3.2.0. Instead we are supposed to use the `File.exist?` and `Dir.exist?` methods. This PR replaces all occurrence of these deprecated methods with their preferred counterpart.

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -298,7 +298,7 @@ class FPM::Command < Clamp::Command
       args << "."
     end
 
-    if !File.exists?(workdir)
+    if !File.exist?(workdir)
       logger.fatal("Given --workdir=#{workdir} is not a path that exists.")
       raise FPM::Package::InvalidArgument, "The given workdir '#{workdir}' does not exist."
     end
@@ -371,7 +371,7 @@ class FPM::Command < Clamp::Command
 
     # If --inputs was specified, read it as a file.
     if !inputs.nil?
-      if !File.exists?(inputs)
+      if !File.exist?(inputs)
         logger.fatal("File given for --inputs does not exist (#{inputs})")
         return 1
       end
@@ -386,7 +386,7 @@ class FPM::Command < Clamp::Command
     # If --exclude-file was specified, read it as a file and append to
     # the exclude pattern list.
     if !exclude_file.nil?
-      if !File.exists?(exclude_file)
+      if !File.exist?(exclude_file)
         logger.fatal("File given for --exclude-file does not exist (#{exclude_file})")
         return 1
       end
@@ -451,7 +451,7 @@ class FPM::Command < Clamp::Command
       # Skip scripts not set
       next if path.nil?
 
-      if !File.exists?(path)
+      if !File.exist?(path)
         logger.error("No such file (for #{scriptname.to_s}): #{path.inspect}")
         script_errors << path
       end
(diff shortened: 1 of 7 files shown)


jordansissel:
Thanks for the patch.
I’m still taking care of sick family and will be unavailable for an unknown amount of time.

ranjib:
Take care @jordansissel . Family is always first, 

for anywho is facing this with ruby upgrade,  update your gemfile like this
```
gem 'fpm', git: 'https://github.com/NicholasBHubbard/fpm.git', branch: 'replace-deprecated-methods'
```

and thank you @NicholasBHubbard for the patch.

romulasry:
gem 'fpm', git: 'https://github.com/NicholasBHubbard/fpm.git', branch: 'replace-deprecated-methods'
ERROR:  While executing gem ... (Gem::UnknownCommandError)
    Unknown command fpm,
	/usr/local/share/ruby/site_ruby/rubygems/command_manager.rb:204:in `find_command'
	/usr/local/share/ruby/site_ruby/rubygems/command_manager.rb:249:in `invoke_command'
	/usr/local/share/ruby/site_ruby/rubygems/command_manager.rb:191:in `process_args'
	/usr/local/share/ruby/site_ruby/rubygems/command_manager.rb:149:in `run'
	/usr/local/share/ruby/site_ruby/rubygems/gem_runner.rb:51:in `run'
	/usr/bin/gem:10:in `<main>'

romulasry:
It would be nice to push an update after this is in because it is broken on latest ruby.

jordansissel:
Fixing newer versions is good, but we gotta check if this breaks older versions. Picking my favorite ancient version of ruby to see how long ago these methods have existed, it appears that `File::exist?` and `Dir::exist?` have been in Ruby since at least Ruby 1.9.3. Citations:

* https://ruby-doc.org/core-1.9.3/Dir.html#method-c-exist-3F
* https://ruby-doc.org/core-1.9.3/File.html#method-c-exist-3F

I'm OK to replace the methods as proposed.

jordansissel:
Checking dependencies:
* pleaserun depends on mustache 0.99.8 which still invokes `File.exists?`
* clamp dependency is pinned to `~> 1.0.0` and v1.0.1 invokes `Dir.exists?`

jordansissel:
I haven't tried running the test suite yet, so I don't know if anything fails on Ruby 3.2.0 (fpm's GitHub Actions also doesn't use 3.2 at this time).

One option which might help without causing too much of a stir in fpm's dependencies: we could monkey patch File and Dir to add back these removed methods. It's possible to attend to these dependencies with upgrades, but we might find upgrading them causes older versions of ruby to fail, even very recent versions of ruby.

jordansissel:
reviewing dependencies further -- https://github.com/jordansissel/fpm/pull/1981#issuecomment-1411330564

* mustache invokes File.exists? in a method not used by fpm or pleaserun
* clamp invokes Dir.exists? in a development tool (guard) and doesn't affect fpm itself.

romulasry:
> I haven't tried running the test suite yet, so I don't know if anything fails on Ruby 3.2.0 (fpm's GitHub Actions also doesn't use 3.2 at this time).

Should be able to fix that here: https://github.com/jordansissel/fpm/blob/main/.github/workflows/ruby.yml

jordansissel:
v1.15.1 released with this patch

NicholasBHubbard:
Thanks Jordan!

ranjib:
Thank you Jordan. Much appreciated

PR #1986, opened 2023-01-22
Fix typos

kianmeng:
Found via `codespell -S templates -L claus`


jordansissel (changes requested):
I did a quick scan and found one issue the spell checker created. I'll need to do a more thorough read later.

PR #1994, opened 2023-02-07
Fix failing test when run on arm64/aarch64 systems 

jordansissel:
This was a bug in the test suite.

```
  1) FPM::Package::Deb#architecture should default to native
     Failure/Error: expect(subject.architecture).to(be == expected)
       expected: == "aarch64"
            got:    "arm64"
     # ./spec/fpm/package/deb_spec.rb:57:in `block (3 levels) in <top (required)>'
```


mkurz:
New release needed please.

PR #2001, opened 2023-03-22
fix building of python modules that have tarballs named .tgz

bugfood:
This fixes building of (at least):
https://pypi.org/project/amqplib/1.0.2/

Wrap the operation within a Dir.chdir() to simplify the use of multiple globs (older ruby does not support the ':base' option to Dir.glob()).


bugfood:
Please

bugfood:
Please disregard in favor of:
https://github.com/jordansissel/fpm/pull/2002

PR #2002, opened 2023-03-22
fix building of python modules that have tarballs named .tgz and .zip

bugfood:
This fixes building of (at least):
https://pypi.org/project/amqplib/1.0.2/
https://pypi.org/project/dnspython/1.15.0/

Wrap the operation within a Dir.chdir() to simplify the use of multiple globs (older ruby does not support the ':base' option to Dir.glob()).


jordansissel:
Thanks for reporting and solving this!

This is now fixed in PR #2104 and I credited you in the commit message - https://github.com/jordansissel/fpm/commit/c76917a4e193ea7555fd16c63a8932154eda43cd

bugfood:
> Thanks for reporting and solving this!
> 
> This is now fixed in PR #2104 and I credited you in the commit message - [c76917a](https://github.com/jordansissel/fpm/commit/c76917a4e193ea7555fd16c63a8932154eda43cd)

Aha, thank you.

-Corey

Issue #2007, opened 2023-04-24
Unable to repackage deb packages using zstd compression


SaltwaterC:
zstd is a [newer addition](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=892664) (if 5 years counts as newer) to the deb package format which in the mean time made its way into stable/LTS distributions.

Unfortunately fpm chokes on these packages with:

```
Invalid package configuration: Unknown compression type 'zst' for control.tar in deb source package foo.deb {:level=>:error}
```

This makes the effing package management in this instance a tad more difficult to use. zstd is supported for both control and data archives after taking a peek with `ar x`.

PR #2009, opened 2023-04-28
Add support for zstd compression for deb packages.

SaltwaterC:
This PR adds what it says on the tin to make fpm compatible with recent Debian and derivatives which implement zstd compression for their packages.

Fixes https://github.com/jordansissel/fpm/issues/2007.

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -27,7 +27,7 @@ class FPM::Package::Deb < FPM::Package
   } unless defined?(SCRIPT_MAP)
 
   # The list of supported compression types. Default is gz (gzip)
-  COMPRESSION_TYPES = [ "gz", "bzip2", "xz", "none" ]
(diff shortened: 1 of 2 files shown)


jordansissel:
This was kinda funny for me to test since it fails on macOS and I forgot I was testing on this OS at the time ;) I dug a little into this, because I'd like it to work on macOS and Linux. 

macOS defaults to bsdtar and `-I` is a synonym for `-T` flag which means something different than GNU tar's `-I` flag. Both appear to support `--zstd` flags, though. It's possible `--zstd` will fail on older tar versions. 

I do see that both bsdtar (on macOS) and GNU tar seem to agree on `--use-compress-program` but they both disagree on what this actually  means. GNU tar assumes the program will be invoked with `-d` and bsdtar does not do this.

Both work with `--use-compress-program 'zstd -d'`

I'll create a new branch using your changes and the above additions :)

jordansissel:
Thanks for improving fpm :)

PR #2017, opened 2023-07-10
pacman.rb: fix aarch64 & arm7hf

MarksonHon:


jordansissel:
Thank you for improving fpm :)

PR #2023, opened 2023-08-30
chore: relax 'clamp' version pin

cwegener:
Ubuntu 22.04 LTS comes with ruby-clamp 1.1.1 which seems to work just
fine with fpm 1.15.1

Relaxing the version pin seems to be the right thing to do.

### DIFF
--- a/fpm.gemspec
+++ b/fpm.gemspec
@@ -35,7 +35,7 @@ Gem::Specification.new do |spec|
 
   # For command-line flag support
   # https://github.com/mdub/clamp/blob/master/README.markdown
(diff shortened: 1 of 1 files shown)


captn3m0:
Similar issues with arch linux as well, trying to install the development branch.

jordansissel:
I reviewed some of the history and didn’t see any reason ~>1.0.0 was necessary, and it might have been a relic of the pre-1.0 clamp versions.

👍👍 looks good

cwegener:
Thanks Jordan.

There also is another one over here. I've been using that patch on Ubuntu 22.04 LTS for 6+ months now.

https://github.com/jordansissel/pleaserun/pull/156

PR #2027, opened 2023-10-07
deb: A local path copied with --config-files should still be included in the package's conffiles list (Fixes #1823)

### LINKED ISSUE #1823: Conffiles with python sources to deb package
Hello,

My question may sound stupid, but I'm not able to generate a deb package with config files in `/etc/`.
As consequences, each upgrade overwrite the files in `/etc/`.

For the context, I have a directory `src` containing a python module which I want to package, a directory `debian` for some stuff, and I use the following command : 

```
fpm -s python \
    -t deb \
    --python-bin python3 \
    -p build/myapp_VERSION_ARCH.deb \
    -x "*.pyc" \
    -x "*/__pycache__" \
    --no-auto-depends \
    -d build-essential \
    -d python3-pip \
    -d python3-dev \
    --deb-auto-config-files \
    --config-files ./etc/myapp/config \
    --deb-changelog CHANGELOG.txt \
    --after-install debian/scripts/postinstall.sh \
    --deb-systemd debian/systemd/myapp.service \
    --deb-systemd-enable \
    --deb-systemd-auto-start \
    src
```

With this command, the package is generated, I can install it without error, service in systemd is well configured, the file `/etc/myapp/config` is present but it will be overwritten with a next version.
I tried the option `--deb-config` or to put the directory `/etc/` in another place, but without success.

The resulted deb package does not contain any `conffiles`.
Am I missing something obvious ?

me21:
Fix #1823.

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -1068,6 +1068,7 @@ class FPM::Package::Deb < FPM::Package
             logger.debug("Adding config file #{path} to Staging area #{staging_path}")
             FileUtils.mkdir_p(File.dirname(dcl))
             FileUtils.cp_r path, dcl
+            add_path(path, allconfigs)
           else
             logger.debug("Config file aready exists in staging area.")
           end


jordansissel:
This code fixes #1823, but the PR and commit description isn't correct. I haven't looked at this section of fpm in a loooong time, but I'll add notes here for my future self (and others)

The bug:
* When `--config-files <path>` is given
* and that path doesn't already exist in the staging area
* then it is assumed <path> is a local path, and it is copied into the staging area (PR #1443)
* however (the bug), fpm forgets to add this path to the list of config files
* so this file is unintentionally excluded from the deb package's "conffiles"

I'd love tests covering this behavior, but given the age of this PR and the simplicity, I'm OK merging.

Here's manual tests:

fpm main branch which has the bug:
```
% bundle exec bin/fpm -f -s empty -t deb -n example --prefix /my/prefix --config-files .gitignore
...
# The file is correctly copied into the package:
% ar p example_1.0_all.deb data.tar.gz | tar -zt | grep -F .gitignore
./.gitignore

# But it is not listed in confffiles (in this case, no conffiles at all)
% ar p example_1.0_all.deb control.tar.gz | tar -zxO ./conffiles
tar: ./conffiles: Not found in archive
```

This PR:

```
# Same fpm invocation as above
% ar p example_1.0_all.deb data.tar.gz | tar -zt | grep -F .gitignore
./.gitignore

# The package's conffiles lists the given --config-files entry as expected
% ar p example_1.0_all.deb control.tar.gz | tar -zxO ./conffiles
/.gitignore
```

PR #2036, opened 2023-11-24
deb: add a --compression-level option

### LINKED ISSUE #2006: Add --deb-compression-level
The sizes of deb and rpm packages can differ greatly because the default compression levels for deb and rpm packages are different:
- RPMs use level 9
- DEBs use level 6 which is used by [tar that creates `data.tar.xz`](https://github.com/jordansissel/fpm/blob/b085edc49c8dd027db6ef3e32d8de531894a26ea/lib/fpm/package/deb.rb/#L661-L668).

Unlike RPMs (`--rpm-compression-level`), there is no way to specify a compression level for DEBs.
This could be achieved by either adding an environment variable to tar, such as `XZ_OPT` for xz, or by first creating a tar archive and then compressing it.

chouquette:
Fix #2006

This was manually tested, but please let me know if you'd wish to add some automated tests. I'm very new to the ruby ecosystem so I'm a bit at lost :sweat_smile:

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -79,6 +79,19 @@ class FPM::Package::Deb < FPM::Package
     value
   end
 
+  option "--compression-level", "[0-9]", "Select a compression level. 0 is none or minimal. 9 is max compression.",
+    # Specify which compression level to use on the compressor backend, when building a package
+    :default => nil do |value|
+    valint = value.to_i
+    # if self.attributes[:deb_compression].nil?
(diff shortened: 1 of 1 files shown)


jordansissel:
@chouquette Looks pretty good! I have some changes I'll make after merging to tweak a few things. Thank you for making fpm better :)

jordansissel:
Looks like I didn't handle the merge conflict correctly (my fault) and so the tests failed. I've manually merged this into main now.

PR #2041, opened 2023-12-14
rpm: generate changelog if none is provided

UiP9AV6Y:
`rpmlint` considers a lack of a changelog an error.

this is a similar behaviour to the DEB packager, where a generic default changelog is generated unless one is provided explicitly.

### DIFF
--- a/lib/fpm/package/rpm.rb
+++ b/lib/fpm/package/rpm.rb
@@ -275,6 +275,24 @@ class FPM::Package::RPM < FPM::Package
     return @iteration ? @iteration : 1
   end # def iteration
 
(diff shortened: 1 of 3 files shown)


jordansissel:
Thank you for improving fpm! I especially appreciate the extra effort to add tests to cover this. Tests are passing and I checked manually:

```
% bundle exec bin/fpm -s empty -t rpm -n example
Created package {:path=>"example-1.0-1.noarch.rpm"}

% rpm -qp example-1.0-1.noarch.rpm --changelog
* Sat Dec 07 2024 <jls@baguette.local> - 1.0-1
- Package created with FPM
```

PR #2047, opened 2024-02-09
[Docs} Remove target-specific options from general options

aaronliu0130:
These are later duplicated in their own sections and clog up the list of general options.


jordansissel:
Ahh, I think I see what happens here - thank you for this improvement!

This file is generated by the documentation building process, so modifying this file directly won't survive very long.

I did a little bit of digging and found this - The docs [generator iterates over every flag](https://github.com/jordansissel/fpm/blob/main/docs/generate-cli-reference.rb#L15-L34) for the general options.  However, this class (FPM::Command) adds [all package flags to itself on startup](https://github.com/jordansissel/fpm/blob/847ded9eba0d707c61c942f6153178de357740b7/lib/fpm/command.rb#L263-L265).

To fix this, maybe having the docs generator skip any package-specific flags for the 'general' section. One way may be to check if the flag's "help" text includes something like `(deb only)` at the beginning as all package-specific flags will do this. https://github.com/jordansissel/fpm/blob/main/lib/fpm/package.rb#L439

Issue #1627, opened 2019-04-08
feature request: option to remove auto provides


Jamesits:
Converting "provides" from one type of package to another might cause issues. For example when I try to convert [smartmontools from OpenSuSE](https://mirrors.ustc.edu.cn/opensuse/tumbleweed/repo/oss/x86_64/smartmontools-6.6-5.2.x86_64.rpm) to deb, it will produce a deb package like this:

```
Provides: config(smartmontools), smartmontools, smartmontools(x86-64)
```

which causes dpkg to panic:

```
dpkg: warning: parsing file '/var/lib/dpkg/tmp.ci/control' near line 8 package 'smartmontools':
 'Provides' field, reference to 'config':
 implicit exact match on version number, suggest using '=' instead
dpkg: error processing archive smartmontools_6.6-5.2_amd64.deb (--install):
 parsing file '/var/lib/dpkg/tmp.ci/control' near line 8 package 'smartmontools':
 'Provides' field, reference to 'config': error in version: version number does not start with digit
Errors were encountered while processing:
 smartmontools_6.6-5.2_amd64.deb
```

PR #2053, opened 2024-03-08
Fix for #1627

JordanStopford:
When packaging a deb sometimes the architecture comes through as part of the provides (not sure why) but this causes problems with the deb package as it seems to only ever expect one set of round brackets which is the version indicator
Also fixed another issue where ppc64le packages from RPM wouldn't install properly on DEB based systems as they use ppc64el.
Also fixed another issue where invalid dependencies were being put in the control file (rpmlib dependencies and file path dependencies i.e. RPMs that depend on /bin/sh)

This has got fpm working for RPMs->DEBs on our systems for ppc64le packages. Appreciate that this fix might not be "complete" or comprehensive in any way, shape or form so any feedback is welcome.

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -707,6 +707,11 @@ class FPM::Package::Deb < FPM::Package
     self.dependencies = self.dependencies.collect do |dep|
       fix_dependency(dep)
     end.flatten
+
+    # If an invalid depends field was found i.e. /bin.sh then fix_depends will blank it
+    # Make sure we remove this blank here
+    self.dependencies = self.dependencies.reject { |p| p.empty? }
+
     self.provides = self.provides.collect do |provides|
       fix_provides(provides)
     end.flatten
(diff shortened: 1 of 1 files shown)


jordansissel:
It’s ok 👍  to include multiple changes in a single PR. My request for keeping this one PR is to update the PR description to refer to the things you improve. 

Otherwise, yes, all changes look like good improvements :)

JordanStopford:
@jordansissel  I've amended the PR description above. Thanks for your feedback!

jordansissel:
I tested manually with the example from #1627 (`smartmontools` from OpenSUSE, and I confirmed the error without this patch:

```
Invalid relationship field for debian package: config(smartmontools) = 7.4-2.3 {:level=>:error}
Invalid package configuration: Found invalid Provides field values ("config(smartmontools) = 7.4-2.3"). This is not valid in a Debian package. {:level=>:error}
```

Testing your patch looks good:

```
% bundle exec bin/fpm -s rpm -t deb smartmontools-7.4-2.3.x86_64.rpm
...
Created package {:path=>"smartmontools_7.4-2.3_amd64.deb"}

# Check the `Provides` field in the resulting 'deb' file:
% ar p smartmontools_7.4-2.3_amd64.deb control.tar.gz | tar -zxO ./control | grep Provides:
Provides: smartmontools (= 7.4-2.3)
```

PR #2011, opened 2023-05-19
fix RPM documentation links in docs and code

fs111:
The previous versions where all returning 404. This change points to the new correct locations.

PR #2054, opened 2024-03-09
Update rpm.rst: Fix links to the rpm documentation page

white-gecko:
Fix links to the rpm documentation page

### DIFF
--- a/docs/packages/cli/rpm.rst
+++ b/docs/packages/cli/rpm.rst
@@ -35,7 +35,7 @@
 * ``--rpm-init FILEPATH``
     - Add FILEPATH as an init script
 * ``--[no-]rpm-macro-expansion``
-    - install-time macro expansion in %pre %post %preun %postun scripts (see: https://rpm.org/user_doc/scriptlet_expansion.html)
+    - install-time macro expansion in %pre %post %preun %postun scripts (see: https://rpm-software-management.github.io/rpm/manual/scriptlet_expansion.html)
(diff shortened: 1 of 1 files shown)


docs/packages/cli/rpm.rst
```diff
@@ -35,7 +35,7 @@
 * ``--rpm-init FILEPATH``
     - Add FILEPATH as an init script
 * ``--[no-]rpm-macro-expansion``
-    - install-time macro expansion in %pre %post %preun %postun scripts (see: https://rpm.org/user_doc/scriptlet_expansion.html)
+    - install-time macro expansion in %pre %post %preun %postun scripts (see: https://rpm-software-management.github.io/rpm/manual/scriptlet_expansion.html)
```
jordansissel, line 38: This file is generated from `lib/fpm/package/rpm.rb` which has code to declare flags and help text per flag, so these changes need to be made to that file otherwise they'll be lost the next time the docs are generated.

Example:

https://github.com/jordansissel/fpm/blob/d81a1ad1ac7b47662bcb54e7446f2affb5b3170c/lib/fpm/package/rpm.rb#L157-L159

jordansissel (changes requested):
Thank you for improving the docs! I added a note about how this particular documentation file is generated and where the change needs to be made instead.

jordansissel:
A similar change was merged in #2092 via #2011. Thanks for improving fpm! :)

PR #2057, opened 2024-04-11
fix the link to ruby installation instructions

brad:
One character change!
I was reading the docs and noticed the link for ruby installation instructions was broken. This fixes it


jordansissel:
Nice catch! Thank you for this fix :)

PR #2063, opened 2024-06-26
Add option --deb-systemd-path

skoef:
This option will allow you to override the path within the staging directory systemd service files are moved to in debian packages.

On some systems systemd should not go in the hardcoded /lib/systemd/system but in /usr/lib/systemd/system instead. Many systems have the both symlinked, but not all so assuming /lib/systemd/system can be dangerous.

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -210,6 +210,11 @@ class FPM::Package::Deb < FPM::Package
     next File.expand_path(file)
   end
 
+  option "--systemd-path", "FILEPATH", "Relative path to the systemd service directory",
+    :default => "lib/systemd/system" do |file|
+    next file.gsub(/^\/*/, '')
+  end
+
   option "--systemd-enable", :flag , "Enable service on install or upgrade", :default => false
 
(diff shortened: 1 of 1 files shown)


jordansissel:
Tested, works! Thank you for improving fpm :)

Test:

```
% touch /tmp/example.service ; bundle exec bin/fpm -s empty -t deb --deb-systemd-path fancypants --deb-systemd /tmp/example.service -n example
Adding action files
Created package {:path=>"example_1.0_all.deb"}

% ar p example_1.0_all.deb data.tar.gz | tar -zt | grep example.service
./fancypants/example.service
```

And testing default behavior:

```
% touch /tmp/example.service ; bundle exec bin/fpm -s empty -t deb --deb-systemd /tmp/example.service -n example
Adding action files
Created package {:path=>"example_1.0_all.deb"}

% ar p example_1.0_all.deb data.tar.gz | tar -zt | grep example.service
./lib/systemd/system/example.service
```

Jmennius:
This also helps with installing user services under `/usr/lib/systemd/user`!

PR #2065, opened 2024-08-12
Add support for .timer units to --deb-systemd

### LINKED ISSUE #1978: Installing a systemd timer
I'd like to install a systemd timer via a Debian package. Is there a particular way of expressing this in fpm, similarly to `--deb-systemd`?

If I use `--deb-systemd` for a timer (`foo.timer`), it seems to append `.service` to the filename (-> `foo.timer.service`).

I can install a timer by specifying the timer file as an ordinary file, to be placed in `/usr/lib/systemd/system` and using an `after-install` hook to create the symlink in the `timer.target.wants` directory, but I was wondering if there's a better way.

phillipp:
Before the change, when using an option --deb-systemd myunit.timer, the actual file created would be
/lib/systemd/system/myunit.timer.service

Because of that, systemd would then not find the timer.

With this change, the correct extension is extracted from the file, defaulting to .service.

Typically, there would be two files for installing a timer:

--deb-systemd myunit.service (where .service could be omitted) and --deb-systemd myunit.timer

So that the .timer references a .service file.

Fixes #1978

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -517,14 +517,23 @@ class FPM::Package::Deb < FPM::Package
 
     attributes[:deb_systemd] = []
     attributes.fetch(:deb_systemd_list, []).each do |systemd|
-      name = File.basename(systemd, ".service")
-      dest_systemd = staging_path("lib/systemd/system/#{name}.service")
+      name = File.basename(systemd)
+      extname = File.extname(name)
+
+      name_with_extension = if extname.empty?
+                              "#{name}.service"
+                            elsif [".service", ".timer"].include?(extname)
+                              name
+                            else
+                              raise ArgumentError, "Invalid systemd unit file extension: #{extname}. Expected .service or .timer, or no extension."
+                            end
+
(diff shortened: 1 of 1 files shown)


lib/fpm/package/deb.rb
```diff
@@ -517,14 +517,23 @@ def output(output_path)
...
-      dest_systemd = staging_path("lib/systemd/system/#{name}.service")
+      name = File.basename(systemd)
+      extname = File.extname(name)
+
+      name_with_extension = if extname.empty?
+                              "#{name}.service"
+                            elsif [".service", ".timer"].include?(extname)
+                              name
+                            else
+                              raise ArgumentError, "Invalid systemd unit file extension: #{extname}. Expected .service or .timer, or no extension."
```
jordansissel, line 528: This should raise FPM::InvalidPackageConfiguration so that fpm prints an error message. With ArgumentError, fpm will exit and print a stack trace.

jordansissel (changes requested):

jordansissel:
I'll merge as-is and make the exception change after.

PR #2066, opened 2024-09-06
fix(rpm): replace perl dependency by perl-interpreter

kduret:
`perl` rpm package is a metapackages which install all the Perl bits and core modules
Then, `perl-devel`, `gcc` and lot of useless dependencies are installed
This should be replaced by `perl-interpreter`

<details>
  <summary><i>perl dependencies</i></summary>
  <pre><code>
  # dnf repoquery --requires perl
  Last metadata expiration check: 0:35:46 ago on Fri Sep  6 07:47:45 2024.
  perl(:MODULE_COMPAT_5.32.1)
  perl-Archive-Tar
  perl-Attribute-Handlers
  perl-AutoLoader
  perl-AutoSplit
  perl-B
  perl-Benchmark
  perl-CPAN
  perl-CPAN-Meta
  perl-CPAN-Meta-Requirements
  perl-CPAN-Meta-YAML
  perl-Carp
  perl-Class-Struct
  perl-Compress-Raw-Bzip2
  perl-Compress-Raw-Zlib
  perl-Config-Extensions
  perl-Config-Perl-V
  perl-DBM_Filter
  perl-DB_File
  perl-Data-Dumper
  perl-Devel-PPPort
  perl-Devel-Peek
  perl-Devel-SelfStubber
  perl-Digest
  perl-Digest-MD5
  perl-Digest-SHA
  perl-DirHandle
  perl-Dumpvalue
  perl-DynaLoader
  perl-Encode
  perl-Encode-devel
  perl-English
  perl-Env
  perl-Errno
  perl-Exporter
  perl-ExtUtils-CBuilder
  perl-ExtUtils-Command
  perl-ExtUtils-Constant
  perl-ExtUtils-Embed
  perl-ExtUtils-Install
  perl-ExtUtils-MM-Utils
  perl-ExtUtils-MakeMaker
  perl-ExtUtils-Manifest
  perl-ExtUtils-Miniperl
  perl-ExtUtils-ParseXS
  perl-Fcntl
  perl-File-Basename
  perl-File-Compare
  perl-File-Copy
  perl-File-DosGlob
  perl-File-Fetch
  perl-File-Find
  perl-File-Path
  perl-File-Temp
  perl-File-stat
  perl-FileCache
  perl-FileHandle
  perl-Filter
  perl-Filter-Simple
  perl-FindBin
  perl-GDBM_File
  perl-Getopt-Long
  perl-Getopt-Std
  perl-HTTP-Tiny
  perl-Hash-Util
  perl-Hash-Util-FieldHash
  perl-I18N-Collate
  perl-I18N-LangTags
  perl-I18N-Langinfo
  perl-IO
  perl-IO-Compress
  perl-IO-Socket-IP
  perl-IO-Zlib
  perl-IPC-Cmd
  perl-IPC-Open3
  perl-IPC-SysV
  perl-JSON-PP
  perl-Locale-Maketext
  perl-Locale-Maketext-Simple
  perl-MIME-Base64
  perl-Math-BigInt
  perl-Math-BigInt-FastCalc
  perl-Math-BigRat
  perl-Math-Complex
  perl-Memoize
  perl-Module-CoreList
  perl-Module-CoreList-tools
  perl-Module-Load
  perl-Module-Load-Conditional
  perl-Module-Loaded
  perl-Module-Metadata
  perl-NDBM_File
  perl-NEXT
  perl-Net
  perl-Net-Ping
  perl-ODBM_File
  perl-Opcode
  perl-POSIX
  perl-Params-Check
  perl-PathTools
  perl-Perl-OSType
  perl-PerlIO-via-QuotedPrint
  perl-Pod-Checker
  perl-Pod-Escapes
  perl-Pod-Functions
  perl-Pod-Html
  perl-Pod-Perldoc
  perl-Pod-Simple
  perl-Pod-Usage
  perl-Safe
  perl-Scalar-List-Utils
  perl-Search-Dict
  perl-SelectSaver
  perl-SelfLoader
  perl-Socket
  perl-Storable
  perl-Symbol
  perl-Sys-Hostname
  perl-Sys-Syslog
  perl-Term-ANSIColor
  perl-Term-Cap
  perl-Term-Complete
  perl-Term-ReadLine
  perl-Test
  perl-Test-Harness
  perl-Test-Simple
  perl-Text-Abbrev
  perl-Text-Balanced
  perl-Text-ParseWords
  perl-Text-Tabs+Wrap
  perl-Thread
  perl-Thread-Queue
  perl-Thread-Semaphore
  perl-Tie
  perl-Tie-File
  perl-Tie-Memoize
  perl-Tie-RefHash
  perl-Time
  perl-Time-HiRes
  perl-Time-Local
  perl-Time-Piece
  perl-Unicode-Collate
  perl-Unicode-Normalize
  perl-Unicode-UCD
  perl-User-pwent
  perl-autodie
  perl-autouse
  perl-base
  perl-bignum
  perl-blib
  perl-constant
  perl-debugger
  perl-deprecate
  perl-devel(x86-64) = 4:5.32.1-481.el9
  perl-diagnostics
  perl-doc
  perl-encoding
  perl-encoding-warnings
  perl-experimental
  perl-fields
  perl-filetest
  perl-if
  perl-interpreter(x86-64) = 4:5.32.1-481.el9
  perl-less
  perl-lib
  perl-libnet
  perl-libnetcfg
  perl-libs(x86-64) = 4:5.32.1-481.el9
  perl-locale
  perl-macros
  perl-meta-notation
  perl-mro
  perl-open
  perl-overload
  perl-overloading
  perl-parent
  perl-perlfaq
  perl-ph
  perl-podlators
  perl-sigtrap
  perl-sort
  perl-subs
  perl-threads
  perl-threads-shared
  perl-utils
  perl-vars
  perl-version
  perl-vmsish
 </code></pre>
</details>

### DIFF
--- a/lib/fpm/package/rpm.rb
+++ b/lib/fpm/package/rpm.rb
@@ -277,7 +277,19 @@ class FPM::Package::RPM < FPM::Package
 
   # See FPM::Package#converted_from
   def converted_from(origin)
-    if origin == FPM::Package::Gem
+    if origin == FPM::Package::CPAN
+      fixed_deps = []
+      self.dependencies.collect do |dep|
+        # RPM package "perl" is a metapackage which install all the Perl bits and core modules, then gcc...
+        # this must be replaced by perl-interpreter
+        if name=/^perl([\s<>=].*)$/.match(dep)
+          fixed_deps.push("perl-interpreter#{name[1]}")
+        else
+          fixed_deps.push(dep)
+        end
+      end
+      self.dependencies = fixed_deps
+    elsif origin == FPM::Package::Gem
       fixed_deps = []
       self.dependencies.collect do |dep|
         # Gem dependency operator "~>" is not compatible with rpm. Translate any found.


jordansissel:
Thank you for this improvement!

Short summary:
* I think this is a good change.
* RHEL 7 (and derivatives) and Fedora 26 and older are negatively impacted by this change.
* A flag should be added to revert to the old behavior (depend on `perl`)
* Add a warning log to help folks better find out about this change, [similar to this warning](https://github.com/jordansissel/fpm/blob/9d3d96a93d6fd667ac21fecf2f91004bfaa77142/lib/fpm/package/deb.rb#L813-L814); something indicating "I am setting the package to depend on `perl-interpreter`" 

Longer version:

Before merging, I needed to know more about the history of the 'perl-interpreter' package name to better understand the risk of this change: older redhat (and derived) systems will produce uninstallable packages if their system does not know the `perl-interpreter` package.

In terms of age, `perl-interpreter` has existed in Fedora for about 7 years, and RHEL maybe 5 years. I would love to believe that it is only a small population that would be negatively affected by this change, but I don't think that's realistic, unfortunately.

A flag might be a good way to enable your feature by default while still letting folks use the old behavior.  I'm open to suggestion on a flag name, maybe `--rpm-old-perl-dependency` or something like that?

To find out the history, I checked Fedora because that was the first place I could find a [Red Hat perl package source](https://src.fedoraproject.org/rpms/perl). The first mention I find of `perl-interpreter` is from [a commit in June of 2017](https://src.fedoraproject.org/rpms/perl/c/d3e98ce2042e254d0c1118ba97ccc7ff6bfa7c05?branch=d3e98ce2042e254d0c1118ba97ccc7ff6bfa7c05) which makes the perl package provide this new dependency name. 

Best I can tell, the first Fedora release containing this change is Fedora 27 ([changelog entry](https://fedoraproject.org/wiki/Changes/perl_Package_to_Install_Core_Modules#Current_status)) which was released November of 2017.

I don't know what RHEL 8 uses as a source for its package, but after some digging, I found documentation for [RHEL 8 in a Docker container](https://developers.redhat.com/blog/2020/03/24/red-hat-universal-base-images-for-docker-users). RHEL 8.0 (released in 2019) appears to have `perl-interpreter` known.

I was unable to find a way to get RHEL7 or CentOS 7 operating under my time constraints tonight. I am assuming RHEL 7 users would be negatively impacted by this change. RHEL 7, now 10 years old, only reached EOL a few months ago, with [extended support offered for another 4 years](https://www.redhat.com/en/technologies/linux-platforms/enterprise-linux/rhel-7-end-of-maintenance).

kduret:
Thank you for the detailed answer !

Indeed, this is not a good change if it impacts older OS versions.

`--rpm-old-perl-dependency` is a good idea 👍🏻 

Otherwise, it could be a flag `--rpm-disable-dependency` which could have the same behaviour than `--gem-disable-dependency`.
Then, `--dependency perl-interpreter --rpm-disable-dependency perl` should do the job.
This allows to get a new fpm version without any breaking changes.

NicholasBHubbard:
Just chiming in since I (along with @wbraswell) have interest in the Perl side of FPM.

This PR looks good to me. I agree that it makes sense to use the `perl-interpreter` dependency by default, but also add a flag to use the old behavior for people on older RHELs.

wbraswell:
I only use Debian-based operating systems, so assuming all the stuff about RedHat and CentOS is accurate, then yes I approve of this Perl-related pull request.

jordansissel:
> Then, --dependency perl-interpreter --rpm-disable-dependency perl should do the job.

I thought about this for a bit, tonight, but decided against this method (`--rpm-disable-dependency perl`) because the `perl` dependency carried from cpan-to-rpm conversion will include versioning, so I think it'll be preferred to make this more automatic.

To that end, I will add a `--rpm-old-perl-dependency-name` flag after merging this.

@NicholasBHubbard @wbraswell thank you for the feedback :)

PR #2068, opened 2024-09-23
Support pip download returning a zip file

mattaezell:
It seems that `debugpy` on pip returns a `zip` file instead of a `tar.gz`


jordansissel:
Looks good me. Thank you for improving fpm! :)

(Also thanks  @hussainbani who filed a similar patch in #2074)

Issue #2072, opened 2024-10-10
fpm method fails while generating package for python if downloaded package in zip format


hbani:
Here’s a GitHub issue description for the problem you're encountering with `fpm`:

---

**Title**: `fpm` Fails to Generate Package for Python When Package is Downloaded in `.zip` Format

**Description**:

I'm experiencing an issue when using `fpm` to create a package for a Python project. Specifically, the failure occurs when `pip download` fetches the package in `.zip` format rather than `.tar.gz`. The `fpm` process fails with an unexpected directory layout error, as shown below:

```
/usr/local/rvm/gems/ruby-2.7.1/gems/fpm-1.15.1/lib/fpm/package/python.rb:177:in `download_if_necessary': Unexpected directory layout after `pip download ...`. This might be an fpm bug? The directory is /tmp/package-python-build-xxxxxxxx (RuntimeError)
```

### Steps to Reproduce:

1. Run `fpm` to create a package for a Python module, which `pip` downloads in `.zip` format:
   ```bash
   fpm -s python -t deb <package_name>
   ```
   
2. If `pip` downloads the package as a `.zip` archive, the process fails with a directory layout error.

### Expected Behavior:

`fpm` should handle Python packages regardless of whether they are downloaded as `.zip` or `.tar.gz` and correctly generate the package.

### Actual Behavior:

The process fails with the following error:
```
Unexpected directory layout after `pip download ...`
```

### Workaround:

Manually extracting the `.zip` file and converting it to `.tar.gz` allows the process to succeed. However, this adds an unnecessary manual step to the packaging process.

### Environment:

- `fpm` version: 1.15.1
- Ruby version: 2.7.1
- Python version: (whichever is relevant)
- OS: (provide your OS and version)

---

**Additional Context**:

This issue seems to arise specifically with `.zip` formatted packages. `fpm` works fine when the package is downloaded in `.tar.gz` format.

---

This description should help developers understand and reproduce the issue while providing sufficient context for further investigation.

hussainbani:
PR to fix this issue
https://github.com/jordansissel/fpm/pull/2074

PR #2074, opened 2024-10-10
"Fix: Handle .zip files when generating Python packages"

hussainbani:
### Description:

This Pull Request addresses an issue where `fpm` fails to properly generate a package when the Python module is downloaded in `.zip` format via `pip`. Previously, `fpm` expected only `.tar.gz` archives and would throw an error when encountering a `.zip` file due to an unexpected directory layout.

#### Changes made:
- Added logic to detect if the downloaded Python package is in `.zip` format.
- Implemented functionality to properly extract `.zip` files using the `unzip` command and handle the packaging process as usual.
- Maintained backward compatibility for `.tar.gz` files.
  
#### Testing:
- Verified the fix by packaging Python modules provided in both `.zip` and `.tar.gz` formats.
- Both archive types now process correctly, and the expected packages are generated without errors.

This PR resolves issue https://github.com/jordansissel/fpm/issues/2072, where users encountered a runtime error when `pip` fetched `.zip` files during the `fpm` packaging process.

--- 

Feel free to tweak the wording to better fit the specifics of your implementation!

### DIFF
--- a/lib/fpm/package/python.rb
+++ b/lib/fpm/package/python.rb
@@ -173,11 +173,17 @@ class FPM::Package::Python < FPM::Package
       # If we expect `pip` to leave an unknown-named file in the `build_path` directory, let's check for
       # a single file and unpack it.  I don't know if it will /always/ be a .tar.gz though.
       files = ::Dir.glob(File.join(build_path, "*.tar.gz"))
+      files += ::Dir.glob(File.join(build_path, "*.zip"))
       if files.length != 1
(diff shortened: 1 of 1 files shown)


mattaezell:
This duplicates #2068

jordansissel:
@hussainbani Thank you for your detailed writeup and the patch. I'm going to close this in favor of a similar solution provided by @mattaezell, but will credit both of you in the changelist when released.

PR #1476, opened 2018-03-23
Add executable and python arguments for virtualenv sources

zmotu:

PR #2083, opened 2024-12-04
Add option to specify python version for virtualenv

GibbonJojo:
I sometimes want to package virtualenvs with other python version than my system default. This adds another cli option to specify an alternative python executable. Naming is the same as for virtualenv
https://virtualenv.pypa.io/en/latest/cli_interface.html

I decided not to check for a valid executable in this code as virtualenv will forward the error anyways

I hope the code is fine as is, I'm not really versed in Ruby, but this might serve as a proposal for the requested change.

Tested it locally and works on my linux machine with a caveat though:
1. virtualenv should not be installed in environments, instead it should be installed in the global interpreter or (preferred) via pipx
2. virtualenv-tools can misfunction if you're trying to run it on a virtualenv, that has a different version than the one it's running on
E.g. having Python3.10 as base, running
virtualenv --python python3.12 environment_name
cd environment_name 
virtualenv-tools --update-path /some/path

will throw an error while updating the .pycs

So given the example of having Python3.10 as your base python, but you want to build your virtualenv for Py3.12, you need to install virtualenv with pipx, then create a Py3.12 environment, install virtualenv-tools3 in there and run fpm in this environment. 
It is a bit finicky, but that would need solving in virtualenv-tools, I guess. Atleast that enables fpm to build virtualenvs for different Python versions

### DIFF
--- a/lib/fpm/package/virtualenv.rb
+++ b/lib/fpm/package/virtualenv.rb
@@ -112,9 +112,9 @@ class FPM::Package::Virtualenv < FPM::Package
         virtualenv_args << virtualenv_build_folder
     end
 
-    if self.attributes[:python?]
-      logger.info("Creating virtualenv with python executable #{self.attributes[:python]}")
-      virtualenv_args.concat(["--python", self.attributes[:python]])
+    if self.attributes[:virtualenv_python]
+      logger.info("Creating virtualenv with python executable #{self.attributes[:virtualenv_python]}")
+      virtualenv_args.concat(["--python", self.attributes[:virtualenv_python]])
     end
 
     safesystem(*virtualenv_args)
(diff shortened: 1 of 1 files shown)


mummynobbit:
Is it a duplicate of #1476?

jordansissel:
> Is it a duplicate of https://github.com/jordansissel/fpm/pull/1476?

@mummynobbit agreed.

> virtualenv-tools can misfunction if you're trying to run it on a virtualenv, that has a different version than the one it's running on

@GibbonJojo - I also noticed this. I wonder if it would help to install `virtualenv-tools3` inside the virtualenv and then that setup it to perform the `virtualenv-tools --update-path ...` step using the same version of python?

PR #2086, opened 2024-12-09
feat(ci): Bump checkout to v4; Bump ci test to non-rol Ruby

xlionjuan:
https://www.ruby-lang.org/en/downloads/branches/

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -10,11 +10,11 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


jordansissel:
Hello! Thank you for improving fpm :)

I am in favor of adding additional ruby versions to test against.

However, I am not in favor of removing older versions of ruby. I'd like to keep testing against older versions of ruby. This is because while the Ruby project may have moved on from Ruby 2.x and even Ruby 1.9.x, there's still a wide world of systems using these older and abandoned versions of ruby, and I think it is useful to continue testing against these versions. I don't want to accidentally break fpm on older versions of ruby :)

Issue #54, opened 2011-06-21
Build rpms without requiring rpmbuild


jordansissel:
rpmbuild isn't available on some systems without going through pains to install it manually.

The RPM format is poorly documented, but in general, it should be doable to generate RPMs directly from ruby instead of rpmbuild. This would also let me avoid any workarounds I might require to use rpmbuild.

kevinkjt2000:
This is the one thing keeping me from using fpm for rpm creation.  Adding the pure ruby capability of writing rpm files would be exactly in line with the first bit of the README.

> * If fpm is not helping you make packages easily, then there is a bug in fpm.
> * If you are having a bad time with fpm, then there is a bug in fpm.

Should this issue be labeled with `bug`?

kevinkjt2000:
> Can you describe your situation more? What about rpmbuild prevents you from using it?

Sure thing :smile: 

> rpmbuild isn't available on some systems without going through pains to install it manually.

jarondl:
It does not handle most metadata yet (significantly package requirements), but if all you want is to put files in an rpm, you may consider rpmpack (tar2rpm), written in golang without calling rpmbuild.
https://github.com/google/rpmpack

PR #2088, opened 2024-12-19
rpm: Escape { and } characters in filenames

### LINKED ISSUE #2087: An issue with fpm 1.16 in OL8 not reading the special character in dir name
I encountered an issue with fpm 1.16 where directory names with {{ label }} enclosed in quotes do not work as expected. When the directory name is in quotes, it causes the following file-not-found error:

```
error: File not found: /tmp/package-rpm-build-d14a72e6d0fc6578070ae6940bf08df171015f7714de4e0a9ff4cd7d1b69/BUILD/opt/sfw/foo-79.4/lib/python3.6/site-packages/cement/cli/templates/generate/plugin/{{?label?}}/controllers/__init__.py {:level=>:info, :file=>"cabin/mixins/pipe.rb", :line=>"47", :method=>"block in pipe"}
error: File not found: /tmp/package-rpm-build-d14a72e6d0fc6578070ae6940bf08df171015f7714de4e0a9ff4cd7d1b69/BUILD/opt/sfw/foo-79.4/lib/python3.6/site-packages/cement/cli/templates/generate/plugin/{{?label?}}/controllers/__pycache__/__init__.cpython-36.pyc {:level=>:info, :file=>"cabin/mixins/pipe.rb", :line=>"47", :method=>"block in pipe"}

```
However, if the quotes around the directory name {{ label }} are removed or dir is removed, everything works fine. It seems that fpm 1.16 has trouble processing directory names that contain {{ label }} in quotes.

Please let me know if there is a known workaround or if further investigation is needed.

I am using fpm 1.16

jordansissel:
* Add test coverage for escaping { and } filename characters

Fixes #2087


jordansissel:
Looks like this fails on the Github ubuntu-22.04 runner, but I'm not sure how to debug further as I cannot find the actual image Github uses published anywhere. There's tooling to build all the images, but the instructions appear to require Azure which I do not have resources to use. 

Tests pass locally for me, but I'm on macOS with rpm v4.19.1.1. The Github actions documentation claims that the Ubuntu 22.04 image includes rpm 4.17.0

I don't know anything more about the failure at this time. I won't merge until I understand that problem.

jordansissel:
I'm unable to find a way to use Github Actions images, so I tried to reproduce this outside.

Best I can tell, rpm v4.19 works, but v4.17 does not.

I tested manually with this rpm spec:

```
Name: test
Version: 1.0
Release: 1
Summary: Nope
License: Nope

%description
none

%prep
# noop
%build
# noop
%install
cp -R . %{buildroot}

%clean
# noop

%files
/[\{][\{]?label?[\}][\}].text
```

* ✅  On macOS and rpm v4.19, this builds an rpm successfully. 
* ❌ On Ubuntu 22.04 and rpm v4.17, this fails.
* ❌ On Ubuntu 24.04 and rpm v4.18, this fails.

Looking at the rpm release notes, I found this excerpt:
> Improve shell-like globbing and escaping in %files (#1749)

This refers to https://github.com/rpm-software-management/rpm/issues/1749 which seems pretty relevant.

Times like these remind me that maybe #54 would be a nice improvement.

I did a bit of testing and learned a few small things:
* This PR only works on RPM v4.19 or newer
* When buliding RPMs on RPM v4.18 or older, fpm will fail because rpmbuild errors.
* All GitHub Action images use versions of RPM that are too old, so all the test machines will fail.
* RPMs built on rpm v4.19 with these weird files *will correctly install* on RPM v4.18 and older!

They install fine on rpm v4.18 and older:

```
root@efed807b289b:/rpmbuild/RPMS/aarch64# rpm --version
RPM version 4.18.2
root@efed807b289b:/rpmbuild/RPMS/aarch64# rpm -qp test-1.0-1.aarch64.rpm -l
/{{ label }}.text
root@efed807b289b:/rpmbuild/RPMS/aarch64# rpm -Uvh test-1.0-1.aarch64.rpm
...
root@efed807b289b:/rpmbuild/RPMS/aarch64# ls /\{\{\ label\ \}\}.text
'/{{ label }}.text'
```

jordansissel:
I think I can make this specific test `skip` if rpmbuild is older than rpm v4.19

jordansissel:
Escaping `{` with `[\{]` builds fine with rpm v4.19, but fails on anything older. I'm gonna try replacing `{` with `?`. If this works, maybe fpm can query the version of rpmbuild and use the best escaping method available.

Manish2481983:
@jordansissel If possible, Can we have any tentative timeline for this fix?

jordansissel:
> Can we have any tentative timeline for this fix?

With limited funding, I'm unable to plan specific timelines for any particular changes.

This change seems to work on rpm v4.19 and if I test manually on rpm v4.17, but I have not run the test suite on rpm v4.17 to know if anything else has broken as a result of this change.

jordansissel:
Tests are passing. This is hopefully safe to merge ;)

PR #2092, opened 2025-03-06
fix RPM documentation links in docs and code

jordansissel:
The previous versions where all returning 404. This change points to the new correct locations.

(This is a rebase of #2011)


jordansissel:
Test suite failing on something unrelated.

Issue #2102, opened 2025-09-15
hard linking causes issues for DEB and RPMs when /opt is on a different drive when used with Electron


rathboma:
A few years ago we documented an issue with Electron applications whereby if a user has `/opt` on it's own device, then electron apps would fail to install, as it tries to hard-link some assets.

https://github.com/electron-userland/electron-builder/issues/5721

Looks like this originates in FPM based on the sleuthing of a community member:
https://github.com/electron-userland/electron-builder/issues/5721#issuecomment-3291426610

Seems to be this code in FPM specifically: https://github.com/jordansissel/fpm/blob/main/lib/fpm/package/dir.rb#L199-L208:

```ruby
      # Otherwise try copying the file.
      begin
        logger.debug("Linking", :source => source, :destination => destination)
        File.link(source, destination)
      rescue Errno::ENOENT, Errno::EXDEV, Errno::EPERM
        # Hardlink attempt failed, copy it instead
        logger.debug("Copying", :source => source, :destination => destination)
        copy_entry(source, destination)
      rescue Errno::EEXIST
        sane_path = destination.gsub(staging_path, "")
        logger.error("Cannot copy file, the destination path is probably a directory and I attempted to write a file.", :path => sane_path, :staging => staging_path)
      end
```

A symbolic link would work fine I think? Posting here in case someone else has ideas on whether hard linking is required

t3chguy:
I think the high level issue is that assuming that the installation target machine supports hardlinks just because the build machine does isn't always the case.

jordansissel:
Ooh nice research into this! If I recall, Fpm uses hard links mainly as a way to save space while transforming packages and isn’t necessarily intended to create hard links in the resulting package.

I’ll check into the code history and your linked issues, and with luck, I think it would be safe to change fpm to not hard link.

t3chguy:
Or maybe one could add something akin to `--hard-dereference` when constructing the tarball

jordansissel:
For my first step, I wanted to reproduce the error based on reports in the linked issues:

```
## Mount /opt as a different filesystem, tmpfs should work:
% docker run --mount type=tmpfs,dst=/opt -it ubuntu:latest

# apt-get update; apt-get install -y wget
# wget https://github.com/beekeeper-studio/beekeeper-studio/releases/download/v4.6.8/beekeeper-studio_4.6.8_amd64.deb
# apt-get install -y ./beekeeper-studio_4.6.8_amd64.deb

...
[...]

Issue #365, opened 2013-02-20
fpm and directories containing hard links yields huge packages


pforai:
I've tried this with a recent build of git. The main culprit is that ie git is a multi-call binary and all the individual git-\* commands are hard links pointing to the same file. This causes a regular RPM size for something like git to be roughly 240MB in size and extracted size is something like 650MB since all the hard links are de-references and copied into the RPM. 

Dunno if the issue here lies within FPM or the underlying rpmbuild command.

PR #2103, opened 2025-09-22
Only hardlink if the source is a hardlink.

jordansissel:
In #2102, it was found that fpm would hardlink any file that's included in the package multiple times.

For example:

```
% echo "Hello world" > /tmp/one

# Create a package with /tmp/one installed in two locations using path mapping:
% fpm -s dir -t deb -n example /tmp/one=/opt/example/one /tmp/one=/var/lib/example/one

% dpkg-deb --fsys-tarfile example_1.0_amd64.deb| tar -vt | grep 'one'
-rw-r--r-- 0/0              12 2025-09-21 21:43 ./var/lib/example/one
hrw-r--r-- 0/0               0 2025-09-21 21:43 ./opt/example/one link to ./var/lib/example/one
```

Notice above, /opt/example/one is a hardlink to /var/lib/example/one.

This behavior is not the intended behavior of fpm. The  intended behavior (introduced in #365) was to ensure hardlinks in the original files are correctly kept as hardlinks in the package. The example in #365 was for git-core which has over 100 files hardlinked to the `git` executable  such as `git-add`, `git-apply`, etc. The problem in #365 was that fpm was copying these as individual files which resulted in hundreds of copies of the same file, and in git's case, the `git` binary is 18 megs.

Now, in this PR, the intent is to retain the solution to #365 (#623) while removing the unintended behavior.

```
# Copy /etc/zshrc into a two locations.
% fpm -s dir -t deb -n example /etc/zshrc=/zshrc1 /etc/zshrc=/zshrc2

% dpkg-deb --fsys-tarfile example_1.0_amd64.deb | tar -vt | grep /zshrc
-rw-r--r-- 0/0            1164 2025-01-18 16:00 ./zshrc2
hrw-r--r-- 0/0               0 2025-01-18 16:00 ./zshrc1 link to ./zshrc2
```

The above shows the buggy behavior.

With this PR, the package created looks like this:

```
% dpkg-deb --fsys-tarfile example_1.0_amd64.deb | tar -vt | grep /zshrc
-rw-r--r-- 0/0            1164 2025-01-18 16:00 ./zshrc2
-rw-r--r-- 0/0            1164 2025-01-18 16:00 ./zshrc1
```

Correctly two files, not hardlinked.

### DIFF
--- a/lib/fpm/command.rb
+++ b/lib/fpm/command.rb
@@ -3,7 +3,6 @@ require "fpm/namespace"
 require "fpm/version"
 require "fpm/util"
 require "clamp"
-require "ostruct"
 require "fpm"
 require "tmpdir" # for Dir.tmpdir
 
--- a/lib/fpm/package.rb
+++ b/lib/fpm/package.rb
@@ -3,7 +3,6 @@ require "fpm/util" # local
 require "pathname" # stdlib
 require "find"
 require "tmpdir" # stdlib
-require "ostruct"
 require "backports/latest"
 require "socket" # stdlib, for Socket.gethostname
 require "shellwords" # stdlib, for Shellwords.escape
--- a/lib/fpm/package/dir.rb
+++ b/lib/fpm/package/dir.rb
@@ -197,10 +197,6 @@ class FPM::Package::Dir < FPM::Package
     else
       # Otherwise try copying the file.
       begin
-        logger.debug("Linking", :source => source, :destination => destination)
-        File.link(source, destination)
-      rescue Errno::ENOENT, Errno::EXDEV, Errno::EPERM
-        # Hardlink attempt failed, copy it instead
         logger.debug("Copying", :source => source, :destination => destination)
         copy_entry(source, destination)
       rescue Errno::EEXIST
--- a/lib/fpm/util.rb
+++ b/lib/fpm/util.rb
@@ -332,7 +332,16 @@ module FPM::Util
 
 
   def copy_entry(src, dst, preserve=false, remove_destination=false)
-    case File.ftype(src)
+    st = File.lstat(src)
+
+    filetype = if st.ftype == "file" && st.nlink > 1
+      "hardlink"
+    else
+      st.ftype
+    end
+
+      
+    case filetype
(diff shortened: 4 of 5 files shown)


jordansissel:
At time of writing, this PR still has a small bug that if you twice include a file that's hardlinked (`stat` reports nlink > 1), fpm will hardlink it. This is a bug, but I'm not sure how critical it is.

Example:

```
% ls -li /tmp/a
total 8
233022 -rw-r--r--. 2 jls jls 6 Sep 21 21:57 bar
233022 -rw-r--r--. 2 jls jls 6 Sep 21 21:57 foo

% fpm -s dir -t deb -n example /tmp/a/foo=/foo /tmp/a/foo=/foo2

% dpkg-deb --fsys-tarfile example_1.0_amd64.deb | tar -vt  | grep foo
-rw-r--r-- 0/0               6 2025-09-21 21:57 ./foo2
hrw-r--r-- 0/0               0 2025-09-21 21:57 ./foo link to ./foo2
```

This PR restricts hardlinking to files known to be hardlinks, that is, if `nlink` > 1 on the file stat.

In this example above, 
* `/tmp/a/foo` and `/tmp/a/bar` are hardlinks to the same inode
* Only `/tmp/a/foo` is included in the package
* `/tmp/a/foo` is included twice
* because `nlink`>1 on this file, fpm will hardlink the second file to the first (foo2 to foo).

Solving this specific issue could be out of scope for #2102, so I think it's ok to leave unsolved.

jordansissel:
The test suite is repeatedly failing on Github Actions due to `apt-get install` problems. The suite passes for me locally (with exception for two unrelated failures in perl/python tests)

lib/fpm/command.rb
```diff
@@ -3,7 +3,6 @@
 require "fpm/version"
 require "fpm/util"
 require "clamp"
-require "ostruct"
```
jordansissel, line 6: I added this to get rid of a Ruby warning about a future change: "warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0."

PR #2105, opened 2025-10-02
Allow packaging local python package directories and also fix broken --python-obey-requirements-txt tests

jordansissel:


jordansissel:
The tests are failing on github actions but I can't reproduce locally on either my Mac nor on Fedora 42.

I'm not sure exactly what's going on, and I don't see a way to debug things on GHA without a complex and long feedback loop to try things out. I'm ignoring the test failure.

Here's the failure output:
```
  1) FPM::Package::Python when given a project containing a pyproject.toml should package it correctly
     Failure/Error: insist { subject.name } == "#{prefix}-example"

     Insist::Failure:
       Expected "python-example", but got "python-unknown"
     # ./vendor/bundle/ruby/3.1.0/gems/insist-1.0.0/lib/insist/assert.rb:8:in `assert'
     # ./vendor/bundle/ruby/3.1.0/gems/insist-1.0.0/lib/insist/comparators.rb:13:in `=='
     # ./spec/fpm/package/python_spec.rb:234:in `block (3 levels) in <top (required)>'

  2) FPM::Package::Python when given a project containing a pyproject.toml should package it correctly even if the path given is directly to the pyproject.toml
     Failure/Error: insist { subject.name } == "#{prefix}-example"

     Insist::Failure:
       Expected "python-example", but got "python-unknown"
     # ./vendor/bundle/ruby/3.1.0/gems/insist-1.0.0/lib/insist/assert.rb:8:in `assert'
     # ./vendor/bundle/ruby/3.1.0/gems/insist-1.0.0/lib/insist/comparators.rb:13:in `=='
     # ./spec/fpm/package/python_spec.rb:243:in `block (3 levels) in <top (required)>'
```

PR #2106, opened 2025-10-02
Replace ostruct with a custom class that acts almost the same.

jordansissel:
This removes a warning in 3.4.0 and prepares for Ruby 3.5.0:
> warning: ostruct was loaded from the standard library, but will no longer
> be part of the default gems since Ruby 3.5.0. Add ostruct to your Gemfile or gemspec.

This seems preferable to adding ostruct as a gem dependency especially since older rubies may not support the gem version of ostruct.

### DIFF
--- a/lib/fpm/rake_task.rb
+++ b/lib/fpm/rake_task.rb
@@ -1,13 +1,39 @@
 require "fpm/namespace"
-require "ostruct"
 require "rake"
 require "rake/tasklib"
 
 class FPM::RakeTask < Rake::TaskLib
+  class Options
+    attr_accessor :args
+
+    def initialize(defaults=nil)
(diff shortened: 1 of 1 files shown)


jordansissel:
Honestly, not sure anyone uses FPM::RakeTask anymore. I was thinking about removing it, but the code to replace ostruct was fast enough to write that I'm not worried about deleting it right now.

```
% bundle exec rspec -fd -e Rake
Run options: include {:full_description=>/Rake/}

FPM::RakeTask
  #new
    requires a package name
    requires a source
    requires a target
    requires package args
    executes FPM::Command with the appropriate arguments

Finished in 0.00741 seconds (files took 0.29507 seconds to load)
5 examples, 0 failures
```

Issue #2110, opened 2025-10-02
Default python package prefix - should it be based on the python version?


jordansissel:
Abstract: The default python package name prefix might not be the best one.

Proposal: The default python package name prefix should be "pythonN" where N is the python major version.

As of fpm 1.17.0, the python package prefix (`--python-package-name-prefix`) has a new behavior. If no `--python-bin` is given, fpm will try to find a python executable - python, python3, or python2. Depending on which one is found, it will update the default name prefix to match. For example, if there's no "python" but there is a "python3", the package name prefix default changes to "python3".

Would it be better to use the python version major as the name prefix? Would it be better to have the following:

* if the python-bin is "python" and "python --version" reports `3.13.7`
* then the package name prefix default should become "python3"

Data point: Fedora appears to name almost all of it's python packages using the python version major as a prefix. For example, on my Fedora 42 workstation, the following installed packages matching `/^python` counted by the prefix:

```
% rpm -qa | grep '^python' | cut -d - -f 1 |sort | uniq -c
      5 python
    112 python3
```

Looking at everything `dnf` will find, I see:

```
% dnf search 'python*' | grep '^ *python' | sed -e 's/^ *//' | cut -d - -f 1 | sed -re 's/\s+.*$//' | sort | uniq -c | sort -n
4 python3.13
     12 python3.10
     12 python3.11
     12 python3.12
     16 python3.14
    448 python
   4447 python3
```

Summarizing:
* Most python packages in Fedora are prefixed "python3"
* Of the 448 prefixed "python", 339 of them are "-doc" packages.

Conclusion: Probably should set the default python package prefix to match the python version major

PR #2111, opened 2025-10-02
Change default python package prefix to match the major version of python.

jordansissel:
For #2110


.github/workflows/ruby.yml
```diff
@@ -9,6 +9,7 @@ jobs:
   test:
     runs-on: ubuntu-24.04
     strategy:
+      fail-fast: false
```
sorensonsamuel22-droid, line 12: True

PR #2116, opened 2025-10-30
Filter newer style bash functions out of shell script template's .ins…

ehealy:
…tall-metadata file.

Exported bash functions take the format of `BASH_FUNC_function_name%%`. Including them in the .install-metadata file causes export errors, so update the filtering logic to account for them.

### DIFF
--- a/templates/sh.erb
+++ b/templates/sh.erb
@@ -137,7 +137,7 @@ function save_environment(){
 
     # just piping env to a file doesn't quote the variables. This does
     # filter out multiline junk, _, and functions. _ is a readonly variable.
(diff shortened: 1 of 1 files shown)


jordansissel:
Code looks ok! I haven't tested, but researching what bash does to export functions shows that this PR looks right.

I tried to find a way to have bash simply unexport all functions, but getting a list of just the exported function names seems difficult to do more easily than you have shown, and as it fits with the existing grep pipeline, this seems ok:)

For posterity, best I could come up with was this one, to unexport any exported functions:

`typeset -f +x $(typeset -xfp | sed -ne 's/^declare -fx //p')`

Issue #2124, opened 2025-11-14
Implementation for finding CPAN provided modules is incorrect


NicholasBHubbard:
Currently fpm's functionality for determining a Perl packages provided modules does not work correctly. Currently `cpan.rb` attempts to find provided modules by searching a modules metadata, but CPAN modules do not provide other modules, instead CPAN distributions provide modules. Please note that a CPAN module cooresponds to a Perl package, however unlike Perl packages every CPAN module must have a version.

Here is an example url for a MetaCPAN api query that fpm makes to find the provided modules when installing `Dist::Zilla`: https://fastapi.metacpan.org/v1/module/Dist::Zilla . It is important to note that there is no field at all for provided modules. If we instead make a similar MetaCPAN api query for the `Dist-Zilla` distribution, we can see the many provided modules under the `provides` field: https://fastapi.metacpan.org/v1/release/Dist-Zilla .

All of this is to say that currently, when using cpan as the source package type, we will never get anything other than a single item `provides` field for the package itself.

```
$ bin/fpm --debug-workspace --no-cpan-test -s cpan -t rpm Dist::Zilla
Created package {path: "perl-Dist-Zilla-6.036-1.noarch.rpm"}
plugin directory {plugin: "cpan", pathtype: :staging_path, path: "/tmp/package-cpan-staging-efacb68469ee28953900696a70716fd8899fd785a37e216ac394471d443a"}
plugin directory {plugin: "cpan", pathtype: :build_path, path: "/tmp/package-cpan-build-802c76d174edd1ba7b7ae057b4d3244da3f5b5cb88c679f83d3c25ab2580"}
plugin directory {plugin: "rpm", pathtype: :staging_path, path: "/tmp/package-cpan-staging-efacb68469ee28953900696a70716fd8899fd785a37e216ac394471d443a"}
plugin directory {plugin: "rpm", pathtype: :build_path, path: "/tmp/package-rpm-build-0860f679198261f4b06c578b29b140f9d936b12643f35b0296a8be62bdcb"}
$ grep 'Provides:' /tmp/package-rpm-build-0860f679198261f4b06c578b29b140f9d936b12643f35b0296a8be62bdcb/SPECS/perl-Dist-Zilla.spec
Provides: perl(Dist::Zilla) = 6.036
```

[...]

PR #2125, opened 2025-11-14
Properly find CPAN provided modules

NicholasBHubbard:
This PR aims to solve issue #2124.

Here is a link to the metacpan api url that is currently used, searching a module `Dist::Zilla` (whos distribution provides many modules): https://fastapi.metacpan.org/v1/module/Dist::Zilla . Notice that there is no field for provided modules. If instead we search for a release of the `Dist-Zilla` distribution, we see a long list of provided modules under the `provides` field: https://fastapi.metacpan.org/v1/release/Dist-Zilla .

I created a function `search_provided_modules()` that finds all the provided modules of some version of some distribution. This function is instead used to set the `self.provides` field in the `input()` function. The `release` api does not show the provided modules versions, so I used different but equivalent logic to search the `module` api for all modules that belong to the given distribution version. This allows us to also determine the module version (which isn't necessarily) the same as the distribution version.

While I was at it I updated the `search()` function, renaming it `search_module()`, and added an optional parameter to specify which version of the module to search. The HTTP GET method for the metacpan api url that was being used only finds the latest version of the module. To find a specific version, I updated the code to use the HTTP POST method for the metacpan api that allows for more powerful queries. The query I used is the same except that if the `version` parameter is defined it queries for that specific version of the module instead of the latest version.

Another thing worth noting is that I found that we had an unused variable named `dep` that was defined in the `input()` function, and was created by a call to `search()`. I deleted this variable. [Here](https://github.com/jordansissel/fpm/blob/78d08ef94f43221be47d7fa149245d7da4c413b2/lib/fpm/package/cpan.rb#L177) is a link to the variable definition.

I added new tests to `cpan_spec.rb` for this new `search_provided_modules()` function, and also updated an exisiting test to check `self.provides` after calling `input()`.

Issue #2124 showed that currently fpm did not find the correct provided modules. Here are the same commands that were given in the issue, showing that this PR fixed the problem (cut off as there are over 140 entries):

```
$ bin/fpm --debug-workspace --no-cpan-test -s cpan -t rpm Dist::Zilla
$ grep 'Provides:' /tmp/package-rpm-build-8625b072deccf5d9b8aa3f85dadafd6df946c124a6f05347ad7f2d541059/SPECS/perl-Dist-Zilla.spec
Provides: perl(Dist::Zilla) = 6.036
Provides: perl(Dist::Zilla::App) = 6.036
Provides: perl(Dist::Zilla::App::Command) = 6.036
Provides: perl(Dist::Zilla::App::Command::add) = 6.036
Provides: perl(Dist::Zilla::App::Command::authordeps) = 6.036
Provides: perl(Dist::Zilla::App::Command::build) = 6.036
Provides: perl(Dist::Zilla::App::Command::clean) = 6.036
Provides: perl(Dist::Zilla::App::Command::install) = 6.036
Provides: perl(Dist::Zilla::App::Command::listdeps) = 6.036
Provides: perl(Dist::Zilla::App::Command::new) = 6.036
Provides: perl(Dist::Zilla::App::Command::nop) = 6.036
...
```

### DIFF
--- a/lib/fpm/package/cpan.rb
+++ b/lib/fpm/package/cpan.rb
@@ -428,14 +428,16 @@ EOS
                    :url => metacpan_api_url)
       raise FPM::InvalidPackageConfiguration, "metacpan release query failed"
     end
-    query_hits = JSON.parse(response.body)['hits']['hits'][0]["_source"]["module"]
+    query_hits = JSON.parse(response.body)['hits']['hits']
 
     provided_modules = []
-    query_hits.each do |m|
-      module_name = m["name"]
-      module_version = m["version"]
-      Array(module_name).zip(Array(module_version)).each do |name, version|
-        provided_modules << cap_name(name) + (version ? " = #{version}" : "")
+    query_hits.each do |h|
+      h["_source"]["module"].each do |m|
+        module_name = m["name"]
+        module_version = m["version"]
+        Array(module_name).zip(Array(module_version)).each do |name, version|
+          provided_modules << cap_name(name) + (version ? " = #{version}" : "")
+        end
       end
     end
     return provided_modules


wbraswell:
@jordansissel 
As usual, I have reviewed and approved this PR before it was submitted by @NicholasBHubbard .
Thanks in advance!  :-)

wbraswell:
@jordansissel 
This is currently our second-most-important PR to get merged as soon as possible, because it is blocking us from fixing other downstream bugs.  Thanks!

wbraswell:
Hello @jordansissel , we are also ready for this PR to be merged immediately today, can you please do it now as well?  Thanks in advance!

jordansissel:
Multiple cpan tests fail for me, though, after some digging, I don't believe the failures are related to this issue.  I'll test by hand before merging.

wbraswell:
@jordansissel 
You may recall that we created this PR before we fixed the CI tests in a different PR, so that is (at least one reason) why we see "Some checks were not successful" here in the GitHub CI.

jordansissel:
Oh, I was ignoring GitHub actions. I tested locally on this pr, and also separately in ‘main’

jordansissel:
I believe my local failures are due to a recent change in metacpan's Elasticsearch. Tracking that separately in #2143

wbraswell:
Okay so the question becomes whether or not this current PR can be safely merged independently of the new MetaCPAN Elasticsearch API issue #2143?

jordansissel:
I pushed a few changes to update this PR to use the new Elasticsearch changes affecting the metacpan API.

Locally, I'm able to successfully package Dist::Zilla and confirm it has a large list of provides:

```
% rpm -qp perl-Dist-Zilla-6.037-1.noarch.rpm --provides | head -10
perl(Dist::Zilla) = 6.037
perl(Dist::Zilla::App) = 6.037
perl(Dist::Zilla::App::Command) = 6.037
perl(Dist::Zilla::App::Command::add) = 6.037
perl(Dist::Zilla::App::Command::authordeps) = 6.037
perl(Dist::Zilla::App::Command::build) = 6.037
perl(Dist::Zilla::App::Command::clean) = 6.037
perl(Dist::Zilla::App::Command::install) = 6.037
perl(Dist::Zilla::App::Command::listdeps) = 6.037
perl(Dist::Zilla::App::Command::new) = 6.037
```

NicholasBHubbard:
Thank you very much @jordansissel for your help with this!

I ran the code with your updates and get the same results. Everything looks good to me.

Issue #2127, opened 2025-11-23
deb: `ZSTD_CLEVEL` is set to negative values when using `--deb-compression=zst`


nickhutchinson:
fpm 1.17.0 appears to be setting the `ZSTD_CLEVEL` environment variable to negative values whenever `--output-type=deb --deb-compression=zst` is specified.

The [ZSTD_CLEVEL](https://manpages.debian.org/unstable/zstd/zstd.1.en.html#Environment_Variables) environment variable is documented as being an integer between 1 and 19 inclusive. Despite the docs, zstd doesn't appear to raise an error if negative values are used here; negative values seem to enable its 'fast' compression levels, and using `ZSTD_CLEVEL=0` sets the compression to the default level.

I suspect the issue lies in `deb.rb`; note the leading `-` for the `ZSTD_CLEVEL` value. This has the perverse effect of making, say, `--deb-compression-level=9` generate a larger Debian package than `--deb-compression-level=3`.

https://github.com/jordansissel/fpm/blob/v1.17.0/lib/fpm/package/deb.rb#L698
```
      when "zst"
        datatar = build_path("data.tar.zst")
        controltar = build_path("control.tar.zst")
        compression_flags = ["--use-compress-program", "zstd"]
        compressor_options = {"ZSTD_CLEVEL" => "-#{self.attributes[:deb_compression_level] || 3}"}
```

Further, `--deb-compression-level` currently restricts the permitted values to 0 to 9 (inclusive), even when using `--deb-compression=zst`. Passing `--deb-compression-level=19` to fpm ought to succeed, but will instead raise an error: `Invalid compression level '19'. Valid values are integers between 0 and 9 inclusive.`

Tested on Ubuntu 22.04.

```
$ zstd --version
 *** zstd command line interface 64-bits v1.4.8, by Yann Collet ***

$ fpm --version
1.17.0

$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.5 LTS"
```

NicholasBHubbard:
Nice find. I was able to reproduce that higher compression levels are producing larger packages.

```
$ git log -1 --oneline
d9be498 (HEAD -> main, origin/main, origin/HEAD) Merge pull request #2116 from Tapjoy/ehealy/sh-filter-newer-shell-fuctions
[...]

PR #2129, opened 2025-11-25
fix: zstd compression level handling

domwst:
(Hopefully) fixes https://github.com/jordansissel/fpm/issues/2127

### DIFF
--- a/lib/fpm/package/deb.rb
+++ b/lib/fpm/package/deb.rb
@@ -719,7 +719,7 @@ class FPM::Package::Deb < FPM::Package
         datatar = build_path("data.tar.zst")
         controltar = build_path("control.tar.zst")
         compression_flags = ["--use-compress-program", "zstd"]
-        compressor_options = {"ZSTD_CLEVEL" => "-#{self.attributes[:deb_compression_level] || 3}"}
+        compressor_options = {"ZSTD_CLEVEL" => "#{self.attributes[:deb_compression_level] || ENV['ZSTD_CLEVEL'] || 3}"}
       when "none"
         datatar = build_path("data.tar")
         controltar = build_path("control.tar")


NicholasBHubbard:
Thanks for the PR! In [this comment](https://github.com/jordansissel/fpm/issues/2127#issuecomment-3586349492) on issue #2127 I reproduced the problem of higher compression levels leading to larger packages. This PR does fix that problem:
```
$ git log -1 --oneline
8150c59 (HEAD -> domwst/main) Fix zstd compression level handling
$ bundle exec bin/fpm -s python -t deb --deb-compression=zst --deb-compression-level=3 django
Created package {path: "python-django_5.2.8_all.deb"}
$ stat -c %s python-django_5.2.8_all.deb
6991954
$ rm python-django_5.2.8_all.deb
$ bundle exec bin/fpm -s python -t deb --deb-compression=zst --deb-compression-level=9 django
Created package {path: "python-django_5.2.8_all.deb"}
$ stat -c %s python-django_5.2.8_all.deb
5992540
```

Issue #2127 mentions that values 0-19 are valid zstd compression levels, but fpm only allows values 0-9 for the `--deb-compression` flag. This problem is not addressed in this PR.

domwst:
Yeah, it's kind of fixed by the environment variable `ZSTD_CLEVEL` which allows to pass an arbitrary compression level. I can try to find out how to broaden `--deb-compression` range for zstd compression, but it will take some time.

NicholasBHubbard:
Ah yes, I see that the `ZSTD_CLEVEL` env var could be used to pass an arbitrary compression level. 

Something like the following (UNTESTED) update could be used to broaden the `--deb-compression` flag:

```
  option "--compression-level", "[0-9]", "Select a compression level. 0 is none or minimal. 9 is max compression.",
    # Specify which compression level to use on the compressor backend, when building a package
    :default => nil do |value|
    valint = value.to_i
    max_compression = 9
    if self.attributes[:deb_compression] == "zst"
      max_compression = 19
    end
    unless value =~ /^\d$/ && valint >= 0 && valint <= max_compression
      raise "Invalid compression level '#{value}'. Valid values are integers between 0 and #{max_compression} inclusive."
    end
    valint
  end
```
Do any of the other available compression types allow levels outside the range 0-9?

Issue #2136, opened 2026-02-05
unable to expand .apk's


mcandre:
Unlike .apk's generated directly by Alpine Linux abuild, those generated by fpm struggle to expand on macOS.

The outer gzip archive is able to expand. However, the inner tar archive is not able to expand.

```console
% file hello
hello: POSIX tar archive, file .PKGINFO, mode 0000644, uid 0000000, gid 0000000, size 00000000335, seconds 15141161147
% tar xvf hello
tar: Error opening archive: Unrecognized archive format
```

Relates to https://github.com/jordansissel/fpm/issues/2131

Issue #2131, opened 2026-01-30
apk v2 package format error


hertznsk:
**Problem**

APK packages created using FPM version 1.17.0 cannot be installed on Alpine Linux 3.23 with the error below:
`ERROR: datadog-php-tracer_1.17.0_aarch64.apk: v2 package format error`.

**Root Cause**

Alpine 3.23 introduced apk-tools v3.0.0, which enforces stricter APK v2 package validation. The new version requires a datahash field in the package's PKGINFO file - a SHA256 hash of the gzipped data.tar that's validated post-installation. Packages built with fpm 1.17.0 don't include this field, causing validation failures.

**Additional information**

https://github.com/DataDog/dd-trace-php/pull/3555

mcandre:
Alpine Linux changelogs:

https://www.alpinelinux.org/posts/Alpine-3.23.0-released.html

apk-tools changelogs:

https://gitlab.alpinelinux.org/alpine/apk-tools/-/releases

https://gitlab.alpinelinux.org/alpine/apk-tools/-/releases/v3.0.0

> Support for the original "alpine v1" packages and for broken v2 packages from foreign tools has been removed

Scrolling the history, one of the breaking changes may be a new, strict requirement for more (SHA256) checksums for all files:

https://gitlab.alpinelinux.org/alpine/apk-tools/-/commit/d26b65d30f55a9f0df7d4a94165a657b33b160f1

There may be stricter validation of ISA's.

https://gitlab.alpinelinux.org/alpine/apk-tools/-/commit/c1a3e69f24e235fc78f4dddb1f85fdf281464446

Unknown which ISA identifier string fpm uses by default, on Intel or Apple Silicon Macs.

mcandre:
Requesting assistance from apk-tools, as the validation error message is unfortunately quite generic.

https://gitlab.alpinelinux.org/alpine/apk-tools/-/issues/11187

https://gitlab.alpinelinux.org/alpine/apk-tools/-/issues/11188

mcandre:
Recommend targeting apk v3 format.

PR #2137, opened 2026-02-05
work in progress: apk 3

mcandre:
Followed @hertznsk's recommendation to introduce `datahash` to the `.PKGINFO` file.

Unfortunately, this change alone is not enough. apk continues to reject the resulting packages. And macOS tar is unable to expand the tar archive.

Relates to https://github.com/jordansissel/fpm/issues/2131

Relates to https://github.com/jordansissel/fpm/issues/2136

### DIFF
--- a/lib/fpm/package/apk.rb
+++ b/lib/fpm/package/apk.rb
@@ -91,9 +91,14 @@ class FPM::Package::APK< FPM::Package
     # data tar.
     tar_path(staging_path(""), datatar_path)
 
+    begin
+      # calculate/rewrite sha1 hashes for data tar
+      hash_datatar_sha1(datatar_path)
+    end
+
     # control tar.
     begin
-      write_pkginfo(control_path)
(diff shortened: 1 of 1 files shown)


dgtlrift:
Is this just adding the hash to the v2 format os is it intended to eventually support v3 schema format?

dgtlrift:
I've dug into the PR and created a PR to the PR https://github.com/mcandre/fpm/pull/1 that I've tested and resolves the hashing issue with the v2 strictness. If @mcandre can review and approve it will be populated into the PR.

jordansissel:
@dgtlrift I'm OK if your work gets split into a separate PR, fwiw, whatever is easiest for folks to move this forward.  I don't have any systems that use apk, so I'm happy to rely on y'all to make it work and I'll merge it in :)

dgtlrift:
Created new PR #2153 for direct merge here bypassing @mcandre fork

PR #2146, opened 2026-03-30
Add --cpan-metacpan-api-url flag for custom MetaCPAN instances

NicholasBHubbard:
This PR adds a new flag `--cpan-metacpan-api-url`, which can be used to point fpm to a custom MetaCPAN instance. Currently we already have a flag `--cpan-mirror` for pointing fpm to a custom CPAN mirror instead of the official CPAN. However, in the case that somebody both has a custom CPAN mirror and they are serving a custom MetaCPAN instance on top of that mirror, they need a way to point fpm at both. Without this flag, fpm will query the official MetaCPAN to resolve module metadata, but then attempt to fetch the tarball from the custom CPAN mirror. If the mirror contains modules or versions that aren't indexed by the official MetaCPAN, those lookups will fail.

### DIFF
--- a/lib/fpm/package/cpan.rb
+++ b/lib/fpm/package/cpan.rb
@@ -44,6 +44,10 @@ class FPM::Package::CPAN < FPM::Package
   option "--cpanm-force", :flag,
     "Pass the --force parameter to cpanm", :default => false
 
+  option "--metacpan-api-url", "URL",
+    "URL for the MetaCPAN api",
+    :default => "https://fastapi.metacpan.org"
+
   private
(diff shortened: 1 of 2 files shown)


wbraswell:
@jordansissel 
Okay I know I said that we didn't think we had any more critical PRs but I spoke too soon, sorry!
However this is a relatively simple PR and I have already reviewed it as always, so please do approve it ASAP.
Thanks in advance!

wbraswell:
@jordansissel 
Thanks for merging!  :-)

PR #2147, opened 2026-03-30
Fix for Build ID conflicts - disabling build id generation

rathboma:
Full disclosure: This fix is generated by Claude. I did review it and did a bit of research on the purpose of Build IDs. Seems like they're mostly used for debugging, and it doesn't really hurt to remove them in FPM's case?

This adds %define _build_id_links none to the spec template, consistent with how FPM already disables other rpmbuild post-processing defaults.

Users who need build-id links can override with:
  --rpm-rpmbuild-define '_build_id_links compat'

Fixes #2142

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Release Notes and Change Log
 ============================
 
+Unreleased
+^^^^^^^^^^
+
+* rpm: Disable build-id links by default to prevent file conflicts when multiple packages include pre-built ELF binaries with the same build-ids, such as Electron apps. (`#2142`_; Matthew Rathbone)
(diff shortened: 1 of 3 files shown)


jordansissel:
Thanks for working on this!

Regarding Claude, the copyright aspects are unclear to me (and to current case law in the US?). Given that, I'm not comfortable accepting this PR.

That said, I am in favor of eliminating the build id files in rpm outputs. I'll take a look at implementing that soon; and fortunately, removing the build id is a pretty simple change <3

jordansissel:
I’ll take a look at this again shortly. It’s small enough that I can probably evaluate it and move forward :)

PR #2149, opened 2026-04-10
[fix] Do not add symbol table in ar utility

### LINKED ISSUE #2148: Xcode 26.4 ar utility handles Mach-O files only
## Description
Seems like Xcode 26.4 has changed ar utility behavior.
```
/usr/bin/ar -qc hello_26.4.bin hello.txt
ranlib: warning: archive member 'hello.txt' not a mach-o file
```

It now processes Mach-O files only which results in fpm not being able to produce correct archives. It's current behavior is only to print warning and exit with zero code thus fpm can't handle such error here
https://github.com/jordansissel/fpm/blob/107c892b9b03c1e6848d4c673ec94e5cce181d42/lib/fpm/util.rb#L244-L245

New file also contains `#1/20` entry with `__SYMDEFS` contents which can be a bug in ar command

[hello_26.3.bin.tgz](https://github.com/user-attachments/files/26611244/hello_26.3.bin.tgz)
[hello_26.4.bin.tgz](https://github.com/user-attachments/files/26611245/hello_26.4.bin.tgz)

## Expecting behavior

ar to pack all files whenever they are mach-o or not like previous behaviour from Xcode 26.3

## 

Not sure yet if this check an Apple's bug or not. If it's intended maybe fpm should implement packaging with Ruby libraries instead of calling OS ulitity?

danpashin:
## Description
Xcode 26.4 and later enabled symbol table generation by default. This behavior causes ar to ignore all non-macho files leaving only SYMDEF entry and results in corrupted archive.

## Compatibility
-S flag is already included in all ar impls: bsd (FreeBSD 13.5), llvm-ar (FreeBSD 15.0), gnu (ArchLinux, Ubuntu 16.04), Apple (MacOS 10.10) so this change is fully backward-compatible will all these tools and should not cause any undesirable behavior.

Closes #2148

### DIFF
--- a/lib/fpm/util.rb
+++ b/lib/fpm/util.rb
@@ -250,7 +250,7 @@ module FPM::Util
       end
     end
     # If no combination of ar and options omits timestamps, fall back to default.
-    @@ar_cmd = ["ar", "-qc"]
+    @@ar_cmd = ["ar", "-qcS"]
     FileUtils.rm_f([testarchive, emptyfile])
     return @@ar_cmd
   end # def ar_cmd


mdkent:
Hit #2148 today and this fixes it. Big +1 to merge.

ntamas:
Same here; looking forward to seeing this PR being merged.

PR #2153, opened 2026-07-13
Strict Apk 2 hash compliance

dgtlrift:
Stale tar header checksums (cut_tar_record, and the main header in hash_datatar_sha1): ownership fields (uid/gid/uname/gname/magic) were rewritten but the tar header checksum was never recalculated afterward, corrupting the checksum apk validates on every header.
SHA1 per-file hash included tar padding: hash_record_sha1 was hashing the full 512-byte-aligned record buffer instead of just the real file content, so every APK-TOOLS.checksum.SHA1 pax extension was wrong.
Datahash computed on the wrong bytes entirely: the .PKGINFO datahash field is checked by apk against the SHA-256 of the compressed data.tar.gz bytes, not the uncompressed data.tar. fpm was hashing the raw tar before compression. I confirmed this by patching apk-tools' extract_v2.c to print the digest it calculates - it matched sha256(data.tar.gz) exactly, not sha256(data.tar). Fixed by restructuring output to gzip the data tar first, hash those compressed bytes, then build .PKGINFO/control tar/compress it, then concatenate the two already-compressed streams.
Found and fixed a related latent bug in hash_datatar_sha1: ownership-header rewriting was being applied even to the null (end-of-archive) terminator blocks, polluting bytes that must stay all-zero.

### DIFF
--- a/lib/fpm/package/apk.rb
+++ b/lib/fpm/package/apk.rb
@@ -91,9 +91,14 @@ class FPM::Package::APK< FPM::Package
     # data tar.
     tar_path(staging_path(""), datatar_path)
 
+    begin
+      # calculate/rewrite sha1 hashes for data tar
+      hash_datatar_sha1(datatar_path)
+    end
+
     # control tar.
     begin
-      write_pkginfo(control_path)
+      write_pkginfo(control_path, datatar_path)
       write_control_scripts(control_path)
       tar_path(control_path, controltar_path)
     ensure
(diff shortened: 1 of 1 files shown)


dgtlrift:
Created new PR per @jordansissel recommendation

dgtlrift:
Can someone review and commit?

dgtlrift:
Can this be pushed in?

jordansissel:
Confirmed that this PR generates packages that install with `apk add --allow-untrusted`. I tested with an alpine container attempting to `apk add` a package generated before/after this PR, and only the one with this PR is accepted. Yay!

I'm not familiar with apk's format, so I'm unable to evaluate the impact on the rest of fpm's apk support, and there aren't any test cases for apk in fpm's suite.

In the absence of a better way for me to test this, I'm OK merging this. Thank you for improving fpm! :)

